PR #1117, opened 2015-01-13
Notification Kill Switch

phil-lopreiato:
Like we discussed in #1105, here's the ability to disable sending push notifications from the admin panel. Merge #1105 first, this is based on it


controllers/admin/admin_mobile_controller.py
```diff
@@ -23,17 +24,46 @@ def get(self):
...
+
         self.template_values.update({
             'mobile_users': all_clients.count(),
             'android_users': android,
             'ios_users': ios,
             'webhooks': webhook,
             'broadcast_success': self.request.get('broadcast_success'),
+            'push_enabled': push_enabled,
+            'disp_enable': disp_enable,
+            'disp_disable': disp_disable,
```
fangeugene: These three can be combined into one boolean that gets passed to the template, and the logic should reside in the template.

Issue #1103, opened 2014-12-11
DatafeedFms urls return an error


mc10:
Specifically, https://my.usfirst.org/frc/scoring/index.lasso?page=eventlist and https://my.usfirst.org/frc/scoring/index.lasso?page=teamlist are broken, causing builds to fail.

gregmarra:
This also prevents bootstrapping a test instance with team data.

gregmarra:
FRC API will be the future here.

PR #1123, opened 2015-01-19
Adds local test team creation to bypass broken FMS list.

gregmarra:
While #1103 exists, allows creating local test teams to create test matches, events, etc.

### DIFF
--- a/helpers/event_team/event_team_test_creator.py
+++ b/helpers/event_team/event_team_test_creator.py
@@ -2,6 +2,7 @@ from helpers.event_team_manipulator import EventTeamManipulator
 from models.event_team import EventTeam
 from models.team import Team
 
+import logging
 
 class EventTeamTestCreator(object):
     @classmethod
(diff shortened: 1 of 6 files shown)


helpers/event_team/event_team_test_creator.py
```diff
@@ -2,6 +2,7 @@
 from models.event_team import EventTeam
 from models.team import Team
 
+import logging
```
gregmarra: Should remove this.

tests/test_team_test_creator.py
```diff
@@ -0,0 +1,30 @@
+import datetime
+import unittest2
+
+from google.appengine.ext import testbed
+
+from helpers.team.team_test_creator import TeamTestCreator
+
```
gregmarra: don't need this linebreak.

gregmarra:
@fangeugene if this looks good to you, you can merge it. It's been a while since I've committed anything, so I want a second set of eyes :)

PR #1098, opened 2014-11-16
More notifications

phil-lopreiato:
Implementation of the rest of our outlined notification types. 

Needs to be thoroughly tested, but I'll take other comments now.

PR #1124, opened 2015-02-03
Notification Testing

phil-lopreiato:
Depends on #1098, so merge that first.

Lets you send test instances of notifications to all of your devices from the webhook doc page (tba.com/apidocs/webhooks). Each enabled notification will have a button that sends a static instance to all of the clients associated with the logged in account. The events/matches are static at 2014necmp and 2014necmp_f1m1

How's that for customer support, @ZachOrr? 

(also, #1124! Wooo!)


ZachOrr:
I can dig it! Thanks Phil.

cron_main.py
```diff
@@ -86,5 +88,6 @@
                                ('/tasks/admin/clear_old_subs', AdminSubsClear),
                                ('/tasks/admin/enqueue/clear_old_webhooks', AdminWebhooksClearEnqueue),
                                ('/tasks/admin/clear_old_webhooks', AdminWebhooksClear),
+>>>>>>> more-notifications
```
fangeugene: wat

cron_main.py
```diff
@@ -79,6 +80,7 @@
                                ('/tasks/math/enqueue/typeaheadcalc', TypeaheadCalcEnqueue),
                                ('/tasks/math/do/typeaheadcalc', TypeaheadCalcDo),
                                ('/tasks/posts/firebase_push', FirebasePushDo),
+                               ('/tasks/test_notification/([0-9]+)', TestNotificationController)
```
fangeugene: this endpoint is only accessible by admins

phil-lopreiato: Why is that?

controllers/test_notification_controller.py
```diff
@@ -0,0 +1,71 @@
...
+            # Not passed a valid int, return
+            self.redirect('/apidocs/webhooks')
+            return
+
+        keys = PushHelper.get_client_ids_for_users([user_id])
+        logging.info("Keys: {}".format(keys))
+        if notification:
+            notification.send(keys)
+
+        self.redirect('/apidocs/webhooks')
```
fangeugene: This redirection isn't working for me.

phil-lopreiato: That's odd. Is there an error in the console?

cron_main.py
fangeugene: In app.yaml:
```
- url: /tasks/.*
  script: cron_main.app
  login: admin
```

because we don't want to allow random people to run our tasks such as `/tasks/enqueue/usfirst_event_alliances/XXXX`

fangeugene: You should make this the same way ping was handled.

fangeugene:
Ah, you're missing commas in cron_main

controllers/test_notification_controller.py
fangeugene: Ah, you're missing commas in cron_main

cron_main.py
phil-lopreiato: Good to know. So just move it to a different name and define it in main.py?

phil-lopreiato:
You're right, I am missing commas. Must've fell through while merging stuff in. Fixes incoming

cron_main.py
fangeugene: Yeah, follow a similar pattern to how ping was done.

main.py
```diff
@@ -66,6 +67,7 @@ def __call__(self, request, response, exception):
       RedirectRoute(r'/logout', AccountLogout, 'account-logout', strict_slash=True),
       RedirectRoute(r'/match/<match_key>', MatchDetail, 'match-detail', strict_slash=True),
       RedirectRoute(r'/notifications/broadcast', UserNotificationBroadcast, 'notification-broadcast', strict_slash=True),
+      RedirectRoute(r'/notifications/test/([0-9]+)', TestNotificationController),
```
fangeugene: `/notifications/test/<type:[0-9]+>`

phil-lopreiato: That's what I get for pushing and deploying to test instance at the same time :P

fangeugene: ;)

phil-lopreiato:
@ZachOrr should be up and live soon-ish (probably)

PR #1140, opened 2015-03-07
Add Firehose to Gameday

phil-lopreiato:
Puts the ticker as seen on http://tba-notification-firehose.appspot.com/ into GameDay


fangeugene:
Continued in https://github.com/the-blue-alliance/the-blue-alliance/pull/1144

PR #1149, opened 2015-03-19
Add Custom Hashtag to Event

popnbrown:
Created a new field to allow events to have custom hashtags.

### DIFF
--- a/models/event.py
+++ b/models/event.py
@@ -271,9 +271,9 @@ class Event(ndb.Model):
             return None
 
     @property
-    def hashtag_event(self):
+    def hashtag(self):
         """
-        Return the hastag used for the event.
+        Return the hashtag used for the event.
         """
         if self.custom_hashtag:
             return self.custom_hashtag
--- a/models/event.py
+++ b/models/event.py
@@ -28,6 +28,7 @@ class Event(ndb.Model):
     official = ndb.BooleanProperty(default=False)  # Is the event FIRST-official?
     first_eid = ndb.StringProperty()  # from USFIRST
     facebook_eid = ndb.StringProperty(indexed=False)  # from Facebook
+    custom_hashtag = ndb.StringProperty(indexed=False) #Custom HashTag
     website = ndb.StringProperty(indexed=False)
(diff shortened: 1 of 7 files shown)


models/event.py
```diff
@@ -269,6 +270,16 @@ def gameday_url(self):
         else:
             return None
 
+    @property
+    def hashtag_event(self):
```
gregmarra: Is there a reason to not name this just `hashtag`?

gregmarra:
I'd prefer the property to just be `hashtag`, but otherwise LGTM. @fangeugene any concerns?

phil-lopreiato:
Do you think it would be worth having a way for users to suggest hashtags? That way, you and Eugene don't get pestered with requests to change them

popnbrown:
@gregmarra I was just following the "facebook_event_url" function, thought that there was a reason for having event in the function name. Easy change, I'll get it up in a bit.

phil - That may be a slippery slope, unlike webcasts, hashtags aren't something a moderator can verify to be correct or not. Generally, they will come from the event organizing committee, and probably from very few. I don't want to put moderators in a situation where they approve an incorrect hashtag. There is a decent chance I'm overthinking this as I've been talking about hashtags with people for the last week.

popnbrown:
@gregmarra committed, will travis re-run for the new commit?

phil-lopreiato:
@popnbrown that's a good point - it probably doesn't matter much in the long run.

And yes, travis will rerun automatically for all new commits on PRs. You can see the output of the build [here](https://travis-ci.org/the-blue-alliance/the-blue-alliance/builds/55179671)

synth3tk:
You can also see the status of the latest run at the bottom of this page (on desktop).

popnbrown:
@gregmarra @fangeugene any chance of this moving live by next week?

fangeugene:
Just checked out this PR. Looks good to me. @gregmarra any other comments?

gregmarra:
I am good w this. Let's land it.

popnbrown:
thanks gents!

PR #1152, opened 2015-03-22
Count WFFA's won after 2015 as blue banners for insights

legoktm:
See <http://www.usfirst.org/roboticsprograms/frc/blog-honoring-woodie-flowers-finalist-award-winners>.

### DIFF
--- a/consts/award_type.py
+++ b/consts/award_type.py
@@ -82,7 +82,12 @@ class AwardType(object):
     PROFESSIONALISM = 65
     GOLDEN_CORNDOG = 66
 
-    BLUE_BANNER_AWARDS = {CHAIRMANS, WINNER}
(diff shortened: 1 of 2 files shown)


fangeugene:
Dang, 604 should have 4 more banners then.

fangeugene:
I'm kind of with Tom on this one.

popnbrown:
While that's true, they (FIRST) wants to make the WFFA a team award as well. I mean by technical definition of having blue banners, the teams now have blue banners.

synth3tk:
It seems like a grey area. Technically, the team now *displays* a blue banner that the individual won. But OTOH, the individual won it, not the team.

I'm leaning towards not counting it.

PR #1157, opened 2015-03-29
Update team_renderer.py to only display year-long WLT for official events

AGPapa:
Removes offseason events in calculations of season-long WLT or averages. I haven't used python or the tba repo before, so can you check if this is correct?

### DIFF
--- a/renderers/team_renderer.py
+++ b/renderers/team_renderer.py
@@ -30,7 +30,6 @@ class TeamRenderer(object):
 
         participation = []
         year_wlt_list = []
-        offseason_wlt_list = []
         year_match_avg_list = []
 
         current_event = None
@@ -51,7 +50,8 @@ class TeamRenderer(object):
             if year == 2015:
                 display_wlt = None
                 match_avg = EventHelper.calculateTeamAvgScoreFromMatches(team.key_name, event_matches)
-                year_match_avg_list.append(match_avg)
+                if event.event_type_enum != EventType.PRESEASON and event.event_type_enum != EventType.OFFSEASON:
+                    year_match_avg_list.append(match_avg)
                     
                 qual_avg, elim_avg, _, _ = match_avg
             else:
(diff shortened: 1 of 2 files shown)


AGPapa:
Just to elaborate on what this does, if you look at http://www.thebluealliance.com/team/254/2014 you'll see that they're season record is "112-16-0". But 254 only had 7 looses in the official season. That record displayed includes all of their offseason matches. This change should exclude offseason events from the overall record, which more accurately reflect's each team's season.

nwalters512:
Is this the best way to do things? I'm sure many people would want to see how teams performed for *all* of their events, not just official ones. If anything, we should show two records, one for all events and one for only official ones. I have a hard time justifying the exclusion of information.

fangeugene:
I like Nathan's idea for showing both averages. @AGPapa do you know what changes need to be made in order to make this happen?

In the current state, your code looks good. One thing that is missing is an import at the top of the file `from consts.event_type import EventType`. Try to keep all `from ...` imports in alphabetical order, so this would go before `from google ...` (line 3 or so)

AGPapa:
I think that including offseason events in with official events can be misleading and hide how strong a team's season was. Teams often don't take offseason events as seriously as official ones. 

Anyway, I tried to modify the code to display offseason wlt in parenthesis. So 254's page would show "70(42)-7(9)-0(0)". This is a much larger change than the previous one, and I don't completely understand how to implement it correctly.

nwalters512:
Without a pretty verbose explanation, people would probably have no idea what the numbers in parentheses mean. A better way of doing things would probably be to just construct a human-readable string, like "The team's overall record was #-#-#; at official events, their records was #-#-#." This should be idiot-proof.

AGPapa:
This is too involved for me right now. I don't understand how this works, and I can't seem to put my changes back to the initial ones where the build passed. I'm just going to leave it here. I still think that only including the official results in the overall record is a good change.

PR #1159, opened 2015-03-30
Add limit to video suggestions review page.

gregmarra:
This will allow us to review matches without totally destroying a browser when there are hundreds pending.

Test plan: I tested this locally by adding a suggestion to a match, and seeing that the suggestion page loaded.


gregmarra:
Test failures all seem related to USFIRST datafeeds, which are currently expected to fail.

PR #1168, opened 2015-04-08
Write API Interface

phil-lopreiato:
Picked up from [fangeugene/write-api-interface](https://github.com/fangeugene/the-blue-alliance/tree/write-api-interface).

Adds ability to submit matches 

TODO: fix broken tests


phil-lopreiato:
Also TODO make sure that you can submit unplanned matched with only the local time string and have the utc time guessed, like the html parsers used to do

PR #1174, opened 2015-04-15
Update country_name key in API docs.

CarlColglazier:
Stubbed my toe on this one a bit earlier. The docs declare `country` as being the key, when the API actually uses `country_name`.


synth3tk:
:+1:

PR #1181, opened 2015-04-20
Adds Facebook page plugin to homepage LHC.

gregmarra:
Will make it easier for people to Like our Facebook page, sets us up for doing something similar on team pages.

Plugin automatically gets narrower when needed (only on page reload, but not a problem for mobile devices).

screen shot 2015-04-19 at 17 09 00


gregmarra:
@fangeugene can you give me a quick LGTM? :)

fangeugene:
Looks good to me, but the slightest bit busy IMO. Thoughts, @gregmarra ?

synth3tk:
Revisiting this, do we still want to do it?

fangeugene:
I'm no a huge fan of how busy it is

SUPERCILEX:
@fangeugene Do you want to close this? TBA now has an "Upcoming events" section which kinda takes care of this:
image

synth3tk:
It doesn't cover all of the other posts types, but I'm not sure if that's even necessary at all. IMHO, we can go without the FB widget.

andrewda:
I agree.

fangeugene:
We can revisit making a "like" button on the homepage more prominent, but I think the widget is overkill.

PR #1183, opened 2015-04-21
Add HLS webcast support

synth3tk:
Feel free to give it a quick lookover. This was a dirty hack but it works. Tested on Chrome, Firefox and IE on Windows 10.

Models look like this:
```[{"type": "html5", "channel": "http://public.infozen.cshls.lldns.net/infozen/public/public.m3u8"}]```


synth3tk:
I'll be merging this right before I get ready to leave. Because what can go wrong?!

phil-lopreiato:
#shipIt

synth3tk:
:+1:

PR #1197, opened 2015-05-09
Event Wizard

phil-lopreiato:
Adds an "Event Wizard" using the trusted API. Can be found at `tba.com/eventwizard`. You can input your auth keys, store them locally, and input Teams Attending, Schedule, Match Play, Match Results, Alliances, and Rankings. Schedule, Match Results, and Rankings can be input via FMS reports, which are all parsed client side. 

Changed the trusted API to take W/L/T out of required parameters (moved to generic breakdown category) to accommodate this year's game. This will require an update to Cheesy Arena (which I'll take care of).

Closes #1115 and closes #1190

PR #1201, opened 2015-05-11
Subdivide trusted api permissions

phil-lopreiato:
Removes EVENT_DATA permission and splits it up so each endpoint has its own permission. 

Split from #1197

### DIFF
--- a/consts/auth_type.py
+++ b/consts/auth_type.py
@@ -4,8 +4,19 @@ class AuthType(object):
     """
     EVENT_DATA = 0
     MATCH_VIDEO = 1
+    EVENT_TEAMS = 2
+    EVENT_SCHEDULE = 3
+    EVENT_MATCHES = 4
+    EVENT_RANKINGS = 5
+    EVENT_ALLIANCES = 6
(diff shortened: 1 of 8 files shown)


fangeugene:
is `EVENT_DATA` completely unused now? If so, we should remove it from auth_type.py and renumber everything. It's okay to remove and renumber because we currently don't have any Auths in the DB.

fangeugene:
Also, can you make the tests test each of the auth types individually?

phil-lopreiato:
I've removed `EVENT_DATA` and updated the tests. Anything else?

gregmarra:
I think it's generally bad practice to renumber enums. The point is you shouldn't be thinking about the numbers, and we aren't constrained in total size.

fangeugene:
Yeah, @gregmarra is right. Should we keep the original numbers but comment out the ones that are deprecated?

phil-lopreiato:
That's probably a better approach. Let me revert that commit

phil-lopreiato:
All right, that should be better

PR #1203, opened 2015-05-18
Event Wizard Updates

phil-lopreiato:
This fixes the bugs I found running the Event Wizard last weekend at CTSC. It also makes some minor improvements.

### DIFF
--- a/controllers/api/api_trusted_controller.py
+++ b/controllers/api/api_trusted_controller.py
@@ -106,10 +106,10 @@ class ApiTrustedEventMatchesUpdate(ApiTrustedBaseController):
                 time=match.get("time", None),
             )
 
-            if (not match.time or match.time == "") and match.time_string:
-                # We can calculate the real time from the time string
-                logging.debug("Calculating time!")
-                MatchHelper.add_match_times(event, [match])
+            # if (not match.time or match.time == "") and match.time_string:
+            #     # We can calculate the real time from the time string
+            #     logging.debug("Calculating time!")
+            #     MatchHelper.add_match_times(event, [match])
 
             matches.append(match)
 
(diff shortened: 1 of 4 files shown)


controllers/api/api_trusted_controller.py
```diff
@@ -106,10 +106,10 @@ def _process_request(self, request, event_key):
                 time=match.get("time", None),
             )
 
-            if (not match.time or match.time == "") and match.time_string:
-                # We can calculate the real time from the time string
-                logging.debug("Calculating time!")
-                MatchHelper.add_match_times(event, [match])
+            # if (not match.time or match.time == "") and match.time_string:
```
fangeugene: Can we try/catch this instead of disabling it completely?

phil-lopreiato: That's a better idea. Added.

controllers/api/api_trusted_controller.py
```diff
@@ -108,11 +109,16 @@ def _process_request(self, request, event_key):
 
             if (not match.time or match.time == "") and match.time_string:
                 # We can calculate the real time from the time string
-                logging.debug("Calculating time!")
-                MatchHelper.add_match_times(event, [match])
-
+                needs_time.append(match)
             matches.append(match)
 
+        if needs_time:
```
fangeugene: Why not keep this in the original for loop? That way one match can fail, but the others can succeed.

phil-lopreiato: Because I actually implemented it wrong the first time. That method should take a list of events, and it will start at the end of the list, and calculate times starting on the last day of the event and work backwards. So for multiday events, it would always think the matches are on the last day, since it will always assume the first match in the list takes place on the final day of the event.

fangeugene: Ah right.

PR #1204, opened 2015-05-20
Trusted API Live Event Auth

phil-lopreiato:
If the flag for the auth is checked (defaults to true), the key will
only be allowed if the event is within_a_day


gregmarra:
@phil-lopreiato  why do you think this is important? Is there a reason to stop trusting keys?

phil-lopreiato:
This isn't hugely important, mostly a convenience feature for @fangeugene
He typically deleted the keys after an event finishes just to be sure they don't get changed. This allows him to forget to do that and still have the keys not work.

PR #1210, opened 2015-05-27
EventWizard import Ranks from FMS server

phil-lopreiato:
Import data using the same ajax data that Pit Display uses. Can't be tested, because I don't have FMS

### DIFF
--- a/static/javascript/tba_js/eventwizard_apiwrite.js
+++ b/static/javascript/tba_js/eventwizard_apiwrite.js
@@ -326,29 +326,6 @@ $('#rankings_file').change(function(){
     reader.readAsBinaryString(f);
 });
 
-$('.update-rankings').click(function(){
-    $(this).css('background-color', '#eb9316');
-    $.ajax({
-            url: '10.0.100.5/pit/getdata?random=' + Math.random(),
-            cache: false,
-            timeout: 5000, success: function (data) {
-                rankings = ['Rank', 'Team', 'Avg', 'CP', 'AP', 'RC', 'TP', 'LP', 'Played'];
-                rankData = JSON.parse(data);
-                for(var i=0; i<rankData.length; i++){
-                    // Turn team number -> team key
-                    rankData[i]['Team'] = "frc"+rankData[i]['Team'];
-                    rankings.push(rankData[i]);
-                }
-                makeRequest('/api/trusted/v1/event/' + $('#event_key').val() + '/rankings/update', JSON.stringify(rankings), $(this));
-            },
-            error: function (error) {
-                // We had an error getting the results so set the top of the screen red
-                alert("Error getting rankings. Are you sure you're connected to the field network?");
-               $(this).parent().css({ 'background-color': 'red' });
-            }
-    });
-});
-
 $('#teams_file').change(function(){
     var f = this.files[0];
     var reader = new FileReader();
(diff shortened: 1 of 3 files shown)


gregmarra:
Do you have dummy data? Can you make dummy data you think is the same
format? Will allow at least a unit test.

phil-lopreiato:
I was thinking that, but how should I unit test javascript? We don't really have anything set up for that.

gregmarra:
Oh good point. That could be a separate summer project.

static/javascript/tba_js/eventwizard_apiwrite.js
```diff
@@ -326,6 +326,29 @@ $('#rankings_file').change(function(){
     reader.readAsBinaryString(f);
 });
 
+$('.update-rankings').click(function(){
+    $(this).css('background-color', '#eb9316');
+    $.ajax({
+            url: '10.0.100.5/pit/getdata?random=' + Math.random(),
+            cache: false,
+            timeout: 5000, success: function (data) {
+                rankings = ['Rank', 'Team', 'Avg', 'CP', 'AP', 'RC', 'TP', 'LP', 'Played'];
```
fangeugene: So we need to change this every year.

phil-lopreiato: Correct :/

static/javascript/tba_js/eventwizard_apiwrite.js
```diff
@@ -326,6 +326,29 @@ $('#rankings_file').change(function(){
...
+                rankData = JSON.parse(data);
+                for(var i=0; i<rankData.length; i++){
+                    // Turn team number -> team key
+                    rankData[i]['Team'] = "frc"+rankData[i]['Team'];
+                    rankings.push(rankData[i]);
+                }
+                makeRequest('/api/trusted/v1/event/' + $('#event_key').val() + '/rankings/update', JSON.stringify(rankings), $(this));
```
fangeugene: Is this the right format to make the trusted API call? https://github.com/the-blue-alliance/the-blue-alliance/blob/master/datafeeds/parsers/json/json_rankings_parser.py

phil-lopreiato: Oh, wait, no it isn't. Will fix tonight

static/javascript/tba_js/eventwizard_apiwrite.js
```diff
@@ -326,6 +326,29 @@ $('#rankings_file').change(function(){
...
+            },
+            error: function (error) {
+                // We had an error getting the results so set the top of the screen red
+                alert("Error getting rankings. Are you sure you're connected to the field network?");
```
fangeugene: This alert isn't firing for me.

phil-lopreiato: Check the network inspector in devtools and see if the request is timing out?

fangeugene: Actually, no request is being made.
`Uncaught TypeError: Cannot read property 'value' of undefined`

phil-lopreiato: I'll fix it all up tonight when I get home from work.

PR #1218, opened 2015-06-15
YT Videos from EventWizard

### LINKED ISSUE #1213: Event Wizard: Ability to add YouTube videos to matches
Needed for Texas Robotics Invitational on 6/20.

phil-lopreiato:
Input YT Key along with score

Closes #1213

### DIFF
--- a/controllers/api/api_trusted_controller.py
+++ b/controllers/api/api_trusted_controller.py
@@ -218,3 +218,5 @@ class ApiTrustedAddMatchYoutubeVideo(ApiTrustedBaseController):
                 match.dirty = True  # This is hacky -fangeugene 2014-10-26
                 matches_to_put.append(match)
         MatchManipulator.createOrUpdate(matches_to_put)
+
+        self.response.out.write(json.dumps({'Success': "Match videos successfully updated"}))
--- a/static/javascript/tba_js/eventwizard_apiwrite.js
+++ b/static/javascript/tba_js/eventwizard_apiwrite.js
@@ -25,6 +25,17 @@ function makeRequest(request_path, request_body, feedback) {
   });
 }
 
+function getYoutubeId(url) {
+    // regex from http://stackoverflow.com/a/9102270
+    var regex = /^.*(youtu.be\/|v\/|u\/\w\/|embed\/|watch\?v=|\&v=)([^#\&\?]*).*/;
+    var match = url.match(regex);
+    if (match && match[2].length == 11) {
+        return match[2];
+    } else {
+        return null;
+    }
+}
+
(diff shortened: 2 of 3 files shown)


fangeugene:
Wow, this is dead simple, isn't it? #goodesign

phil-lopreiato:
That's how I like it!

fangeugene:
I think this would be better it accepted the whole YT link, and we regexed out the ID.

phil-lopreiato:
:+1: 
To future me, regex: http://stackoverflow.com/a/8260383

I'll also fix the tests tonight, which it looks like I broke

controllers/api/api_trusted_controller.py
```diff
@@ -218,3 +218,5 @@ def _process_request(self, request, event_key):
                 match.dirty = True  # This is hacky -fangeugene 2014-10-26
                 matches_to_put.append(match)
         MatchManipulator.createOrUpdate(matches_to_put)
+
+        self.response.out.write(json.dumps({'Success': "Match videos successfully updated"}))
```
phil-lopreiato: This is probably what broke the tests.

Added because without some return value, the ajax callback was reporting successes as errors. I can fix the tests to account for that

fangeugene:
I think tests were already broken.

phil-lopreiato:
[This one looks like my fault](https://travis-ci.org/the-blue-alliance/the-blue-alliance/builds/66803439#L768)

phil-lopreiato:
Oh wait, test failures come from [this commit](https://github.com/the-blue-alliance/the-blue-alliance/commit/072058b64f18375f0be319d0d1e6b474b13592d6). Teams aren't in db at time of test, don't get added. We should fix this case and add a separate case ensuring that.

static/javascript/tba_js/eventwizard_apiwrite.js
```diff
@@ -426,4 +437,14 @@ $('#match-table').on('click', 'button', function(e) {
...
     makeRequest('/api/trusted/v1/event/' + $('#event_key').val() + '/matches/update', request_body, $(this).parent());
+
+    var video_request = {};
+    var yt_url = $("#"+matchKey+"_video").val();
+    var video_key = getYoutubeId(yt_url);
+    if (!video_key) {
+        return;
+    }
+    video_request[matchKey.split('_')[1]] = video_key;
+    alert(JSON.stringify(video_request));
```
fangeugene: Was this for debugging? I'm getting rid of it.

phil-lopreiato: Oh, whoops. Yeah, you can kill it

PR #1228, opened 2015-06-23
Ignore matches without a score

### LINKED ISSUE #1223: Event Wizard - FMS Results input after Quarters returns frcundefined error
Attempting to import match results after Quarterfinals returns an frcundefined error due to Finals schedule being unknown.

Workaround: Manually delete Finals rows from the spreadsheet before importing.

BowlesCR:
This should fix #1223

### DIFF
--- a/static/javascript/tba_js/eventwizard_apiwrite.js
+++ b/static/javascript/tba_js/eventwizard_apiwrite.js
@@ -197,8 +197,8 @@ $('#results_file').change(function(){
         for(var i=0; i<matches.length; i++){
             var match = matches[i];
 
-            // check for invalid match
-            if(!match['Time']){
+            // check for invalid/unplayed match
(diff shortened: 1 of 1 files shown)


static/javascript/tba_js/eventwizard_apiwrite.js
```diff
@@ -198,7 +198,7 @@ $('#results_file').change(function(){
             var match = matches[i];
 
             // check for invalid match
-            if(!match['Time']){
+            if(!match['Time'] || !match['Red 1']){
```
phil-lopreiato: Let's make this line `if(!match['Time'] || !match['Red Score'] || !match['Blue Score']){` to be safest

phil-lopreiato: Actually, time shouldn't be a mandatory field here. Just check for 'Red Score' and 'Blue Score'

Issue #1191, opened 2015-04-27
Create "summer hack" project list, supporting resources


gregmarra:
Identify a list of projects people might contribute towards over the summer. Probably small projects like "fix this typo" and large projects like "build this new feature".

Projects should be reasonably well laid out, and have a clear "point of contact" of an existing contributor if the person who picks them up has questions.

I'm going to start applying "Help Wanted" to these, including this one :)

Tips:
https://18f.gsa.gov/2015/04/03/how-to-welcome-new-coders-to-a-civic-hackathon/

phil-lopreiato:
Some ideas/things that have been kicking around for a while:
 - [ ] myTBA support on event & team pages (manage favorite/notification settings)
 - [ ] Implement 'Final Results' notifications - show the final status of a team@event
 - [ ] Some extension of #576 would be cool - have the ability to add "Pit Display"-like thing to GameDay instead of webcast: show next/last X matches, live-updating rankings
 - [ ] iOS support for push notifications
 - [ ] Sitevar for FRCAPI status: dict with boolean for each endpoint, check/uncheck each in admin panel
 - [ ] Disable FRCAPI on a per-event basis (same dict as above)
 - [ ] Make event alliances endpoint cronjob run more frequently (every 5 mins?), but only actually hit that endpoint that frequently on the final day of the event. This means we don't have to wait half an hour for notifications to go out. 
 - [ ] Infer event alliances from quarterfinal schedule
 - [ ] Break big json properties (alliances, rankings, etc.) out into their own models? Allow for easier querying/updating
 - [ ] Fix admin broadcast. I'm guessing the query is probably too big with >5,000 clients. Datastore queries should be batched, even though gcm should be able to handle that many
 - [ ] Server side support for https://github.com/the-blue-alliance/the-blue-alliance-android/issues/418
 - [ ] Moar insights, and add them to API
 - [ ] Internal support for DQs and surrogates in matches (and account for them in OPR, API, schedule view)

synth3tk:
We could "assign" the point of contact to each of these issues so they can easily see who it is, as well as add the "Help Wanted" tag.

Also, add Gameday 2 to the list? It seems to have stagnated, so I don't know where we stand on that.

fangeugene:
Good point about GameDay 2. I have ideas for cleaning up the interface (no more drag/drop because that's clunky) but keeping the same functionality. Unsure if building off of Greg's GameDay2 branch that uses React is the right thing to do though.

[...]

PR #1239, opened 2015-09-07
Make event alliances endpoint cronjob run more often

bvisness:
Accomplishes a wish-list item from #1191. The event alliances endpoint cronjob now runs every 5 minutes, but event endpoints are only actually hit that often on the final day of each event. This way alliance selection notifications can go out more quickly with minimal extra load.

I am a newbie to this, so I expect there will be plenty of feedback on this. :)

Also, one question - should I have included tests for the `starts_today` and `ends_today` methods I added to the event model?

### DIFF
--- a/controllers/datafeed_controller.py
+++ b/controllers/datafeed_controller.py
@@ -107,9 +107,20 @@ class FMSAPIEventAlliancesEnqueue(webapp.RequestHandler):
     Handles enqueing getting alliances from the FMS API
     """
     def get(self, when):
+        events = skipped_events = []
         if when == "now":
             events = EventHelper.getEventsWithinADay()
             events = filter(lambda e: e.official, events)
+
+            # Check if we should do a full query, or temporarily skip events
+            # that are not ending today. (All events will be checked on
+            # half-hour intervals, and events ending today whenever the cron
+            # job runs.)
+            minutes_from_full = (((int(time.time()) / 60) + 15) % 30) - 15
+            full_query = abs(minutes_from_full) <= 2
+            if not full_query:
+                skipped_events = filter(lambda e: not e.ends_today, events)
+                events = filter(lambda e: e.ends_today, events)
         else:
             event_keys = Event.query(Event.official == True).filter(Event.year == int(when)).fetch(500, keys_only=True)
             events = ndb.get_multi(event_keys)
@@ -122,6 +133,7 @@ class FMSAPIEventAlliancesEnqueue(webapp.RequestHandler):
 
         template_values = {
             'events': events,
+            'skipped_events': skipped_events
         }
 
         path = os.path.join(os.path.dirname(__file__), '../templates/datafeeds/usfirst_event_alliances_enqueue.html')
--- a/controllers/datafeed_controller.py
+++ b/controllers/datafeed_controller.py
@@ -107,20 +107,13 @@ class FMSAPIEventAlliancesEnqueue(webapp.RequestHandler):
     Handles enqueing getting alliances from the FMS API
     """
     def get(self, when):
-        events = skipped_events = []
+        events = []
         if when == "now":
             events = EventHelper.getEventsWithinADay()
             events = filter(lambda e: e.official, events)
-
-            # Check if we should do a full query, or temporarily skip events
-            # that are not ending today. (All events will be checked on
-            # half-hour intervals, and events ending today whenever the cron
-            # job runs.)
-            minutes_from_full = (((int(time.time()) / 60) + 15) % 30) - 15
-            full_query = abs(minutes_from_full) <= 2
-            if not full_query:
-                skipped_events = filter(lambda e: not e.ends_today, events)
-                events = filter(lambda e: e.ends_today, events)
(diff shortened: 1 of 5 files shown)


controllers/datafeed_controller.py
```diff
@@ -107,9 +107,20 @@ class FMSAPIEventAlliancesEnqueue(webapp.RequestHandler):
...
+        events = skipped_events = []
         if when == "now":
             events = EventHelper.getEventsWithinADay()
             events = filter(lambda e: e.official, events)
+
+            # Check if we should do a full query, or temporarily skip events
+            # that are not ending today. (All events will be checked on
+            # half-hour intervals, and events ending today whenever the cron
+            # job runs.)
+            minutes_from_full = (((int(time.time()) / 60) + 15) % 30) - 15
```
gregmarra: I don't totally understand the logic here. Is the intent to check all events every 30 minutes, and only events ending today every 5?

Would it be clearer to split into two cronjobs, each with more specific jobs and less logic?

gregmarra:
Thanks for getting going with this one! It will speed up how quickly we land alliance selections on the site.

controllers/datafeed_controller.py
bvisness: Yeah, that is what that logic is doing. I realize that line is not very clear. I wasn't sure exactly when the cronjobs would run (if they were exactly at XX:00 and XX:30 or slightly off) so I came up with that to let it run anytime within a 5-minute interval.

Two cronjobs would probably be much clearer and more reliable though. Any ideas for what the URLs should be for each endpoint?

fangeugene: I agree with Greg for having two cron jobs.

Maybe the two endpoints could be something along the lines of:
`/tasks/enqueue/fmsapi_event_alliances/now`
`/tasks/enqueue/fmsapi_event_alliances/last_day_only`

Every 30 minutes we would end up running the cron jobs twice, but that's not too big of an issue IMO.

bvisness:
Ok, I switched it to use two separate cronjobs. Much easier to understand, I think!

gregmarra:
This looks pretty solid. 

When putting up pull requests, it's helpful to explain how you've tested this to make sure it works as expected. This could be anything from "I tested this by creating test events and manually calling the cron functions" to adding full tests for things.

A small upgrade would be adding tests for the new time-related functions you've added to Events to make sure they don't break in the future.

@fangeugene if this looks good to you, you can pull it!

fangeugene:
Looks solid to me as well. I'll run a few tests tonight and then merge it.

P.S. I like what you did with the refactoring of `local_time`.

bvisness:
Thanks! I added a couple tests – should I put those on this PR as well?

fangeugene:
Yep! Add them to this PR.

helpers/event/event_test_creator.py
```diff
@@ -71,3 +71,49 @@ def createPastEvent(self, only_event=False):
             mtc = MatchTestCreator(event)
             mtc.createCompleteQuals()
         return event
+
```
fangeugene: There's no need to create the test events here, since they're not used anywhere else. Just create the Event objects within `test_event.py`

fangeugene:
Also, it would be great to see some tests that verify `event.local_time` works as expected with various timezones.

helpers/event/event_test_creator.py
bvisness: Gotcha, will change that quick.

tests/test_event.py
```diff
@@ -46,3 +52,57 @@ def test_datesPresent(self):
...
+            event_short="teststartstoday",
+            event_type_enum=EventType.REGIONAL,
+            first_eid="5561",
+            name="Test Event (Starts Today)",
+            start_date=datetime.datetime.today(),
+            year=datetime.datetime.now().year,
+            venue_address="123 Fake Street, California, USA",
+            website="http://www.google.com"
+        )
+        event = EventManipulator.createOrUpdate(event)
```
fangeugene: I don't think you need lines 79-85 (and 102-108 below)

You should be able to get away with having this in `setUp()` (and one for `self.event_ends_today`)
The matches aren't needed because the test for starts_today and ends_today only depends on the Event object.

```
self.event_starts_today = Event(
            id=id_string,
            end_date=datetime.datetime.today() + datetime.timedelta(days=2),
            event_short="teststartstoday",
            event_type_enum=EventType.REGIONAL,
            first_eid="5561",
            name="Test Event (Starts Today)",
            start_date=datetime.datetime.today(),
            year=datetime.datetime.now().year,
            venue_address="123 Fake Street, California, USA",
            website="http://www.google.com"
        )
```

fangeugene:
@bvisness I can merge once you fix up the tests. No rush though. Great work!

gregmarra:
@bvisness anything we can do to help here? I'd love to get this merged in :D

bvisness:
I’ll hopefully have time soon to look into this again. College started back up for me with a vengeance so I haven’t had much time to dedicate to stuff like this.

PR #1242, opened 2015-09-07
Use HTTPS link for instagram.com

legoktm:


fangeugene:
We should use the instagram more... @gregmarra

gregmarra:
Yea! If anyone wants to insta lmk and I'll set you up the password.

PR #1257, opened 2015-10-14
Read actualStartTime from FMSAPI Match

phil-lopreiato:
We can do cool things like predicted schedules and determining if an
event is live posting data or not

Test failures are UsFirstDatafeed related, so not affected by this. Updated test cases for this parser pass


gregmarra:
This looks good to me.

PR #1258, opened 2015-10-14
Fetch all team details from fmsapi

phil-lopreiato:
Create new datafeed functions to fetch team data using only FMSAPI.

Add team website (the only thing currently missing) to team details
parser and update tests accordingly

Team website added to team responses: https://usfirst.collab.net/sf/go/artf4060


gregmarra:
@fangeugene are you comfortable with these changes? This looks good to me if all the tests pass.

fangeugene:
We should make use of the fact that the FRC API allows making requests for multiple teams in a single call.

phil-lopreiato:
@fangeugene how would you recommend doing that?

phil-lopreiato:
@fangeugene it doesn't look like you can request multiple specific teams from the [TeamListing endpoint](http://docs.frcevents2.apiary.io/#reference/season-data/team-listings/retrieve-team-listings).

We could, however, change the rolling team details fetch and just query the entire list of teams for a season (paginated, of course) and roll on the page number, not the team number.

There are currently 45 pages in the response, querying one page every 4 hours is 6 pages/day so every team would get updates about once per week. Sound good?

I like that idea much better than doing the rolling by team number

phil-lopreiato:
Additionally, #1271 might make this entire cronjob obsolete, since it already gets the same data back for every team (registered for an event) in the season.

gregmarra:
This looks good to me, excited to get rid of the legacy datafeeds when we can.

Issue #1262, opened 2015-10-20
Idea: Provide an embeddable frame of event data for use in other websites


tytremblay:
Allow teams to embed the information below as a frame with internal scrolling on other websites.

![image](https://cloud.githubusercontent.com/assets/6652493/10620347/90f8da7c-7748-11e5-96e9-785020496733.png)

fangeugene:
Wondering what the best way to do this is.

The simplest is an iframe embed, but that would offload all rendering and bandwidth on us.

Another is a JS widget that hits the API to display the info. Since the API is heavily cached, this may be a better option. This option can already be implemented by third parties if they want to, but having a provided widget can standardize the look and perhaps brand it with "Provided by The Blue Alliance" or something.

gregmarra:
We can also cache the iframe html

tytremblay:
Does that mean this is doable? NE FIRST would like to use this on their new site.

phil-lopreiato:
@tytremblay yeah, it should be pretty doable.

Maybe we could structure the client side stuff similar to the [FB Page Plugin](https://developers.facebook.com/docs/plugins/page-plugin), where you just have to include a script + add a div with some data elements to configure what you see. 

As long as we can get the js file cached on a CDN and make sure all API requests hit cache (which they already do), it shouldn't be a burden at all

tytremblay:
Excellent. Is it too early to ask about lead time?  Any way I can help?

phil-lopreiato:
@fangeugene how do you want to run this? I'm partial to your second idea, the js widget (iframes suck and we can rely on API caching and stuff. We can also leverage firebase to post to an event-specific feed, so we can just render a [notification ticker](http://tba-notification-firehose.appspot.com/) client side, like on GameDay

@tytremblay what do you want it to look like? I'm not sure just a cut/paste of the event page would translate well to a widget. I was thinking of something like this...
![](http://imgur.com/XDAi3qJ.jpg)
If you want to write some client side stuff, you should just have to hit the Event Info/Event Teams endpoints, plus some JS for firebase. I can direct you to places in the TBA repo, if you want

tytremblay:
NE FIRST is thinking something like this:http://i.imgur.com/8Nv3BDF.png

tytremblay:
The idea is to reduce the amount of variability between event pages in NE. Id like users to have a unified experience of team lists, scores, rankings, awards, etc. 

[...]

PR #1264, opened 2015-10-22
Add Firebase feeds for events/districts

phil-lopreiato:
Working towards #1262

For every notification (which we currently push to the ticker firebase), also push it to an endpoint for the specific event (at event/<key>) and for the parent district, only if the event is within a district (at district/<abbrev>).


gregmarra:
This seems reasonable to me, @fangeugene any concerns?

Issue #1256, opened 2015-10-14
[Spec] Mobile/API Status


phil-lopreiato:
We want to be able to send some metadata to the client apps about TBA state and such.

Things we want to include:
 - Current max competition season (so we can flip to the next year early fall when events start to get posted)
 - Is FIRST datafeed down?
 - Minimum android version (so we can prompt updates when we release fixes)
 - Are any current events without internet (no data at one specific event)

We would determine if the entire datafeed is down if any FMSAPI response returns a 5XX error. For an individual event, we could use the "Upcoming Match" cronjob. If the "next match" was supposed to happen >30 minutes ago, we can assume that the datafeed is down and store the result in a sitevar (which we could clean out weekly). We can also use the AutoStartTime field ([now in the HybridResponse](https://usfirst.collab.net/sf/go/artf4041)) to make sure we don't call events running behind broken.

Example API response:
```
{
  max_season: 2015,
  is_datafeed_down: false,
  android: {
    min_app_version: 123456
  },
  is_event_down: {
    2015test: true,
    2015meow: true,
    2015down: false
  }
}
```

fangeugene:
Should `is_event_down` include the false events or just the true ones?

phil-lopreiato:
Good idea, maybe we do an array of down event keys instead

first3234:
I am haveing trouble trying to import a test schedule to the event and I am using The tba match importer excel sheet 

Date: Wed, 14 Oct 2015 12:09:42 -0700

phil-lopreiato:
Any final comments about this spec?

```
{
  max_season: 2015,
  is_datafeed_down: false,
  android: {
    min_app_version: 123456
  },
  ios: {
    min_app_version: 1243
  },
  down_events: {
    2015test,
    2015down
  }
}
```

fangeugene:
lgtm

fangeugene:
down_events should be a list. JSON doesn't do sets.

phil-lopreiato:
Right, good catch

phil-lopreiato:
Also add `latest_app_version` param. Being lower than `min_app_version` won't allow the app to load at all, but being below `latest` can so a slight nag or something

PR #1268, opened 2015-11-07
[RFC] Determine Event Status

phil-lopreiato:
Based off #1257 
Working towards #1256 

Basic ways to determine if an individual event is down (e.g. not reporting data because FMS has no internet) or if the entire FMSAPI is down (throwing 5XX errors).

How does the base look?


gregmarra:
@phil-lopreiato  want to rebase this on top of #1257 ?

phil-lopreiato:
Rebased on master

PR #1270, opened 2015-11-13
Events From FMSAPI

phil-lopreiato:
Use FMSAPI for event list. Hits [Event Listing](http://docs.frcevents2.apiary.io/#reference/season-data/event-listings/retrieve-event-listings) endpoint. 

Current functionality maintained (e.g. skips over CMP events).

The only things missing from the API are the venue address and event website and I've submitted a feature request for them: https://usfirst.collab.net/sf/go/artf4727

This does not import EventTeams yet, that will be a separate PR

### DIFF
--- a/consts/event_type.py
+++ b/consts/event_type.py
@@ -3,6 +3,7 @@ class EventType(object):
     DISTRICT = 1
     DISTRICT_CMP = 2
     CMP_DIVISION = 3
+    CMP_SUBDIVISION = 5
     CMP_FINALS = 4
     OFFSEASON = 99
     PRESEASON = 100
@@ -13,6 +14,7 @@ class EventType(object):
         DISTRICT: 'District',
         DISTRICT_CMP: 'District Championship',
         CMP_DIVISION: 'Championship Division',
+        CMP_SUBDIVISION: 'Championship Subdivision',
         CMP_FINALS: 'Championship Finals',
         OFFSEASON: 'Offseason',
         PRESEASON: 'Preseason',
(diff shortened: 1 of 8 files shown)


consts/event_type.py
```diff
@@ -3,6 +3,7 @@ class EventType(object):
     DISTRICT = 1
     DISTRICT_CMP = 2
     CMP_DIVISION = 3
+    CMP_SUBDIVISION = 5
```
phil-lopreiato: Not sure if this is a good thing to add in for future use? Thoughts?

gregmarra:
This looks good to me once conflicts are resolved. Sorry @phil-lopreiato for taking so long to review.

phil-lopreiato:
Rebased on master

consts/event_type.py
gregmarra: Are there subdivisions now? I'm not sure the data model.

I think we should keep these enum lists in numerical order, so you can make sure you're not using a value already used.

gregmarra:
Cool, just my question on subdivisions, then I'll pull.

consts/event_type.py
phil-lopreiato: There (sorta) are. For last 2015/2016 CMP, there were 4 divisions, each with 2 "subdivision" to get the 8 "divisions". Some awards were done at the subdivision level, some at the division level. 

FMSAPI makes the distinction between division and subdivision, but we do all those manually anyway. So, I guess I should get rid of these for now, since the change to import CMP divisions will be better suited for a separate diff, anyways.

phil-lopreiato:
Should be all set

gregmarra:
Thanks!

PR #1282, opened 2015-11-23
Bumps Django version.

gregmarra:
Old Django is getting deprecated. https://cloud.google.com/appengine/docs/deprecations/django

Tested locally by browsing around to a few pages, they all worked. Tests pass.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -11,7 +11,7 @@ builtins:
 
 libraries:
 - name: django
-  version: "1.2"
+  version: "1.5"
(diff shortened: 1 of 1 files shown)


gregmarra:
@fangeugene @phil-lopreiato I read the Django changelog and don't think this will break anything. Any concerns?

phil-lopreiato:
This PR is on https://tba-dev-phil.appspot.com currently and I don't see anything broken 
:+1:

PR #1292, opened 2015-12-02
Datafeed rework

fangeugene:
Remove all unused datafeeds, add FIRSTElasticSearch datafeed.


phil-lopreiato:
So much simpler! :+1:

gregmarra:
Does this completely remove the FMS API datafeeds? Do we anticipate wanting to use them (official) over the ES queries (unofficial)?

fangeugene:
It keeps the FMS API and ES queries.
It removes the HTML datafeeds. We were only using 3 out of the dozen or so anyways, since they already got migrated to the FMS API.

gregmarra:
Love it.

PR #1298, opened 2015-12-02
Api Status

### LINKED ISSUE #1256: [Spec] Mobile/API Status

phil-lopreiato:
Implementation for https://github.com/the-blue-alliance/the-blue-alliance/issues/1256

Closes #1256

### DIFF
--- a/controllers/api/api_status_controller.py
+++ b/controllers/api/api_status_controller.py
@@ -6,7 +6,7 @@ from models.sitevar import Sitevar
 
 
 class ApiStatusController(ApiBaseController):
-    CACHE_KEY_FORMAT = "apiv2_status_controller"  # year
+    CACHE_KEY_FORMAT = "apiv2_status_controller"
     CACHE_VERSION = 0
     CACHE_HEADER_LENGTH = 61
 
@@ -26,6 +26,7 @@ class ApiStatusController(ApiBaseController):
 
     def _render(self):
         status_sitevar = Sitevar.get_by_id('apistatus')
+        fmsapi_sitevar = Sitevar.get_by_id('apistatus.fmsapi_down')
         down_events_sitevar = Sitevar.get_by_id('apistatus.down_events')
 
         # Error out of no sitevar found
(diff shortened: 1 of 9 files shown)


fangeugene:
This looks great except nothing is clearing the caches. Not sure what the best way to handle that is.

fangeugene:
Also, there might be concurrency issues when writing to the sitevars for fmsapi_down and down_events.

fangeugene:
I think `clear_cache_if_needed` shouldn't be taking in the json string but rather an object. Dict->JSON isn't deterministic so the same thing may end up being a different string.

controllers/api/api_status_controller.py
```diff
@@ -0,0 +1,51 @@
...
+        return json.dumps(status_dict, ensure_ascii=True)
+
+    @classmethod
+    def clear_cache_if_needed(cls, old_content, new_content):
+        """
+        Clears the cache associated with this response
+        Only clears if old_content != new_content (e.g. response changes)
+        """
+        if old_content != new_content:
+            cls.delete_cache_multi(cls.get_cache_key_from_format())
```
fangeugene: `cls.delete_cache_multi([cls.get_cache_key_from_format()])`

fangeugene:
Sweet looks great.

Issue #1162, opened 2015-04-02
API: Teams in a district/District Ranking


flyingcow93:
More of a request rather then an issue, currently you can grab /api/v2/district/<district>/<year>/rankings which generates a list of teams that have played at least one event in the district. This causes some issues:

-If a team goes out of district, they are now included in the other district as well (Example: 3314 (MAR) went to NE in 2015.)
-Teams that have not played a district yet are not listed at all, so a 'list of teams in x district in y year' list cannot be built.
-Events also seem to return in a random order, it would be nice if they were ordered either chronologically, or a week value was also returned for the event, or a 'play' value (1st play, 2nd play, 3rd play, would be nice so we could ignore 3rd plays), or all of the above were implemented.

I'm a bit new to figuring out how github works, but more then happy to help figure out a solution if someone could point me in the right direction.

flyingcow93:
Not exactly related to above but I'm trying to follow along with the readme to setup everything on my pc, having trouble with the 'paver setup' command. Says its missing a file, but doesn't specify which file as far as I can tell. Heres the output:

C:\Users\Brandon>cd C:\Users\Brandon\Dropbox\tba git\the-blue-alliance

C:\Users\Brandon\Dropbox\tba git\the-blue-alliance>paver setup
---> pavement.setup
---> pavement.make
---> pavement.javascript
Combining and Compressing Javascript
python do_compress.py js
Compressing Main JavaScript...
 + static/jwplayer/jwplayer.js
 + static/xcharts/d3.v2.min.js
 + static/xcharts/xcharts.min.js
 + static/javascript/tba_js/tablesorter.js
 + static/javascript/tba_js/tba_charts.js
 + static/javascript/tba_js/tba_countdown.js
 + static/javascript/tba_js/tba_sidebar.js
 + static/javascript/tba_js/tba_typeahead.js
 + static/javascript/tba_js/tba.js
=> static/javascript/tba_combined_js.main.min.js
Original: 303.81 kB
Compressed: 298.35 kB
Reduction: 1.8%

Compressing GameDay JavaScript...
 + static/jwplayer/jwplayer.js
 + static/xcharts/d3.v2.min.js
 + static/xcharts/xcharts.min.js
 + static/javascript/tba_js/tablesorter.js
 + static/javascript/tba_js/tba_charts.js
 + static/javascript/tba_js/tba_countdown.js
 + static/javascript/tba_js/tba_sidebar.js
[...]

PR #1300, opened 2015-12-06
added admin script to generate past district teams.

### LINKED ISSUE #1231: Create DistrictTeams for previous years

cmlicata:
Admin task to run at /admin/district_teams_update/<year> that will fetch all district events in the year,
and select the district each team competed in most often to use as the "home" district.

Closes #1231

### DIFF
--- a/controllers/admin/admin_cron_controller.py
+++ b/controllers/admin/admin_cron_controller.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from datetime import date
 import logging
 import os
@@ -8,10 +9,16 @@ from google.appengine.ext import ndb
 from google.appengine.ext.webapp import template
 
 from consts.client_type import ClientType
+from consts.district_type import DistrictType
 from controllers.base_controller import LoggedInHandler
+from helpers.district_team_manipulator import DistrictTeamManipulator
 from helpers.notification_sender import NotificationSender
+from models.district_team import DistrictTeam
+from models.event import Event
+from models.event_team import EventTeam
 from models.mobile_client import MobileClient
 from models.subscription import Subscription
+from models.team import Team
 from notifications.ping import PingNotification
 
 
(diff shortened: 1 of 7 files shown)


phil-lopreiato:
After this gets merged and run, we'll have all the data we need for #1162

controllers/admin/admin_cron_controller.py
```diff
@@ -128,3 +135,48 @@ def get(self):
...
+        new_district_teams = []
+        for team_key, districts in team_districts.iteritems():
+            most_frequent_district = max(set(districts), key=districts.count)
+            logging.info("Assuming team {} belongs to {}".format(team_key, DistrictType.type_names[most_frequent_district]))
+            dt_key = DistrictTeam.renderKeyName(year, most_frequent_district, team_key)
+            new_district_teams.append(DistrictTeam(id=dt_key, year=year, team=ndb.Key(Team, team_key), district=most_frequent_district))
+
+        logging.info("Finishing updating old district teams from event teams")
+        DistrictTeamManipulator.createOrUpdate(new_district_teams)
+
```
gregmarra: lots of extra linebreaks

gregmarra:
Can you provide some quick context on the purpose of this? The other issue doesn't have a concise summary.

phil-lopreiato:
@gregmarra this is a one time script to run that builds [DistrictTeam](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/models/district_team.py) models for previous years. We currently only have DistrictTeams for 2015 events, since that data was introduced in FMSAPI and we couldn't get it before.

This PR guesses at a team's home district by finding the district they competed at most frequently in each season and stores the proper DistrictTeam model.

cron_main.py
```diff
@@ -63,5 +62,6 @@
                                ('/tasks/admin/clear_old_subs', AdminSubsClear),
                                ('/tasks/admin/enqueue/clear_old_webhooks', AdminWebhooksClearEnqueue),
                                ('/tasks/admin/clear_old_webhooks', AdminWebhooksClear),
+                               ('/tasks/admin/rebuild_district_teams/([0-9]+)', AdminCreateDistrictTeamsDo),
```
gregmarra: should these be in the /math/ path? They mostly do math, vs administer user accounts.

cmlicata: I'm not sure, @phil-lopreiato told me to put them there. Do you think they should be somewhere else?  If so, please let me know to where you like me to move them.

gregmarra: I don't think we've been super consistent here. I guess since these are mostly triggered by admins, this is probably semantically correct.

gregmarra:
@fangeugene you've written a few other admin one-time jobs. Do you have any other feedback here? Otherwise I think this is good to pull.

@cmlicata is the "run once" here just to load the URL to enqueue these once?

cmlicata:
Yes, it's just a run once.

fangeugene:
Er, don't run this yet @gregmarra

gregmarra:
Roger that. I did push a build with this to prod, but won't run it.

PR #1273, opened 2015-11-23
Suggestion approval with permissions

gregmarra:
Builds on top of #1272 to start using permissions for reviewing suggestions.

This will be rebased on top of #1272 before merging.

PR #1304, opened 2015-12-12
Use Permissions for Suggestion approval.

gregmarra:
Moves suggestion reviews to be possible by anyone with the permission to MUTATE_DATA.

A clearer version of #1273 .


controllers/base_controller.py
```diff
@@ -117,6 +117,17 @@ def _require_login(self, target_url="/"):
                 abort=True
             )
 
+    def _require_permission(self, permission):
+        self._require_login()
+        logging.info("logged in")
+        self._require_registration()
+        logging.info("registered")
+        if permission not in self.user_bundle.account.permissions:
+            return self.redirect(
```
fangeugene: Is redirecting to the homepage the thing we want to do here? Not saying it's not, just wondering what other options could be considered.

synth3tk: 1. Redirect to homepage with a flash message stating they don't have permission
2. Redirect to previously-visited page (or homepage if no history) with flash message
3. Redirect to a page built specifically to tell the user that they're not authorized (like a 404 page)

There may be more, but those are a few.

gregmarra: I agree this should be improved, let's address it in a future pull. A common "you don't have permission to do that, here's how to ask for permission" screen seems good.

PR #1306, opened 2015-12-12
Move suggestion approval controllers

gregmarra:
Builds on #1304 to enable any user with appropriate permissions to accept suggestions!

Test plan:
* Create test teams, events, matches using buttons in admin panel
* Submit suggestions for a webcast, team media, and match video
* "Register" my test user at /account/register
* Approve my own suggestions


fangeugene:
Would it make sense to put the url at `/suggestion_review/...` to differentiate from `/suggest...`?

gregmarra:
@fangeugene I wanted to leave these paths similar to each other. I'm going to land this to unblock new changes people are working on. We can move these paths later if we want to tidy things up.

fangeugene:
Alright

PR #1310, opened 2015-12-12
Updated getting started docs to use https:// repo address

nwalters512:
`git://` repo addresses cannot be used for pushing (https://gist.github.com/grawity/4392747)


synth3tk:
:+1:

PR #1313, opened 2015-12-17
Send API Status-Related GCM Messages

### LINKED ISSUE #1303: Add Admin Message to API Status Endpoint
Will allow us to write a custom message (with expiration date) to the app somehome. Basically, this is the server side component for https://github.com/the-blue-alliance/the-blue-alliance-android/issues/418

Extends #1256 to also add `text` and `expiration` (UNIX timestamp, seconds since epoch) to the response, inside a sub-dict `message` (or `null` if no message available). Both will be input from `/admin/apistatus` and stored with the other data in the sitevar. 

```
{
  max_season: 2015,
  is_datafeed_down: false,
  android: {
    min_app_version: 123456
  },
  ios: {
    min_app_version: 1243
  },
  down_events: [
    2015test,
    2015down
  ],
  message: {
    "text": "blah",
    "expiration": 123456
  }
}
```

phil-lopreiato:
Adds a notification for API status sync as well as notifying when an event goes down

Fixes  #1303

### DIFF
--- a/controllers/api/api_status_controller.py
+++ b/controllers/api/api_status_controller.py
@@ -1,7 +1,7 @@
 import json
 
 from controllers.api.api_base_controller import ApiBaseController
-
+from helpers.notification_helper import NotificationHelper
 from models.sitevar import Sitevar
 
 
@@ -42,10 +42,13 @@ class ApiStatusController(ApiBaseController):
         return json.dumps(status_dict, ensure_ascii=True)
 
     @classmethod
-    def clear_cache_if_needed(cls, old_content, new_content):
+    def clear_cache_if_needed(cls, old_content, new_content, users_not_to_notify=[]):
         """
         Clears the cache associated with this response
         Only clears if old_content != new_content (e.g. response changes)
         """
         if old_content != new_content:
             cls.delete_cache_multi([cls.get_cache_key_from_format()])
+
+            # Send update notifications
+            NotificationHelper.send_status_sync(users_not_to_notify)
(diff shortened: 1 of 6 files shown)


controllers/api/api_status_controller.py
```diff
@@ -42,10 +42,13 @@ def _render(self):
         return json.dumps(status_dict, ensure_ascii=True)
 
     @classmethod
-    def clear_cache_if_needed(cls, old_content, new_content):
+    def clear_cache_if_needed(cls, old_content, new_content, users_not_to_notify=[]):
```
gregmarra: How are you anticipating users_not_to_notify to work? It seems like typically we'd want the opposite? Seems weird to specify a large list of users not to notify.

phil-lopreiato: Take a look here: https://github.com/the-blue-alliance/the-blue-alliance/pull/1313/files#diff-571c44c2d22fd07a2e30005f1e579956R123

This is where we send notifications for events down to users who have subscribed to that event. This will show a notification saying the event is down and clear the app-side cache of the API Status endpoint. A little farther down, we want the rest of the mobile clients to clear their cache, but I don't want to send duplicate notifications. Hence, the solution you see.

Can you think of a better way to do it? [Relevant app code](https://github.com/the-blue-alliance/the-blue-alliance-android/blob/master/android/src/main/java/com/thebluealliance/androidclient/gcm/GCMMessageHandler.java#L141)

PR #1334, opened 2016-01-10
pavement: Reformat file

mc10:
This makes indentation 4 spaces and converts all strings to be double quoted.

### DIFF
--- a/pavement.py
+++ b/pavement.py
@@ -9,7 +9,7 @@ path = path("./")
 
 @task
 def deploy():
-  sh("python deploy.py")
+    sh("python deploy.py")
 
 @task
 def javascript():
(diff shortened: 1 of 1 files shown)


gregmarra:
Looks good to me. How did you test this, perhaps by running paver preflight?

mc10:
Tbh I didn't really test this because I couldn't figure out how to set up Paver on my local machine.

gregmarra:
Tested by running `paver preflight`, verifying same output as before.

gregmarra:
@mc10 we should make the paver setup instructions better. What roadbumps did you hit?

mc10:
Does `paver setup` need to be run before `paver preflight`? Also, I ran into a "no `jinja2` module" error, but I guess that can be fixed with a simple install.

phil-lopreiato:
Yeah, we need to add `jinja2` into the readme so that it gets installed with other npm (I think? Or is it from pip?) dependencies.

gregmarra:
I'm editing it now :)

PR #1335, opened 2016-01-10
travis: Remove deprecated --use-mirrors

mc10:
It'll be removed in a later version of `pip`, so we might as well remove it now.


gregmarra:
Thanks!

PR #1341, opened 2016-01-17
Fix urls without scheme

### LINKED ISSUE #1339: Websites without HTTP not linking properly
example: http://www.thebluealliance.com/team/4990

phil-lopreiato:
Uses urlparse for urls. Fixes #1339


gregmarra:
I'd rather fix this by doing the same thing in the parsers at write time than patching it every place at read time. This will fix it for all callers and make the data more correct.

phil-lopreiato:
Good call, that's a better idea. I'll rebase this PR onto a new approach

phil-lopreiato:
@gregmarra look better? 

I didn't run this on my instance (having billing issues), but these classes should be test covered

gregmarra:
@phil-lopreiato  this approach looks good. Can you add tests that confirm that various kinds of malformed URLs will get corrected as expected?

* www.ourwebsite.com
* ourwebsite.com
* http://www.ourwebsite.com
* https://www.ourwebsite.com
* ftp://www.ourwebsite.com (idk the future's weird maybe?)

phil-lopreiato:
Tests updated, all seems well

gregmarra:
This looks good to me. @fangeugene any nits? Given this touches core data parsing code I want to make sure we're happy with it :)

PR #1352, opened 2016-01-31
Alphabetizes functions in pavement.py

gregmarra:
Tested by running `paver preflight`, everything worked.


gregmarra:
Subsumed by #1353

PR #1376, opened 2016-02-26
Add 2016 score breakdown

phil-lopreiato:
See http://tba.lopreiato.me/match/2016scmb_qm3 as an example

### DIFF
--- a/controllers/match_controller.py
+++ b/controllers/match_controller.py
@@ -16,17 +16,6 @@ class MatchDetail(CacheableHandler):
     CACHE_VERSION = 4
     CACHE_KEY_FORMAT = "match_detail_{}"  # (match_key)
 
-    defense_render_names = {
-        'A_ChevalDeFrise': 'Cheval De Frise',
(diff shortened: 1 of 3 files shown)


controllers/match_controller.py
```diff
@@ -16,6 +16,17 @@ class MatchDetail(CacheableHandler):
     CACHE_VERSION = 4
     CACHE_KEY_FORMAT = "match_detail_{}"  # (match_key)
 
+    defense_render_names = {
```
fangeugene: This year specific stuff (along with `add_defense_render_names`) is awkward in the default match controller... I'm trying to think of a better way to do this.

phil-lopreiato: Yeah, not a huge fan of this. I should case the year, at least

fangeugene: Oh, idea. Add a filter to `common/my_filters.py`.

fangeugene: Call it like `render_defense_name` or something.

PR #1378, opened 2016-02-27
Allow deleting references to Media. Fixes #1370

### LINKED ISSUE #1370: Build a way to delete media
There is at least one media on the wrong team's page. We should build "remove media" buttons for the admin page.

gregmarra:
Twice recently I've been asked to delete Medias. This allows it in the admin interface.

### DIFF
--- a/controllers/admin/admin_media_controller.py
+++ b/controllers/admin/admin_media_controller.py
@@ -26,6 +26,21 @@ class AdminMediaDashboard(LoggedInHandler):
         self.response.out.write(template.render(path, self.template_values))
 
 
+class AdminMediaDeleteReference(LoggedInHandler):
+    def post(self, media_key_name):
+        self._require_admin()
+
+        media = Media.get_by_id(media_key_name)
+
+        media.references.remove(media.create_reference(
+            self.request.get("reference_type"),
+            self.request.get("reference_key_name")))
+
+        media.put() ## Todo, update manipulators to support removing an item that ends a list
+
+        self.redirect(self.request.get('originating_url'))
+
+
 class AdminMediaAdd(LoggedInHandler):
     def post(self):
         self._require_admin()
--- a/controllers/admin/admin_media_controller.py
+++ b/controllers/admin/admin_media_controller.py
@@ -35,8 +35,9 @@ class AdminMediaDeleteReference(LoggedInHandler):
         media.references.remove(media.create_reference(
             self.request.get("reference_type"),
             self.request.get("reference_key_name")))
+        media.dirty = True # hacky -greg 20150302
 
-        media.put() ## Todo, update manipulators to support removing an item that ends a list
+        MediaManipulator.createOrUpdate(media, auto_union=False)
 
         self.redirect(self.request.get('originating_url'))
 
--- a/templates/admin/team_details.html
+++ b/templates/admin/team_details.html
@@ -92,11 +92,40 @@
   {% else %}
     <h3>{{year}}:</h3>
   {% endif %}
+  <ul>
   {% for media in medias %}
-    {% include "partials/admin_media_display.html" %}
+    <li>{% include "partials/admin_media_display.html" %}</li>
   {% endfor %}
-  <br>
+  </ul>
 {% endfor %}
+
+<div class="modal fade" id="confirm-delete" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
+    <div class="modal-dialog">
+        <div class="modal-content">
+            <div class="modal-header">
+                Confirm media reference removal
+            </div>
+            <div class="modal-body">
+                Are you sure you want to remove this reference to this media? References to other models will persist.
+            </div>
+            <div class="modal-footer">
+                <form id="delete_media_form" action="/admin/media/delete_reference" method="post">
+                <input name="reference_type" type="hidden" value="team" />
+                <input name="reference_key_name" type="hidden" value="{{team.key_name}}" />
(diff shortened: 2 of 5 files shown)


controllers/admin/admin_media_controller.py
```diff
@@ -26,6 +26,21 @@ def get(self):
...
+    def post(self, media_key_name):
+        self._require_admin()
+
+        media = Media.get_by_id(media_key_name)
+
+        media.references.remove(media.create_reference(
+            self.request.get("reference_type"),
+            self.request.get("reference_key_name")))
+
+        media.put() ## Todo, update manipulators to support removing an item that ends a list
```
fangeugene: Use `MediaManipulator.createOrUpdate()` or else cache won't properly get cleared.

templates/admin/team_details.html
```diff
@@ -92,11 +92,40 @@
...
+
+                <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+                <button class="btn btn-danger" type="submit"><span class="glyphicon glyphicon-delete"></span> Remove reference</button>
+            </div>
+        </div>
+    </div>
+</div>
+
+<script>
+$('#confirm-delete').on('show.bs.modal', function(e) {
```
fangeugene: This modal is cool. We should use it in more places when confirming actions (right now we redirect to separate pages and it's clunky).

controllers/admin/admin_media_controller.py
gregmarra: I tried to do this and hit a roadblock that's going to require changing `BaseManipulator`.

To get a manipulator to *remove* items from a list (like `Media.references`), you have to call `MediaManipulator.createOrUpdate(...,auto_union=False)`. CreateOrUpdate ends up calling in to `MediaManipulator.updateMerge()`, where the default behavior is to merge lists. auto_union false will replace the old list with the new list.

However, then the `list_attr` comparer section only replaces a list if the new list has `len > 0`: https://github.com/the-blue-alliance/the-blue-alliance/blob/master/helpers/media_manipulator.py#L46

To get around this, we can:

1. Modify `MediaManipulator` to replace populated list_attr's with empty list list_attrs all the time
2. Modify `MediaManipulator` to replace populated list_attr's with empty list list_attrs when `auto_union == False`
3. Modify `ManipulatorBase` to thread a `consider_empty_lists` flag through all the method calls

I am not sure where else in code we may be relying on Manipulators to not clobber populated lists, so I am apprehensive to do anything except the 3rd option, but that felt clunky as I started to do it.

Any better ideas?

fangeugene: I agree with the 3rd option, since it's less likely to break anything.

How about extending that beyond lists and pass a `force_override` flag through everything that bypasses all the `!= None`, `len > 0`, etc. checks? Seems like it would be useful if we ever want to set anything to be null.

gregmarra: I will add a parameter called `allow_deleting_data` to make it really explicit what you're doing when you set it :)

fangeugene: Sounds scary, which is good!

gregmarra: I'm running into a weird problem.

When I delete the reference here, then go to do a `MediaManipulator.createOrUpdate()`, the "old model" the MediaManipulator is getting back from its internal ndb call already has `references = []`, even though the only place that mutation has happened is against the media object I am attempting to write back to the database. It's like the "uncommitted changes" are leaking back in to the DB.

I can tell this is happening because if I comment out the reference removal, the objects come back with references.

### Comment out removal
```
old_models = ndb.get_multi([ndb.Key(type(model).__name__, model.key_name) for model in new_models])
print "ze old models %s" % old_models

>> ze old models [Media(key=Key('Media', 'youtube_dQw4w9WgXcQ'), created=datetime.datetime(2016, 2, 27, 22, 50, 10, 257700), details_json=None, foreign_key=u'dQw4w9WgXcQ', media_type_enum=0, private_details_json=None, references=[Key('Team', 'frc1')], updated=datetime.datetime(2016, 3, 3, 5, 1, 21, 456624), year=2016)]
```

### Leave Removal In
```
old_models = ndb.get_multi([ndb.Key(type(model).__name__, model.key_name) for model in new_models])
print "ze old models %s" % old_models

ze old models [Media(key=Key('Media', 'youtube_dQw4w9WgXcQ'), created=datetime.datetime(2016, 2, 27, 22, 50, 10, 257700), details_json=None, foreign_key=u'dQw4w9WgXcQ', media_type_enum=0, private_details_json=None, references=[], updated=datetime.datetime(2016, 3, 3, 5, 1, 21, 456624), year=2016)]
```

Either way, the Manipulator decides the object isn't dirty and doesn't bother putting it back in the database. If I manually add `old_media.dirty = True` to the bottom of the `MediaManipulator.updateMerge()` method, it does successfully wipe itself.

Why am I pulling the already-modified media out the ndb with that query? :/ Is this a dev server only problem, or will it exist in prod?

fangeugene: "When I delete the reference here, then go to do a MediaManipulator.createOrUpdate(), the "old model" the MediaManipulator is getting back from its internal ndb call already has references = [], even though the only place that mutation has happened is against the media object I am attempting to write back to the database."

I believe this is the case due to instance caching. It's the same reason we have `mode.dirty = True  # this is hacky -greg` in some places.

gregmarra:
Thanks for the pointer on instance caching. Just borrowed the old approach here.

Manipulator is definitely a class that could use some refactoring. There is a ton of copypasta code across them all.

fangeugene:
This looks good to merge.

PR #1381, opened 2016-02-29
Suggestion home

gregmarra:
1. Creates a `SuggestionFetcher` helper class to avoid duplicating code between controllers.
2. Creates `SuggestHomeReviewController` to create a standalone page to review suggestions for people with the `MUTATE_DATA` permission who can't check the /admin/ page.
3. Refactor pending suggestion partial out into separate partial.

Problem I ran into: Couldn't put this partial above the admin directory, because the admin template can't reference "up a directory". Maybe Jinja2 templates can fix this? Didn't need to fix it for this.

### DIFF
--- a/controllers/admin/admin_main_controller.py
+++ b/controllers/admin/admin_main_controller.py
@@ -29,9 +29,10 @@ class AdminMain(LoggedInHandler):
         users = Account.query().order(-Account.created).fetch(5)
         self.template_values['users'] = users
 
-        self.template_values['video_suggestions'] = SuggestionFetcher.count(Suggestion.REVIEW_PENDING, "match")
-        self.template_values['webcast_suggestions'] = SuggestionFetcher.count(Suggestion.REVIEW_PENDING, "event")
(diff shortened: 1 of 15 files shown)


gregmarra:
I should have done the typo fix as a separate pull request, but oh well. This is good to review.

fangeugene:
This looks good. Maybe we should add a link in each suggestion page to go back to the home page?

gregmarra:
Yea, I regret calling some things `Suggest` and some things `Suggestion`, probably will clear that up over time.

I agree adding links back o the portal seems good. I think we should also build sections into your profile page if you're a data admin to link to this page, so you don't need to remember a URL...

gregmarra:
@fangeugene added some bootstrap breadcrumbs.

image

fangeugene:
This looks good. Exposing the url to those with permissions is the next step. Maybe we should do that in another PR and clean up the Profile page while we're at it?

gregmarra:
Let's do that in a separate PR to make profile pages great.

PR #1389, opened 2016-03-04
Team Social Media!

phil-lopreiato:
Same as #1320, except is uses the existing team `Media` suggestion flow to store social connections. `Media`s are not stored with year, only referenced by team key. This doesn't display them yet on team detail pages, but that's still the next step.

There is also opportunity to refactor the partial dict creation in `MediaHelper`, but I'm not sure if that's in scope for this pull. It's more extensible to move the existing media types to the more generic way I have it currently.

screenshot from 2016-03-04 16-35-28
screenshot from 2016-03-04 16-37-10


gregmarra:
Are there little widgets from these social networks we can embed to make it
so you don't need to click to confirm the data seems right?

phil-lopreiato:
I can add the basic profile plugins. Do you think they'll actually be useful to have, they don't seem to really convey a whole lot of info, because we'd just add the simple versions (not full feeds or anything)

FB: https://developers.facebook.com/docs/plugins/page-plugin
Twitter: https://dev.twitter.com/web/embedded-timelines
GitHub: https://github.com/lepture/github-cards
Couldn't find anything for YouTube...

gregmarra:
Helps make sure it's not a typo and faster than opening the profile in a
new tab to check. The YouTube video embeds really help for video
suggestions.

phil-lopreiato:
I've updated this pull to preview social media accounts (except for YouTube - I couldn't find a way to do it)

yt_suggest
twitter_suggest
gh_suggest
fb_suggest

helpers/media_helper.py
```diff
@@ -24,17 +24,44 @@ def group_by_slugname(cls, medias):
     def get_images(cls, medias):
         return [media for media in medias if media.media_type_enum in MediaType.image_types]
 
+    @classmethod
+    def get_socials(cls, medias):
+        return [media for media in medias if media.media_type_enum in MediaType.social_types]
```
fangeugene: Can this be done with a filter more succinctly?

fangeugene:
Can you throw in a bunch of parser tests?

fangeugene:
Medias arent showing up for me under suggestion review. any ideas?

phil-lopreiato:
I'll test it out merged with master and take a look

fangeugene:
You should probably strip query params.

image

Include it in the tests.

fangeugene:
Can we put social media review under a separate suggestion review?
That way we don't have to special case "preferred" things like this:

image

phil-lopreiato:
Yeah, sure, I can break it out into its own review page

phil-lopreiato:
Social media is now reviewed on its own page. More tests incoming...
screenshot from 2016-05-05 16-18-52

phil-lopreiato:
And added tests

helpers/suggestions/suggestion_creator.py
```diff
@@ -14,7 +14,12 @@ class SuggestionCreator(object):
     def createTeamMediaSuggestion(cls, author_account_key, media_url, team_key, year_str, private_details_json=None):
         """Create a Team Media Suggestion. Returns status (success, suggestion_exists, media_exists, bad_url)"""
 
-        media_dict = MediaParser.partial_media_dict_from_url(media_url.strip())
+        # Sanitize input url
+        media_url = media_url.strip()
+        if '?' in media_url:
```
fangeugene: Is there a better way to do this with some existing library? This doesn't deal with `#`

phil-lopreiato: Looks like we can handle this more cleanly with `urlparse`

fangeugene:
LGTM

technonerdz:
There is a way to embed the latest videos of a youtube channel to the Admin pannel to make it easier. The Script is: `<script type="text/javascript" src="http://ytchannelembed.com/ytce-frgallery.js"></script><div style="width:100%;max-width:420px;" class="ytce" data-user="UCLm-jhnXUfkqV_NTTMgvasg" data-rows="2" data-ratio="hd" data-width="130" data-margin="15" data-desc="100" data-desc_color="9E9E9E" data-title="30" data-title_color="000000" data-https="0" data-duration="1" data-views="0" data-likes="0" data-dislikes="0" data-fav="0" data-cols="3" ></div>`

It looks like that for my team:
evoembed

fangeugene:
@technonerdz that's great! Opened an issue in #1551.

PR #1395, opened 2016-03-06
[updated] Make event alliances endpoint cronjob run more often

phil-lopreiato:
Takes #1239 and implemented suggestion from review.

Closes #1239 also


controllers/datafeed_controller.py
```diff
@@ -105,9 +105,13 @@ class FMSAPIEventAlliancesEnqueue(webapp.RequestHandler):
     Handles enqueing getting alliances
     """
     def get(self, when):
+        events = []
```
gregmarra: Why is this needed? `events` will always ben set in the if below.

phil-lopreiato: This does not appear to be needed, want me to remove it? (I didn't write the original PR)

fangeugene: Yeah remove it.

phil-lopreiato:
All set

PR #1405, opened 2016-03-10
README Updates

brycematsuda:
Added some Linux specific setup instructions. Also fixed the indentation in the mydeploy.sh code block so it displays correctly.


synth3tk:
Hadn't realized that this was sitting here. :+1:

PR #1414, opened 2016-03-16
Show favorites in match tables

fangeugene:
DO NOT MERGE YET!

In order to preserve caching, this uses AJAX to fetch a user's favorites (much like GameDay does) and uses the favorites to customize the user's viewing experience. Favorites are stored on the user's device in a cookie with a 5 min expiration to prevent hammering the server.

I'm not too sure about what the best way to highlight favorited teams in matches. I'd like people's input.

The current implementation puts a star next to the match name if any team in that match is a favorite. Hovering over the star provides a tooltip saying: "A Favorited team is playing in this match!"
This seems to be the cleanest, but I feel like the user still has to visually search through the match to figure out which of their favorite teams are playing. Also, we technically support favorite matches too, so how do we distinguish that from this?
image

Another idea is to add borders to the cells of favorited teams, which makes favorites more obvious, but seems more cluttered. Also, this leaves no clear hover location for a tooltip.
image
In yellow: 
image

Using small dots instead of borders reduces clutter, but also reduces visibility (despite trying various colors). The dots provide a nice tooltip hover spot.
image


nwalters512:
I like either the stars or dots. They're more versatile; in the future, they could be used throughout the site to indicate favorite things where borders wouldn't make sense. They'd also be easier to carry over to the app than borders.

I feel like the stars are the best overall. "yellow star = favorite thing" is pretty universal iconography, whereas it won't be immediately obvious what a dot means. On mobile (both web and app), where we don't have the benefit of hovering tooltips, it might take a while for users to figure out "dot = favorite", if they even do figure that out.

synth3tk:
I definitely agree on the stars for the reasons listed by @nwalters512. It'll become immediately clear that they're favorites to just about most people, and for everyone else you have the tooltip. With it being yellow, it stands out a lot better than the dots could.

However, it would be nice to remind them of just who it is they favorited. Let's say I have 30 favorite teams. Well is it one team in the match? Three? Which ones are they? Maybe use a combination of the star (to pull attention towards the specific match(es)) as well as something for each individual cell. Maybe the shading itself should change? Text color? Use the dot?

fangeugene:
@synth3tk Oh i love the idea of the star + dots. Let me try that

tytremblay:
Stars as dots?

fangeugene:
@tytremblay I think the problem with dot-like-things is that they're so hard to notice.

fangeugene:
Stars and dots
image

phil-lopreiato:
That last one looks great!

synth3tk:
Agreed. Doesn't add too much clutter to the already busy table. :+1:

gregmarra:
This is very cool.

PR #1415, opened 2016-03-16
Prevent team media review page from scrolling to top

nwalters512:
The "accept all"/"reject all"/"deselect all" buttons use `href="#"`, which scrolls the page to the top when the element is clicked. This changes the javascript click handlers to prevent that default action.


fangeugene:
I've always wanted to fix this.

Issue #1225, opened 2015-06-22
iframes on API docs page are rendered


synth3tk:
But they shouldn't!

![tba-iframe](https://cloud.githubusercontent.com/assets/2099988/8292880/e845ea60-18ff-11e5-8a14-115f5ee80380.png)

gregmarra:
Oh jeez, we should be escaping the display of this for sure, otherwise we're exposing ourselves to XSS attacks.

PR #1418, opened 2016-03-17
Escape HTML tags in API docs.

brycematsuda:
For consistency's sake (and maybe in case html is put in other places in the future), I enclosed the .text(...).html(); around all of the JSON.stringify's used in the api docs page. 

It seemed to work okay when I put the iframe code used in #1225 in a test event.

iframe-test


synth3tk:
You're awesome. I never got around to setting up my dev environment again.

Issue #1224, opened 2015-06-21
Offseason Rankings Shouldn't Link to FIRST Event Page


phil-lopreiato:
Probably not unless the FIRST EID is set

nightpool:
looks like this didn't get closed by my commit message. can someone close?

PR #1425, opened 2016-03-25
Fix the link to the FIRST event page

### LINKED ISSUE #1417: FIRST event page link outdated
The link to "FIRST's Event page" on the top of some pages (like https://www.thebluealliance.com/event/2016scmb#rankings) is outdated. 

The new links are formatted with year and event code (such as http://frc-events.firstinspires.org/2016/SCMB)

Thank you.

nightpool:
Done on the github web editor, so sorry if the formatting comes out weird. Also stops the link from showing up for non-official events.

This may not work for Championship divisions, which seems to use a brand new identifier, i.e. `http://frc-events.firstinspires.org/2016/ARCHIMEDES`. Not sure I have a good enough handle on TBAs codebase to decide what to do here.

Fixes #1417 and #1224

### DIFF
--- a/templates_jinja2/event_details.html
+++ b/templates_jinja2/event_details.html
@@ -174,7 +174,9 @@
     <div class="tab-pane" id="rankings">
       <div class="row">
         <div class="col-sm-8 col-sm-offset-2 col-md-offset-2 col-lg-offset-2">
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
LGTM. Thanks for the fix!

I wouldn't worry about the Championship divisions yet, we usually end up special-casing those anyway.

synth3tk:
LGTM! :+1:

PR #1450, opened 2016-04-01
Add simple webcast url validation

phil-lopreiato:
Disallows submitted urls with spaces and also TBA urls. I see these kind
of invalid url submitted in the queue pretty often


fangeugene:
This will help a lot

PR #1466, opened 2016-04-09
Add Vimeo video support

### LINKED ISSUE #916: Vimeo video support for Media

synth3tk:
## WIP - DO NOT MERGE

This adds support for submitting and approving Vimeo videos, currently only for matches.

Still needs:
- [ ] Team media implementation
- [ ] Test coverage (need help in this area)

Closes #916

### DIFF
--- a/controllers/suggestions/suggest_match_video_controller.py
+++ b/controllers/suggestions/suggest_match_video_controller.py
@@ -2,6 +2,7 @@ import logging
 import os
 
 from controllers.base_controller import LoggedInHandler
+from helpers.media_helper import MediaParser
 from helpers.suggestions.suggestion_creator import SuggestionCreator
 from helpers.youtube_video_helper import YouTubeVideoHelper
 from models.event import Event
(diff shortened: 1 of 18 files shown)


controllers/suggestions/suggest_match_video_controller.py
```diff
@@ -42,11 +43,26 @@ def post(self):
...
-        status = SuggestionCreator.createMatchVideoYouTubeSuggestion(self.user_bundle.account.key, youtube_id, match_key)
-
+        video_url = self.request.get("video_url")
+        
+        if any(s in video_url for s in MediaParser.YOUTUBE_URL_PATTERNS):
+            youtube_id = YouTubeVideoHelper.parse_id_from_url(video_url)
+
+            status = SuggestionCreator.createMatchVideoYouTubeSuggestion(self.user_bundle.account.key, youtube_id, match_key)
+        elif any(s in video_url for s in MediaParser.VIMEO_URL_PATTERNS):
+            vimeo_id = None
```
gregmarra: This logic is pretty complicated. It feels like we should try to move the logic for searching for suggestions inside of something else so we can have clean, testable functions for each regex searcher.

helpers/match_manipulator.py
```diff
@@ -119,7 +119,8 @@ def updateMerge(self, new_match, old_match, auto_union=True):
 
         auto_union_attrs = [
             "tba_videos",
-            "youtube_videos"
+            "youtube_videos",
+            "vimeo_videos"
```
gregmarra: I don't love that this is extending models. I want to try to move to a world where Match videos are also Media objects, and Matches just have a set of Media attached to them.

templates/suggest_match_video_review_list.html
```diff
@@ -33,7 +33,12 @@
                     </div>
                     <div class="col-xs-7 fitvids">
                         <p><a href="/match/{{ suggestion.target_key }}" target="_blank"><strong>{{ suggestion.target_key }}</strong></a></p>
-                        <iframe width="240" height="180" src="https://www.youtube.com/embed/{{suggestion.youtube_video|yt_start}}" frameborder="0" allowfullscreen></iframe>
+                        {% if suggestion.youtube_video %}
+                          <iframe width="240" height="180" src="https://www.youtube.com/embed/{{suggestion.youtube_video|yt_start}}" frameborder="0" allowfullscreen></iframe>
+                        {% endif %}
+                        {% if suggestion.vimeo_video %}
+                          <iframe src="https://player.vimeo.com/video/{{suggestion.vimeo_video}}?byline=0&portrait=0" width="240" height="180" frameborder="0" allowfullscreen></iframe>
```
gregmarra: We should use a partial for each of these so we can update it everywhere if the embed code ever changes.

gregmarra:
This is a good first pass at this, though there are some places we should clean things up.

In general, I think we'd be better served by moving videos to Media objects, instead of adding more model properties to keep track of each provider type. I think we're going to want Facebook videos soon as well, and it seems weird to have a new model property for each kind of video.

synth3tk:
Yeah, I thought it was kind of messy too. I was trying to get quick Vimeo
support, but a rewrite now is probably the better option.

helpers/match_manipulator.py
phil-lopreiato: I agree, especially if we want to ship #1389 and link Medias to other models in the future.

controllers/suggestions/suggest_match_video_controller.py
phil-lopreiato: I tried to simplify the code I added for the social links, you can see how I did it here: https://github.com/the-blue-alliance/the-blue-alliance/pull/1389/files#diff-40f412c989d5abffbb74ac686847c330R61

synth3tk:
Gonna have to update this to reflect the refactored media stuff.  I'll try to get the tests in then.

PR #1469, opened 2016-04-10
Fixed API Link

rselwyn:
The API changed name, which caused the link to change.


synth3tk:
Just curious, but any particular reason for dropping the Java suffix?

rselwyn:
Yes.  It isn't part of the name of the API.  The API is called "TheBlueAlliance"

Issue #1468, opened 2016-04-10
District Championship Points Calculated Incorrectly


caitlynb:
Hello,

I am noticing that the district championship points are not correlating with FIRST's official rankings.  

On the event webpage for the championship (http://www.thebluealliance.com/event/2016chcmp#district-points), the total district points appears correct.  However, those total district points are not being carried into the district ranking page (http://www.thebluealliance.com/events/chs/2016#rankings).   This results in the district rankings being out of line with FIRST official rankings.  

Thanks for addressing this,
Caitlyn

phil-lopreiato:
This looks like another page caching problem if the event-based points display correctly. TBA only calculates district points once per day (which we should fix, see #1457), so I'll bet that the points shown on the rankings page are about 24 hours old. If it doesn't fix itself in a day or so, we'll take a look further.

It's on our offseason plan to rework how a lot of this is done on the backend to avoid this kind of confusion.

gregmarra:
We could also write "Last updated time.time()" on the page (all pages?),
and just admit that the page is stale.

phil-lopreiato:
That sounds like a great short-term fix

PR #1473, opened 2016-04-12
Show when pages are rendered

phil-lopreiato:
To help avoid some cache-based confusion (we just admit to stale data)
Suggested by @gregmarra in #1468

screenshot from 2016-04-11 22-55-44

### DIFF
--- a/controllers/base_controller.py
+++ b/controllers/base_controller.py
@@ -1,5 +1,6 @@
 import cPickle
 import logging
+import time
 import webapp2
 import zlib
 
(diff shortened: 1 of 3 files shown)


controllers/base_controller.py
```diff
@@ -52,6 +53,7 @@ def get(self, *args, **kw):
         if cached_response is None:
             self._set_cache_header_length(self.CACHE_HEADER_LENGTH)
             self.template_values["cache_key"] = self.cache_key
+            self.template_values["render_time"] = time.strftime('%b %d %Y at %l:%M%p %Z')
```
fangeugene: Can we make this  `%b. %d, %Y`?

nwalters512:
Nice! The Jinja template should probably say "on" instead of "at".

controllers/base_controller.py
phil-lopreiato: Fixed

PR #1482, opened 2016-04-20
Gameday2 update

nwalters512:
This is still very much a work in progress and is not ready to be merged. But while I'm working on this, I did want to get some thoughts on the tooling used to make React work nicely.

The major change so far is the use of a gulp pipeline to transform the source JSX file into what the client will use. This is much more efficient than shipping the raw JSX to the client and compiling it there.

With gulp, we can also use tools like browserify to use node.js-style imports in the browser. This gives us modern dependency management with React and other libraries the frontend needs.

However, I know the existing build pipeline uses Paver, and mixing gulp and Paver feels a little weird to me. Looking around, there seem to be some Python options if we wanted to keep all the build stuff in python, namely [webassets](http://webassets.readthedocs.org/en/latest/). That being said, I've sort of fallen in love with gulp's way of doing things. @gregmarra @fangeugene @phil-lopreiato any thoughts on how this should be handled?


phil-lopreiato:
It's all Greek to me! :stuck_out_tongue:

nwalters512:
I'll do you one better: full ES6 support with [babel](https://babeljs.io/)!

nwalters512:
I was scared that React would destroy and recreate the DOM nodes containing the embeds whenever the layout was changed, meaning we'd lose the players and any state they had. Luckily React is smart enough to save and reuse them! I'm loving this library :smile:

gregmarra:
@nwalters512 I would try to break this into fewer smaller commits over time, or else you're going to end up with something monolithic and hard to explain / review.

@fangeugene has done a lot of the javascript work on TBA and would be the best reviewer, but I don't think is familiar with React. I started to do the GameDay2 React implementation a long time ago, but I'm sure React has evolved since then and I am definitely not a JS expert. 

It'd be great to get someone else who can review these diffs, whether that is @fangeugene , me, or someone else. I'm excited about the idea of porting TBA JS into React or some other framework to make it more manageable, especially for the big complicated GameDay feature that's our most popular part of the site :)

nwalters512:
@gregmarra for sure, a lot of that monolithic stuff is hard to avoid up front though. I'm mostly working on getting a solid base to build on right now. React and the associated tooling has evolved a lot since the last time this was touched, so most of this work is to account for that.

Frankly, I'd probably be happy merging this soon. It has all the basic functionality (hiding/showing panels, adding/removing webcasts) working again, and gives us something to iterate on. Integrating tooling nicely is my main concern right now.

gregmarra:
Can you explain more why you don't like running gulp through paver?

nwalters512:
Looking back at paver, it seems like I misunderstood its role. As it is, LESS is already processed with, well, less, which is itself a node module. So it probably isn't so bad.

If we start to transition more parts of the site to React/etc., maybe it would make sense to do more stuff in gulp to have a more unified assets pipeline?

One question I had (I know nothing about this though, so I'll ask) was if the paver scripts are run when the app is deployed to appengine to make sure all the generated assets are up to date. Or is it up to the deploying individual to make sure that's done? If it is done on the server, is GAE smart enough to install the required node modules? It seems like it would already have to do that with LESS.

gregmarra:
We use the `paver preflight` command, which generates all of the required
assets. I believe `paver push` does this now too, now that you can't push
with the GoogleAppEngine app anymore.

nwalters512:
Good to know. My gulp task is in the paver `make` task, so that should work as expected.

phil-lopreiato:
Just skimming through the diff, it's very readable. React is cool! I'm a fan!

11408322_909862675736523_1138726707_n

#ShipIt 

(btw, @gregmarra, how do you manage to use GitHub and not miss Phabricator's image macros?)

nwalters512:
Whoever wants to try and review this, go ahead!

A few things to note:
* Make sure Node and npm are installed, then run `npm install && gulp build` to get everything installed and built for testing.
* Instead of using `lessc` and `do_compress.py` to process gameday2's `.less` files, I added that to the gulp build pipeline. The way I see it, gameday is almost a separate product from the main site, so handling its resources a bit differently is ok, and I'd rather keep all of gameday's tooling in one place anyways.
* Video layout now is done entirely with CSS instead of manual positioning in code or trying to use Bootstrap's grid. Manual positioning is gross, and because of how React responds to changes in its DOM, all the `VideoCell` components must be direct children of `VideoGrid`.
* I'd like to eventually write tests with [Jest](https://facebook.github.io/jest/), but this PR is getting unweildy as it is. That can be done in the future.

phil-lopreiato:
I took a stab at getting this up and running...
 - Make sure the docs are updated so people can easily figure out how to install the new dependencies
 - After I ran `npm install`, I was unable to run `gulp` without manually installing the following dependencies: `gulp-debug` and `gulp-uglify`. These should be added to `package.json`?
 - You'll want to add the following to the `skip_files` block in `app_shared.yaml` so it doesn't try and deploy *all* the local node modules to app engine (they're not needed)
```
- ^(.*/)?.*/node_modules/.*$
```

Aside from that, I was able to get it running. Try it out! http://tba.lopreiato.me/gameday2

nwalters512:
Good catches! I also excluded `react/*` from being deployed; the only things that matter are the compiled files, which are generated and placed in `static`.

nwalters512:
Every React component now lives in its own file. This should hopefully make things easier to review and makes the structure much more modular for future development.

phil-lopreiato:
Last bit of feedback: I think that running `npm install` should be integrated into the paver pipeline somehow, so that people don't have to remember to update node modules, should dependencies get updated and such

nwalters512:
That should be better. I used `npm update` instead of `npm install` though; the user should have done the install step during project setup.

phil-lopreiato:
LGTM :+1:

fangeugene:
The general setup looks good here, going to merge it now before it gets too bloated.

PR #1483, opened 2016-04-20
Team Tracker

phil-lopreiato:
Adds a tab to the district overview showing teams currently competing (by event) and their current status. Code is shared with the nigthbot status generation.

I want to do this for myTBA favorites as well, but I'll put this up for now.

screenshot from 2016-04-20 15-24-12

Closes #1481

PR #1484, opened 2016-04-20
MyTBA Team Tracker

phil-lopreiato:
This depends on #1483. One that gets merged, I'll rebase this onto the resulting commit. I wish GitHub did PR-dependencies better....

Found at `/mytba`

screenshot from 2016-04-20 17-25-03


controllers/mytba_controller.py
```diff
@@ -0,0 +1,42 @@
...
+
+        live_eventteams_futures = []
+        for event in live_events:
+            live_eventteams_futures.append(EventTeamsQuery(event.key_name).fetch_async())
+
+        favorite_teams = [team_future.get_result() for team_future in favorite_teams_future]
+        live_events_with_teams = EventTeamStatusHelper.buildEventTeamStatus(live_events, live_eventteams_futures, favorite_teams)
+
+        self.template_values.update({
+            'year': year,
```
fangeugene: I believe you can just do event.year in the template instead of doing the datetime thing here.

controllers/mytba_controller.py
```diff
@@ -0,0 +1,42 @@
...
+            'live_events_with_teams': live_events_with_teams,
+        })
+
+        path = os.path.join(os.path.dirname(__file__), '../templates/mytba_live.html')
+        self.response.out.write(template.render(path, self.template_values))
```
fangeugene: Can you make this a jinja2 template?

fangeugene: Nevermind I see you're reusing the template partials.

controllers/mytba_controller.py
fangeugene: Make sure you change it in the districts controller too.

PR #1485, opened 2016-04-21
myTBA on Navbar & Cleanup Login Flow

fangeugene:
Don't merge until we have a better empty landing page for `/mytba` or Champs starts (since it won't be empty then.)

Instead of kicking a user straight to the Google login page when visiting a page that requires login, we now show the following page:
image

Also did a bunch of refactoring so that 
`self._require_login()` and `self._require_registration()` are much more intuitive to use.

### DIFF
--- a/controllers/suggestions/suggest_event_webcast_controller.py
+++ b/controllers/suggestions/suggest_event_webcast_controller.py
@@ -13,7 +13,7 @@ class SuggestEventWebcastController(LoggedInHandler):
     """
 
     def get(self):
-        self._require_login("/suggest/event/webcast?event=%s" % self.request.get("event_key"))
+        self._require_login()
 
         if not self.request.get("event_key"):
             self.redirect("/", abort=True)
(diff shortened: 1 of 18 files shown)


controllers/suggestions/suggest_event_webcast_controller.py
```diff
@@ -13,7 +13,7 @@ class SuggestEventWebcastController(LoggedInHandler):
     """
 
     def get(self):
-        self._require_login("/suggest/event/webcast?event=%s" % self.request.get("event_key"))
```
gregmarra: Does this mean this controller no longer redirects back to itself after login?

fangeugene: It does. `_require_login()` will always redirect to `self.request.url` unless otherwise specified. I see no use case for it being specified since everywhere it was being used was redirecting back to itself but I kept it in just in case.

PR #1488, opened 2016-04-21
Clean Up MyTBA Page

phil-lopreiato:
Now shows upcoming events as well!
screenshot from 2016-04-21 19-00-38

### DIFF
--- a/common/my_filters.py
+++ b/common/my_filters.py
@@ -35,8 +35,3 @@ def yt_start(value):
 @register.filter
 def get_item(dictionary, key):
     return dictionary.get(key)
-
-
-@register.filter
-def strftime(d, str):
-    return d.strftime(str)
--- a/common/my_filters.py
+++ b/common/my_filters.py
@@ -35,3 +35,8 @@ def yt_start(value):
 @register.filter
 def get_item(dictionary, key):
     return dictionary.get(key)
+
+
+@register.filter
+def strftime(d, str):
+    return d.strftime(str)
--- a/controllers/mytba_controller.py
+++ b/controllers/mytba_controller.py
@@ -1,3 +1,4 @@
+import collections
 import datetime
 import os
 
@@ -6,7 +7,7 @@ from google.appengine.ext.webapp import template
 
 from consts.model_type import ModelType
 from controllers.base_controller import LoggedInHandler
-from database.event_query import EventListQuery
+from database.event_query import EventListQuery, TeamYearEventsQuery
 from database.team_query import EventTeamsQuery
 from helpers.event_helper import EventHelper
 from helpers.event_team_status_helper import EventTeamStatusHelper
@@ -25,14 +26,8 @@ class MyTBALiveController(LoggedInHandler):
         team_favorites_future = Favorite.query(Favorite.model_type == ModelType.TEAM, ancestor=user).fetch_async()
 
         live_events = EventHelper.getWeekEvents()
-        all_events_this_year_future = EventListQuery(now.year).fetch_async()
         favorite_team_keys = map(lambda f: ndb.Key(Team, f.model_key), team_favorites_future.get_result())
         favorite_teams_future = ndb.get_multi_async(favorite_team_keys)
-
-        future_events = filter(lambda e: e.start_date > now, all_events_this_year_future.get_result())
-        future_eventteams_futures = []
-        for event in future_events:
-            future_eventteams_futures.append(EventTeamsQuery(event.key_name).fetch_async())
 
         live_eventteams_futures = []
         for event in live_events:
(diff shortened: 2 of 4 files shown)


phil-lopreiato:
Added button to view the live event webcasts, if they exist

screenshot from 2016-04-21 19-30-24

gregmarra:
Nice! The alignment on "upcoming competitions" vs other headers indicates somethings funky w your div structure here.

For one day events, do we have a way of showing just the single day? See Einstein in your ss.

common/my_filters.py
```diff
@@ -35,3 +35,8 @@ def yt_start(value):
 @register.filter
 def get_item(dictionary, key):
     return dictionary.get(key)
+
+
+@register.filter
+def strftime(d, str):
```
fangeugene: You should be able to use this instead.

https://docs.djangoproject.com/en/1.9/ref/templates/builtins/#date

phil-lopreiato:
Good catch, I'll fix that wonky alignment. There's probably a way to fix the date there, I just copied this code from the team detail page. But this definitely worth a fix in both locations

fangeugene:
Re @gregmarra 's comment about one day events, look at how the event details template does it.

controllers/mytba_controller.py
```diff
@@ -19,21 +21,38 @@ def get(self):
...
 
         live_events = EventHelper.getWeekEvents()
+        all_events_this_year_future = EventListQuery(now.year).fetch_async()
         favorite_team_keys = map(lambda f: ndb.Key(Team, f.model_key), team_favorites_future.get_result())
         favorite_teams_future = ndb.get_multi_async(favorite_team_keys)
 
+        future_events = filter(lambda e: e.start_date > now, all_events_this_year_future.get_result())
+        future_eventteams_futures = []
+        for event in future_events:
+            future_eventteams_futures.append(EventTeamsQuery(event.key_name).fetch_async())
```
fangeugene: Is this fetching eventteams for all future events and then filtering for favorites? That sounds insanely inefficient. Perhaps try querying for eventteams of favorite teams, then querying for events that are in the future?

phil-lopreiato: That approach won't be able to use `DatabaseQuery`s though. I don't have a good grasp of the performance differences - I'll do whichever approach you think is better.

fangeugene: How about making a TeamEvents DatabaseQuery for each team. That's one db read per team, which you can't really expect to get any better. Then filter the events accordingly by time.

phil-lopreiato: There already is a TeamEvents query, but it fetch all events over all years. Is that good enough to use?

fangeugene: There's a TeamYearEvents query

I guess efficiently compared to your original methods still depends on if the average user have more favorites than the average number of future events.

phil-lopreiato: True. Which would you prefer?

fangeugene: I'm going to say do the TeamYearEvents query on all favorites, then filter.

I think the average user has less than 50 favorites, and filtering each team's events for the next one is less work than filtering all events' event teams for all favorite teams.

phil-lopreiato:
Updated screenshot

screenshot from 2016-04-21 20-55-22

nwalters512:
Alignment is still weird...

phil-lopreiato:
Let's try again...
screenshot from 2016-04-22 11-43-37

controllers/mytba_controller.py
```diff
@@ -1,3 +1,4 @@
+import collections
```
fangeugene: Unused import.

controllers/mytba_controller.py
```diff
@@ -6,9 +7,11 @@
 
 from consts.model_type import ModelType
 from controllers.base_controller import LoggedInHandler
+from database.event_query import EventListQuery, TeamYearEventsQuery
```
fangeugene: `EventListQuery` unused

controllers/mytba_controller.py
```diff
@@ -19,6 +21,7 @@ def get(self):
         self._require_registration('/account/register')
 
         user = self.user_bundle.account.key
+        now = datetime.datetime.now()
         team_favorites_future = Favorite.query(Favorite.model_type == ModelType.TEAM, ancestor=user).fetch_async()
 
         live_events = EventHelper.getWeekEvents()
```
fangeugene: Perhaps change this to `getEventsWithinADay()` so a team doesn't show up in this list as well as the upcoming list? You might need to change the logic below to:

`e for e in events if e.start_date > now and not e.within_a_day()`

phil-lopreiato: Good catch, I'll update the logic

PR #1495, opened 2016-04-28
Hot matches

fangeugene:
image

### DIFF
--- a/consts/notification_type.py
+++ b/consts/notification_type.py
@@ -96,3 +96,8 @@ class NotificationType(object):
         ALLIANCE_SELECTION,
         AWARDS
     ]
+
+    enabled_match_notifications = [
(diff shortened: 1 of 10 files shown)


engunneer:
Feature request: Add a match time column to the top matches page. I want to subscribe to top matches today, but not tomorrow, or not get subscribed to matches during my big meeting later, etc.

fangeugene:
@phil-lopreiato I won't be able to do this, but it's pretty simple to do by looking at the upcoming matches table and copying the appropriate header and row `<td>`

phil-lopreiato:
Yeah, I'm going to spend some time this morning tweaking that page

joelg236:
This is so great. At some point can there be a button to subscribe to all top 25 matches?

PR #1512, opened 2016-05-06
iOS Support in myTBA

phil-lopreiato:
Things we need to support iOS and myTBA


phil-lopreiato:
This should be good to go, I believe this is all we need to do server-side

ZachOrr:
Sorry to bump this - is there something else we need before a merge?

phil-lopreiato:
Nope, this is it

ZachOrr:
Thank you! 😊

Issue #1240, opened 2015-09-07
Improve Match View (Score Breakdown + Videos)


phil-lopreiato:
http://www.thebluealliance.com/match/2015hop_qm8

I think videos should go above the breakdown? Or have to click a chevron in order to expand the score breakdown? Or maybe we make it a two-column layout?

There's got to be a better way to organize the content on the match pages.

gregmarra:
I think matches might need to go two columns :)

On mobile, we should definitely make sure videos is above the fold.

PR #1515, opened 2016-05-07
Two-column layout for match view.

tsteward:
This makes a two-column layout for the match view as described in #1240.
screen shot 2016-05-07 at 3 09 14 pm
screen shot 2016-05-07 at 3 09 46 pm
screen shot 2016-05-07 at 3 11 37 pm


fangeugene:
Nice

gregmarra:
What happens on mobile?

fangeugene:
Looks the same as it did before

PR #1517, opened 2016-05-07
[WIP] Run JS tests on Travis

nwalters512:
Runs the Jest JavaScript unit tests on Travis CI.

Also enables dependency caching for pip to speed up the CI process.


nwalters512:
That was surprisingly easy. This should be good to go.

Issue #383, opened 2012-12-28
Insights: Embed video above highest scoring match each year if video available.


gregmarra:
Would be nice to know if there's a video, and buttons are so clickable!

gregmarra:
This would be a great task for a new contributor to pick up.

Look at the template code in `templates/match_details.html` that embeds `video_partials`, and re-use some of it in the `templates/insights_details.html` template.

PR #1518, opened 2016-05-08
Embed video for highest scoring match if available

tsteward:
This embeds the video(s) for the highest scoring match on the insights page for each year if it is available as described in #383. Should I also do this to the insights on the landing page?

### DIFF
--- a/helpers/insights_helper.py
+++ b/helpers/insights_helper.py
@@ -102,7 +102,9 @@ class InsightsHelper(object):
                 'verbose_name': match.verbose_name,
                 'event_name': event.name,
                 'alliances': match.alliances,
-                'winning_alliance': match.winning_alliance
+                'winning_alliance': match.winning_alliance,
+                'tba_video': match.tba_video,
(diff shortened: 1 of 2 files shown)


phil-lopreiato:
How does this look if there are multiple videos for a match? Can you share a screenshot?

tsteward:
screen shot 2016-05-10 at 10 45 33 pm
screen shot 2016-05-10 at 10 45 37 pm
It takes up a lot of space with two videos. Maybe we could limit it to one. I am not sure how we would pick which video to show if we limited it to one.

tsteward:
Maybe we could tile them horizontally if there are two. There is still a problem if there are three or more.
screen shot 2016-05-10 at 10 51 36 pm

gregmarra:
I'd slice off the first one for now – this isn't a problem for 2016, so we
can punt – we'll probably redesign this screen again before it is :)

tsteward:
I made it so that only the first video is shown.

When this is merged I think that the insights page for each year will need to be recalculated.

PR #1520, opened 2016-05-08
Read Backup Teams, Alliance Names from FIRST

phil-lopreiato:
This way, we don't have to explicitly calculate backup teams used. Plus,
we read alliance names, so we can use consistent terms to render them
(also, for Einstein, it shows the Division name)

Also, note that after merging, you should rescrape all 2016 alliances to populate backup team info (we probably don't need to manually clear the API caches, we'll just let them update as they expire)


fangeugene:
API caches don't expire.

PR #1526, opened 2016-05-10
Add warning for people using IE<10

nwalters512:
Jumping through hoops to support ~5% of our users doesn't make sense. This prompts everyone running IE <10 to update to a more modern browser.

Imgur

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -35,6 +35,8 @@ handlers:
   expiration: "1h"
 - url: /javascript
   static_dir: static/compiled/javascript
+- url: /outdatedbrowser
+  static_dir: static/outdatedbrowser
 - url: /jwplayer
   static_dir: static/jwplayer
 - url: /flowplayer
--- a/static/outdatedbrowser/outdatedbrowser.css
+++ b/static/outdatedbrowser/outdatedbrowser.css
@@ -1,0 +1,69 @@
+/* Based on code from outdatedbrowser.com */
+
+#outdated {
+    display: none;
+    position: fixed;
+    top: 0;
+    left: 0;
+    width: 100%;
+    padding-bottom: 30px;
+    text-align: center;
+    z-index: 1500;
+    background-color: #f25648;
+    color: #ffffff;
+  }
+#outdated.visible {
+  display: block;
+}
+* html #outdated {
+    position: absolute;
+}
+#outdated h6 {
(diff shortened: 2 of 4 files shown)


fangeugene:
What is this "screenshot"

nwalters512:
It's the warning shown to people using old versions of IE.

fangeugene:
I mean did you take it with a phone or something?

nwalters512:
Oh, yeah.

fangeugene:
Now might be a good time to update do_compress.py to do a combined style for gameday2, and remove the changes to app.yaml.

nwalters512:
... that was not the commit history I was expecting.

fangeugene:
Lol wut

nwalters512:
A rebase went wrong, I think that fixed it though.

SUPERCILEX:
@fangeugene do we really care about IE?

synth3tk:
If my Google Analytics-fu is correct, then during the 2017 season (Jan 7 - Apr 29), IE 9 and lower accounted for 3.31% of site-wide traffic and roughly less than 2% of GameDay views.

The GameDay stat is based on 7.67% of sessions, due to some thing GA does that I honestly don't fully understand. But even being generous with the numbers, I can't imagine we're even close to 5% anymore.

bdaroz:
At this point if anyone is using IE < 10 would one more warning really matter?

fangeugene:
GA generates results based on a fraction of data if there are too many samples, but I got it to use 100% of the data.
from Jan 1 to April 30: 3.29% of sessions were IE, and <1.5% of that were IE<10.

synth3tk:
IMHO, no. If you're still using IE < 10 at this point, you'd have to know it breaks a lot of sites, and you probably see warnings on a daily basis elsewhere.

fangeugene:
I'm fine with not having this.

synth3tk:
Gotcha. Thanks, Eugene. Is there some magical way to pull that up?

fangeugene:
@synth3tk Any "default" statistic should be using 100% of the data. Custom filters may limit what gets returned (e.g. if you filter by X-TBA-App-ID when looking at API requests)

synth3tk:
That's probably what got me then. I did a custom report to show only /gameday with an IE browser.

PR #1529, opened 2016-05-12
Add Offseason Events Via Suggestion

phil-lopreiato:
screenshot from 2016-05-11 20-33-00
screenshot from 2016-05-11 20-32-45
screenshot from 2016-05-11 20-32-21

### DIFF
--- a/controllers/suggestions/suggest_offseason_event_controller.py
+++ b/controllers/suggestions/suggest_offseason_event_controller.py
@@ -1,0 +1,29 @@
+from controllers.base_controller import LoggedInHandler
+from helpers.suggestions.suggestion_creator import SuggestionCreator
+from template_engine import jinja2_engine
+
+
+class SuggestOffseasonEventController(LoggedInHandler):
+    """
+    Allow users to suggest Offseason Events
+    """
+
+    def get(self):
+        self.template_values.update({
+            "status": self.request.get("status"),
+        })
+        self.response.out.write(
+            jinja2_engine.render('suggest_offseason_event.html', self.template_values))
+
+    def post(self):
+        self._require_login()
+
+        status = SuggestionCreator.createOffseasonEventSuggestion(
+            author_account_key=self.user_bundle.account.key,
+            name=self.request.get("name", None),
+            start_date=self.request.get("start_date", None),
+            end_date=self.request.get("end_date", None),
+            website=self.request.get("website", None),
+            address=self.request.get("venue_address", None),
+        )
+        self.redirect('/suggest/offseason?status=%s' % status)
--- a/controllers/suggestions/suggest_offseason_event_controller.py
+++ b/controllers/suggestions/suggest_offseason_event_controller.py
@@ -9,6 +9,7 @@ class SuggestOffseasonEventController(LoggedInHandler):
     """
 
     def get(self):
+        self._require_login()
         self.template_values.update({
             "status": self.request.get("status"),
         })
--- a/controllers/suggestions/suggest_offseason_event_review_controller.py
+++ b/controllers/suggestions/suggest_offseason_event_review_controller.py
@@ -1,0 +1,102 @@
+from datetime import datetime
+
+from consts.account_permissions import AccountPermissions
+from consts.district_type import DistrictType
+from consts.event_type import EventType
+from controllers.suggestions.suggestions_review_base_controller import \
+    SuggestionsReviewBaseController
+from helpers.event_manipulator import EventManipulator
+from models.event import Event
+from models.suggestion import Suggestion
+from template_engine import jinja2_engine
+
+
+class SuggestOffseasonEventReviewController(SuggestionsReviewBaseController):
+
+    def __init__(self, *args, **kw):
+        self.REQUIRED_PERMISSIONS.append(AccountPermissions.REVIEW_OFFSEASON_EVENTS)
+        super(SuggestOffseasonEventReviewController, self).__init__(*args, **kw)
+
+    def get(self):
+        suggestions = Suggestion.query().filter(
+            Suggestion.review_state == Suggestion.REVIEW_PENDING).filter(
+            Suggestion.target_model == "offseason-event")
+        events_and_ids = [self._create_candidate_event(suggestion) for suggestion in suggestions]
+
+        self.template_values.update({
+            'success': self.request.get("success"),
+            'event_key': self.request.get("event_key"),
+            'events_and_ids': events_and_ids,
+        })
+        self.response.out.write(
+            jinja2_engine.render('suggest_offseason_event_review_list.html', self.template_values))
+
+    def post(self):
+        self.verify_permissions()
+        suggestion = Suggestion.get_by_id(int(self.request.get("suggestion_id")))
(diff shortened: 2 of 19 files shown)


fangeugene:
If the review guidelines say to only approve events with websites, we should just not accept suggestions without websites. In fact, we should make (and specify) that name, date, and website is required, and that venue address is not.

controllers/suggestions/suggest_offseason_event_controller.py
```diff
@@ -0,0 +1,29 @@
...
+from helpers.suggestions.suggestion_creator import SuggestionCreator
+from template_engine import jinja2_engine
+
+
+class SuggestOffseasonEventController(LoggedInHandler):
+    """
+    Allow users to suggest Offseason Events
+    """
+
+    def get(self):
```
fangeugene: This should require login as well, I believe .Check the other suggestion controllers.

controllers/suggestions/suggest_offseason_event_controller.py
```diff
@@ -0,0 +1,29 @@
...
+        self.template_values.update({
+            "status": self.request.get("status"),
+        })
+        self.response.out.write(
+            jinja2_engine.render('suggest_offseason_event.html', self.template_values))
+
+    def post(self):
+        self._require_login()
+
+        status = SuggestionCreator.createOffseasonEventSuggestion(
```
fangeugene: Check that required fields are not None.

controllers/suggestions/suggest_offseason_event_review_controller.py
```diff
@@ -0,0 +1,102 @@
...
+from helpers.event_manipulator import EventManipulator
+from models.event import Event
+from models.suggestion import Suggestion
+from template_engine import jinja2_engine
+
+
+class SuggestOffseasonEventReviewController(SuggestionsReviewBaseController):
+
+    def __init__(self, *args, **kw):
+        self.REQUIRED_PERMISSIONS.append(AccountPermissions.REVIEW_OFFSEASON_EVENTS)
```
fangeugene: SuggestionsReviewBaseController doesn't make sense now, since we changed MUTATE_DATA to REVIEW_MEDIA. perhaps it's best to make `REQUIRED_PERMISSIONS = []` in the base controller, and follow this new pattern for the media review controllers.

controllers/suggestions/suggestions_review_base_controller.py
```diff
@@ -2,10 +2,18 @@
 
 from controllers.base_controller import LoggedInHandler
 
+
 class SuggestionsReviewBaseController(LoggedInHandler):
     """
     Base controller for reviewing suggestions.
     """
+
+    REQUIRED_PERMISSIONS = [AccountPermissions.REVIEW_MEDIA]
```
fangeugene: This doesn't make sense. Why does everything need review_media permissions?

helpers/suggestions/suggestion_creator.py
```diff
@@ -128,3 +129,33 @@ def createMatchVideoYouTubeSuggestion(cls, author_account_key, youtube_id, match
                 return 'video_exists'
         else:
             return 'bad_url'
+
+    @classmethod
+    def createOffseasonEventSuggestion(cls, author_account_key, name, start_date, end_date, website, address):
```
fangeugene: Should we make website required? Also, make it clear in the UI which fields are required.

phil-lopreiato:
> If the review guidelines say to only approve events with websites, we should just not accept suggestions without websites. In fact, we should make (and specify) that name, date, and website is required, and that venue address is not.

I made all 4 fields in the suggestion form required. An event needs a name and dates, obviously. I also think we should require a website (or even a CD thread) as a place to verify the information suggested. And finally, we'll require a place for the event - this can be as simple as a City, State and it'll get formatted into the proper fields at review time. 

I did it that way to try and simplify the suggestion form - the {address, venue, location} fields have a lot of duplicate information, so I figured to just have users suggest the most specific one and let the reviewer split it out as necessary

helpers/suggestions/suggestion_creator.py
phil-lopreiato: It should be required as it. And yes, the UI should make it more clear

controllers/suggestions/suggest_offseason_event_review_controller.py
phil-lopreiato: Yeah, this is the better pattern. Updated.

controllers/suggestions/suggest_offseason_event_controller.py
phil-lopreiato: Good catch

controllers/suggestions/suggestions_review_base_controller.py
```diff
@@ -1,11 +1,17 @@
...
+
+    REQUIRED_PERMISSIONS = []
+
     def __init__(self, *args, **kw):
         super(SuggestionsReviewBaseController, self).__init__(*args, **kw)
-        self._require_permission(AccountPermissions.MUTATE_DATA)
+        self.verify_permissions()
+
+    def verify_permissions(self):
+        for permission in self.REQUIRED_PERMISSIONS:
```
fangeugene: Is `REQUIRED_PERMISSIONS` ever more than just one value?

templates_jinja2/suggest_offseason_event.html
```diff
@@ -0,0 +1,88 @@
...
+                        <ul>
+                            <li>Your suggestion will be reviewed by a moderator</li>
+                            <li>Your account info (like {{ user_bundle.user.email() }}
+                                and {{ user_bundle.user.nickname() }}) will also be submitted. <a
+                                        href="{{ user_bundle.logout_url }}">log in as a different
+                                    user</a></li>
+                            <li>All fields below are required, please do the best you can to fill them accurately</li>
+                            <li>You can only suggest offseason events, please <a href="/contact">contact us</a> you find errors with official regional and district events on the site</li>
+                        </ul>
+                    </div>
```
fangeugene: I think you closed the panel body too early. The "Suggest an Event" header and form should be enclosed.

controllers/suggestions/suggestions_review_base_controller.py
gregmarra: I can imagine cases in the future where it is, so I made it flexible.

fangeugene: But you didn't do that, Phil did.

fangeugene:
The form probably shouldn't clear if not all fields are entered.

http://getbootstrap.com/css/#forms

controllers/suggestions/suggestions_review_base_controller.py
phil-lopreiato: Greg wrote the original class though, I just tweaked it. Anyway, I think it's possible that there will be more than one required permissions for a page, so it's better to keep as a list

fangeugene:
Going to merge this for now. There's some work that can be done cleaning up the submission form so it's more user friendly, but since the suggestion URL isn't exposed I'll merge this now just so I can remove me being the bottleneck for adding offseason events.

PR #1535, opened 2016-05-16
Change Facebook http for https (minor change)

technonerdz:
This is only to change the http for https on the facebook links.


synth3tk:
LGTM

PR #1539, opened 2016-05-20
Ranking predictions

scottrpp:
it better


phil-lopreiato:
Hey, thanks for contributing! Can you please clarify what this diff is supposed to do? It doesn't look like there are any new commits here

PR #1550, opened 2016-06-08
The blue alliance messenger chatbot framework

technonerdz:
This is a framework for an eventual Facebook Messenger Chatbot for TBA as discussed with Greg. *Do not merge this* we need to add things to it before.

### DIFF
--- a/static/php/messengerchatbot.php
+++ b/static/php/messengerchatbot.php
@@ -1,0 +1,27 @@
+<?php
+// parameters
+$hubVerifyToken = 'TOKEN123456abcd';
+$accessToken = "xxx";
+// check token at setup
+if ($_REQUEST['hub_verify_token'] === $hubVerifyToken) {
+  echo $_REQUEST['hub_challenge'];
+  exit;
+}
+// handle bot's anwser
+$input = json_decode(file_get_contents('php://input'), true);
+$senderId = $input['entry'][0]['messaging'][0]['sender']['id'];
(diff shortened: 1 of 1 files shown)


technonerdz:
Do not merge this right now! We need to add things before.

technonerdz:
Just realized that TBA does not have php hosting. Well... We will have to find another solution.

phil-lopreiato:
Can you explain a little bit what the end vision of messenger bots for TBA is (what functionality will it provide)?

Additionally, TBA is all in python, which can also process the messenger webhooks. Look in `*-main.py` files to see url handlers and which controllers they map to.

technonerdz:
A messenger bot to interact with the users on facebook. We could ask questions like: what is the status for a specific team, what is the next match for a team, informations about a team or an event like the website url or the next event date, the rankings for an event and much more.

phil-lopreiato:
Have you taken a look at the [nightbot controller](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/controllers/nightbot_controller.py), which handles those kinds of queries for the gameday twitch chat?

technonerdz:
Yup! That's a great idea to copy this code and modify it for messenger. I will check that later.

technonerdz:
I am looking at the nightbot controller and it only show me how bad I am with Python.

technonerdz:
I got how it works with nightbot! But Messenger is really different than nightbot. We have to send webhooks to the facebook API with an Key and all of those stuff. If someone knows pyhon more than me and want to help, you are welcome!

PR #1564, opened 2016-07-22
Update match validator to support Michigan Eighths Finals

### LINKED ISSUE #1562: API endpoint for single match - does not support eigthfinals in MICMP
Requesting the match data for MICMP eigthfinals does not return any data. For example, the endpoints `/api/v2/match/2016micmp_ef1m1` and `/api/v2/match/2015micmp_ef3m1`.

This appears to affect match lists as well.

brandondean:
Should fix #1562.


phil-lopreiato:
@brandondean can you update https://github.com/the-blue-alliance/the-blue-alliance/blob/master/tests/test_validation_helper.py with some tests to make sure all the expected match ids validate successfully?

fangeugene:
Can you merge with master? That should fix the failing test. Thanks.

Issue #1572, opened 2016-07-25
Improve Rankings Display


phil-lopreiato:
Some good feedback from this thread: https://www.chiefdelphi.com/forums/showthread.php?t=149455

## Ideas
 - Add an "offset" to the "# Matches Played" column that shows how many matches behind the current max a team is.
 - Add RP/match

Both of these items, I feel, should go in parentheses alongside the official number so it's clear they aren't official metrics.

## Implementation
TBA stores rankings in a JSON array of arrays. The first row is an array of column headers and then following are rows in the rankings table. This is because we want the storage format to be generic enough to store multiple years of ranking data (which vary pretty widely). We need to be careful about backwards compatibility - existing versions of the android expect the format to be a certain way, and if we change it, we'll cause crashes.

 We can calculate these fields in [the parser](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/datafeeds/parsers/fms_api/fms_api_event_rankings_parser.py).

fangeugene:
For simplicity, we should just render this on the fly for now. Add a property `rankings_enhanced` or something right below https://github.com/the-blue-alliance/the-blue-alliance/blob/master/models/event.py#L205 and make the appropriate changes in the templates.

phil-lopreiato:
For the played offsets (only for live events, I think), the table should look something like this:

| Team | RP | Played |
| ----- | --- | ------- |
| 254   | 10  | 3          |
| 123   | 8    | 2  *(-1)*  |
| 1114 | 8     | 3  |

PR #1573, opened 2016-07-26
Add Rank Points Per Match to Rankings Table

maherb:
added rankings_enhanced that uses model {"team_number": {"ranking_score_per_match": number}}

towards #1572

### DIFF
--- a/controllers/event_controller.py
+++ b/controllers/event_controller.py
@@ -173,13 +173,16 @@ class EventDetail(CacheableHandler):
         # rankings processing for ranking score per match
         full_rankings = event.rankings
         rankings_enhanced = event.rankings_enhanced
-        if rankings_enhanced:
-            rp_index = full_rankings[0].index("Ranking Score")
-            for row in full_rankings[1:]:
-                team = row[1]
-                if team in rankings_enhanced:
-                    rp_per_match = rankings_enhanced[team]['ranking_score_per_match']
-                    row[rp_index] = "{} ({} per match)".format(row[rp_index], rp_per_match)
+        if rankings_enhanced is not None:
+            rp_index = 2
+            ranking_criterion_name = full_rankings[0][rp_index]
+            full_rankings[0].append(ranking_criterion_name + "/Match*")    
+            if rankings_enhanced:
+                for row in full_rankings[1:]:
+                    team = row[1]
+                    if team in rankings_enhanced:
+                        rp_per_match = rankings_enhanced[team]['ranking_score_per_match']
+                        row.append(rp_per_match)
 
 
         self.template_values.update({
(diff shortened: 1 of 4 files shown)


models/event.py
```diff
@@ -214,6 +215,23 @@ def rankings(self):
                 self._rankings = None
         return self._rankings
 
+    @property 
+    def rankings_enhanced(self):
+        if self._rankings is not None and self.year == 2016:
+            self._rankings_enhanced = {}
+            team_number_index = self._rankings[0].index("Team")
```
phil-lopreiato: use `self.rankings` for these. If `self.rankings` hasn't been called yet, then `self._rankings` will be `None` since it hasn't been lazy loaded yet

maherb: Functionality implemented, screenshot attached. I see two problems with the current implementation:
1) Doesn't look great
2) Not possible to sort by average RS, which seems to be a big part of why users want this

I missed Phil's comment there. I will include that correction in the next commit.

image

phil-lopreiato:
I'm a fan!

> 2) Not possible to sort by average RS, which seems to be a big part of why users want this

How about we make average RS its own column? These tables have a cool trick where you can click on the header to sort the values. You can [insert](http://www.tutorialspoint.com/python/list_insert.htm) values into the `full_rankings` rows. Be sure to insert a header in the first row and a blank value for rows where no average is available, to keep the columns lined up.

templates_jinja2/event_details.html
```diff
@@ -211,6 +211,7 @@ <h3 class="panel-title">Event in progress</h3>
               {% endfor %}
             </tbody>
           </table>
+          {% if event.year in [2012, 2013, 2014, 2016] %}<p>*This column is calculated for your convenience by The Blue Alliance using data from FIRST and is not official.</p>{% endif %}
```
fangeugene: italicize FIRST

maherb:
Added consts.ranking_indexes to all manipulating rankings from all years, and added the <ranking criterion>/match column to all events 2007 onward except 2015, which already ranks by average
image

fangeugene:
Did you commit ranking_indexes?

controllers/event_controller.py
```diff
@@ -170,6 +171,21 @@ def _render(self, event_key):
         if event_insights:
             event_insights_template = 'event_partials/event_insights_{}.html'.format(event.year)
 
+        # rankings processing for ranking score per match
+        full_rankings = event.rankings
+        rankings_enhanced = event.rankings_enhanced
+        if rankings_enhanced is not None:
```
phil-lopreiato: add `and event.year in RankingIndexes.CUMULATIVE_RANKING_SCORE` so we don't get crashes when 2017 events are added before the constants are updated

phil-lopreiato:
LGTM! 👍

Issue #1549, opened 2016-06-07
Differentiate Overall Record From Official Events


phil-lopreiato:
As suggested: http://www.chiefdelphi.com/forums/showthread.php?p=1591740#post1591740

## Affected files:
 - [Records are calculated here](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/renderers/team_renderer.py#L55). We will want to only run the overall record calculation for official events

PR #1576, opened 2016-07-28
Official record

maherb:
In response to Issue #1549. Offseason events are no longer included in teams' season WLT record.

### DIFF
--- a/renderers/team_renderer.py
+++ b/renderers/team_renderer.py
@@ -61,7 +61,8 @@ class TeamRenderer(object):
                 qual_avg = None
                 elim_avg = None
                 wlt = EventHelper.calculateTeamWLTFromMatches(team.key_name, event_matches)
-                year_wlt_list.append(wlt)
+                if event.official:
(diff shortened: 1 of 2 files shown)


fangeugene:
Can you rebase with master?

renderers/team_renderer.py
```diff
@@ -61,7 +61,8 @@ def render_team_details(cls, handler, team, year, is_canonical):
                 qual_avg = None
                 elim_avg = None
                 wlt = EventHelper.calculateTeamWLTFromMatches(team.key_name, event_matches)
-                year_wlt_list.append(wlt)
+                if event.official:
```
fangeugene: This doesn't work. Certain offseason events that use the FRC API count as "official" too. Kind of awkward but unfortunately that's how it is. You should create a new set in https://github.com/the-blue-alliance/the-blue-alliance/blob/master/consts/event_type.py called something like `OFFICIAL_SEASON_EVENT_TYPES` and check that the event type is in that set.

PR #1579, opened 2016-07-29
Implemented match offset for the rankings with (-1) next to matches played

### LINKED ISSUE #1572: Improve Rankings Display

maherb:
Closes #1572, formatted per Phil's request.

image


controllers/event_controller.py
```diff
@@ -176,14 +176,20 @@ def _render(self, event_key):
...
+
+            for row in full_rankings[1:]:
+                team = row[1]
+                if rankings_enhanced["ranking_score_per_match"] is not None:
+                    rp_per_match = rankings_enhanced['ranking_score_per_match'][team]
+                    row.append(rp_per_match)
+                if rankings_enhanced["match_offset"] is not None:
+                    match_offset = rankings_enhanced["match_offset"][team]
+                    if match_offset == -1:
+                        row[matches_index] = "%s (-1)" % row[matches_index]
```
phil-lopreiato: Two things. I wouldn't assume the only possible offset is `-1`, since surrogate matches don't count as official "played" ones.

Also, use `row[matches_index] = "{} ({})".format(row[matches_index], offset)` instead

phil-lopreiato:
Thanks for doing this! 👍

controllers/event_controller.py
maherb: Done!

phil-lopreiato:
LGTM

PR #1587, opened 2016-08-03
[GD2] Show "No webcasts" message when appropriate

nwalters512:
This displays a message if people try to visit GD2 when there aren't any active webcasts. It also includes a link to the TBA homepage. It's barebones, but it gets the job done.

There's also some behind-the-scenes refactoring that renders the main page contents (video grid, layout selector, message from this PR) as the child of another component that handles positioning when the sidebars are visible. This reduces dependencies between components and makes it easier to add different "main" views.

Imgur


nwalters512:
3ad746d doesn't *totally* belong here, but I added a new selector so I think adding tests and fixing the old one is fine.

PR #1590, opened 2016-08-05
[GD2][WIP] Wrote some component tests

nwalters512:
Adds dependencies, setup, and some initial code for testing React components.

[Airbnb's enzyme](http://airbnb.io/enzyme/) is used to assert, manipulate, and traverse over the output of components. Jest's recently-announced [snapshot testing](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html) is used to prevent unwanted changes or regressions in the output of components.

At the moment, it looks like it's not possible to use both enzyme and snapshot tests in the same test file. I really want to be able to use both so that I can both test the behavior of components by simulating events with enzyme and make sure that rendering doesn't change unexpectedly with snapshots. This may mean having to write two sets of tests for each component: one to check behavior, and one to check rendering. I'm still deciding what the best way to do this is.


nwalters512:
I split the tests into two files for each component; one does snapshot (rendering) testing/verification, the other does behavior testing. This is probably going to have to stay this way because we can't inject two React rendering environments in the same scope; see https://github.com/facebook/react/issues/7386

PR #1610, opened 2016-09-09
Fixed ESLint prefer-const errors

nwalters512:
Not sure why the linter started erroring only recently; I couldn't track down any relevant changes in ESLint or the Airbnb style guide we use as our base. Regardless, with this PR, our CI builds shouldn't error any more.


nwalters512:
The latest build error is caused by something else that's Python-related; I'm probably not the best one to debug that. @fangeugene @gregmarra?

PR #1613, opened 2016-09-13
Updated blue banner styling

ErikBoesen:
Looks like this:

Screenshot


phil-lopreiato:
Can you share a before/after screenshot?

ErikBoesen:
Before:
screen shot 2016-09-13 at 9 51 32 am
After:
screen shot 2016-09-13 at 9 47 29 am

ErikBoesen:
Not completely sure about the italics. Thoughts?

nwalters512:
Any particular reason you're favoring the white outline over the colored FIRST logo? IIRC the real blue banners use the colored one.

ErikBoesen:
No real reason other than aesthetics.

fangeugene:
I like the font changes -- makes the award name pop more. I'm not wild about how the event name starts at different heights, since it looks messy. I'm still not sure myself whether to go colored or white FIRST logo. I'm leaning toward colored because 1) when the logo is colored, your eye can snap to the award name whereas having everything white makes the text less readable and 2) the pages need more color anyways

Also, I see you made the banners wider. Can you make sure they still show up as 2 per row on the team history page?

ErikBoesen:
I'll take a look at the changes you've proposed.

What do you mean by 2 per row?

fangeugene:
image

fangeugene:
Did you push?

ErikBoesen:
Oh, that. I hadn't looked at the history page. Currently it's a single column, so I'll look into fixing that.

Also, on the misalignment of the event names, I'm not sure what to do. While I'd like it to match, it doesn't look very good when there aren't any chairmans' to compare to. One possibility would be to remove "AWARD" from the title. But that doesn't seem very professional.

ErikBoesen:
Okay, pushed some changes. They now look like this:

screen shot 2016-09-13 at 10 42 24 am

Overall they're now fairly similar to before, but there's still some notable font differences.

fangeugene:
Cool I like it. Can you give the award name just a tiny tiny bit more space on top and bottom? like 1-2 px. I know it's a balance between making "WINNER" not look too spaced out and "CHAIRMAN's AWARD" not be too squished, but I think right now it's a bit on the squished side.

Also, I don't think you committed reverting back to the colored logo.

ErikBoesen:
Okay, I've done that.

fangeugene:
Sweet

ErikBoesen:
Thanks for the merge!

PR #1620, opened 2016-09-20
Add admin bar to make tasks easier

fangeugene:
Adds a bar to the bottom of all pages if the user is logged in as an admin. This bar currently allows for easier cache clearing of the page, but can be extended to do more.

If the user is an admin, none of generated HTML is cached in either memcache or edge cache. Adding the admin bar makes used of existing memcache by find/replacing an HTML comment.

I'm confident that the current setup doesn't improperly leak the admin bar to non admins (in the worst case, the button they click will still prompt for login). However, the bar may not show up for admins if the page already exists in edge cache (I'm not sure), but through experience and [this documentation](https://cloud.google.com/appengine/docs/python/how-requests-are-handled#cache-control_expires_and_vary), it seems like GAE never sets cache headers if the user is an admin so it probably won't be a problem.

image

### DIFF
--- a/controllers/base_controller.py
+++ b/controllers/base_controller.py
@@ -13,6 +13,7 @@ from google.appengine.api import memcache
 import tba_config
 
 from helpers.user_bundle import UserBundle
+from template_engine import jinja2_engine
 
 
 class CacheableHandler(webapp2.RequestHandler):
@@ -27,6 +28,8 @@ class CacheableHandler(webapp2.RequestHandler):
         super(CacheableHandler, self).__init__(*args, **kw)
         self._cache_expiration = 0
         self._last_modified = None  # A datetime object
+        self._user_bundle = UserBundle()
+        self._is_admin = self._user_bundle.is_current_user_admin
         if not hasattr(self, '_partial_cache_key'):
             self._partial_cache_key = self.CACHE_KEY_FORMAT
         self.template_values = {}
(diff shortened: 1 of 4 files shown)


gregmarra:
"Admin" here will mean GAE admin, not "permissions" based admins.

controllers/base_controller.py
```diff
@@ -48,16 +51,24 @@ def _render_cache_key(cls, cache_key):
             cls.CACHE_VERSION,
             tba_config.CONFIG["static_resource_version"])
 
+    def _add_admin_bar(self, html):
+        if self._is_admin:
+            self.template_values["cache_key"] = self.cache_key
+            self.template_values["user_bundle"] = self._user_bundle
+            admin_bar = jinja2_engine.render('admin_bar.html', self.template_values)
+            return html.replace('<!-- Admin Bar -->', admin_bar)
```
phil-lopreiato: :smile:

fangeugene: Do you have a better solution?

phil-lopreiato: Could you modify the base template to include something like

```
{% if should_show_admin_bar %}
{% include "admin_bar.html" %}
{% endif %}
```

And put the value for `should_show_admin_bar` in `self.template_values`?

fangeugene: I had that initially, but then it wouldnt be using memcache.

phil-lopreiato: Ah, true. No other ideas come to mind...

PR #1624, opened 2016-09-21
Add way to override cache headers via sitevar

phil-lopreiato:
The `turbo_mode`sitevar looks like:
```
{
"regex": "main_.*",
"valid_until": 1474488207,
"cache_length": 128
}
```

And takes a regex to match cache keys on, a new expiration time, and a UNIX timestamp of how long the override is valid. The new value will be set for memcache and response header cache expiration times.


fangeugene:
I implemented it in a cleaner way here https://github.com/the-blue-alliance/the-blue-alliance/commit/4a3b74834c2f50c12db04fbabd9da4cd7932ea4f

PR #1663, opened 2016-10-23
Suggestion Flow for Trusted API Keys

### LINKED ISSUE #1659: Build tool for users to request API write keys
We can tie keys to an account, users can view keys when logged in, approval can just be a single button click instead of copy/pasting keys and sending them.

We should probably require full name and a description of their role (offseason event FTC, event organizer, etc.) when requesting keys. We already have emails associated with the account if we need to contact them.

phil-lopreiato:
Does a few things:
 - Adds expiration to Trusted API Keys
 - Add permission to review trusted keys
 - Request & review trusted keys via suggestion flow
 - Can display a user's keys on the account dashboard

Last question: how should we notify a user when their keys are approved? Should we just email them?

screenshot from 2016-10-23 11-37-50
screenshot from 2016-10-23 11-37-27
screenshot from 2016-10-23 11-34-46

Closes #1659


bdaroz:
In the interest of convenience, with a dash of security, emailing the requestor that their key was approved or declined (stating which), but directing them back to the site to get the key would be best. One could argue including the keys in email is easier for the end-user (one stop resource), and would allow them to be forwarded to others that may need it (which may / may not be desired, from the TBA side of things), but best practice security-wise is to not send multi-use security credentials via email.

controllers/suggestions/suggest_apiwrite_controller.py
```diff
@@ -0,0 +1,37 @@
...
+            jinja2_engine.render('suggest_apiwrite.html', self.template_values))
+
+    def post(self):
+        self._require_login()
+
+        status = SuggestionCreator.createApiWriteSuggestion(
+            author_account_key=self.user_bundle.account.key,
+            event_key=self.request.get("event_key", None),
+            affiliation=self.request.get("role", None),
+            auth_types=self.request.get_all("auth_types", [])
```
fangeugene: I think this should sanitize `auth_types` to make sure they're in `AuthType.type_names.keys()` Or else I can request trusted API keys for an event in the future, spoof auth_types to include numbers we haven't yet defined, and if we do define them in the future then suddenly I have access.

controllers/suggestions/suggest_apiwrite_review_controller.py
```diff
@@ -0,0 +1,86 @@
...
+        })
+        self.response.out.write(
+            jinja2_engine.render('suggest_apiwrite_review_list.html', self.template_values))
+
+    def post(self):
+        self.verify_permissions()
+        suggestion = Suggestion.get_by_id(int(self.request.get("suggestion_id")))
+        verdict = self.request.get("verdict")
+        if verdict == "accept":
+            user = suggestion.author.get()
```
fangeugene: All this should be in a transaction to ensure it's treated as an atomic operation. See https://github.com/the-blue-alliance/the-blue-alliance/blob/master/controllers/suggestions/suggest_team_media_review_controller.py#L147

helpers/suggestions/suggestion_creator.py
```diff
@@ -183,3 +184,31 @@ def createOffseasonEventSuggestion(cls, author_account_key, name, start_date, en
...
+            event = Event.get_by_id(event_key)
+            if event and event.event_type_enum not in EventType.SEASON_EVENT_TYPES:
+                suggestion = Suggestion(
+                    author=author_account_key,
+                    target_model="api_auth_access"
+                )
+                suggestion.contents = {
+                    'event_key': event_key,
+                    'affiliation': affiliation,
+                    'auth_types': [int(type) for type in auth_types],
```
fangeugene: You can also sanitize the auth_types (as mentioned above) here.

templates_jinja2/suggest_apiwrite.html
```diff
@@ -0,0 +1,101 @@
+{% extends "base.html" %}
+
+{% block title %}The Blue Alliance - Request Write Keys{% endblock %}
+
+{% block meta_description %}Request auth tokens to import your offseason event to The Blue Alliance{% endblock %}
```
fangeugene: add "using tools like TBA EventWizard (link) or Cheesy Arena (link)"

templates_jinja2/suggest_apiwrite.html
```diff
@@ -0,0 +1,101 @@
...
+                            <li>Your suggestion will be reviewed by a moderator</li>
+                            <li>After review, we will send you an email at the address associated
+                                with your TBA account, <strong>{{ user_bundle.user.email() }}</strong>.
+                                You can also find auth tokens associated with your account on your
+                                <a href="/account">account overview</a>. You can also
+                                <a href="{{ user_bundle.logout_url }}">log in as a different
+                                    user</a> if this is not correct.</li>
+                            <li>You can only request tokens for offseason events, please
+                                <a href="/contact">contact us</a> you find errors with official
+                                regional and district events on the site
```
fangeugene: Missing period.

controllers/suggestions/suggest_apiwrite_controller.py
phil-lopreiato: Nice catch

templates_jinja2/suggest_apiwrite.html
phil-lopreiato: This is in the `<meta>` tag for the page, you sure about that? I can add some links to the overview paragraph, however

fangeugene: Oh I didn't realize that was in the meta tag. Yes, put it in the overview paragraph.

phil-lopreiato: Done in 0c72b43

controllers/suggestions/suggest_apiwrite_review_controller.py
```diff
@@ -0,0 +1,145 @@
...
+        email_body = None
+        user = None
+        event_key = None
+        status = ''
+        if verdict == "accept":
+            status = 'accept'
+            user, event_key, email_body = self._process_accepted(suggestion_id, message)
+
+        elif verdict == "reject":
+            suggestion = Suggestion.get_by_id(suggestion_id)
```
fangeugene: Nevermind, there's only one DB write for reject, and if the transaction fails it will throw an error so no email will go out.

templates_jinja2/suggest_apiwrite.html
```diff
@@ -0,0 +1,104 @@
...
+                        <h1 class="panel-title">Add data for your offseason event!</h1>
+                    </div>
+                    <div class="panel-body">
+                        <p>Thanks for helping make The Blue Alliance better! Let us know if you're
+                            willing to import data for an offseason event you run! These auth keys
+                            allow you to import data via the <a href="/eventwizard">TBA Event
+                            Wizard</a> as well as the
+                            <a href="https://github.com/Team254/cheesy-arena">Cheesy Arena FMS</a></p>
+                        <ul>
+                            <li>Your suggestion will be reviewed by a moderator</li>
```
fangeugene: suggestion -> request

fangeugene: suggestion -> request

main.py
```diff
@@ -101,6 +104,8 @@ def __call__(self, request, response, exception):
       RedirectRoute(r'/predictions', PredictionsHandler, 'predictions', strict_slash=True),
       RedirectRoute(r'/record', RecordHandler, 'record', strict_slash=True),
       RedirectRoute(r'/search', SearchHandler, 'search', strict_slash=True),
+      RedirectRoute(r'/suggest/apiwrite/', SuggestApiWriteController, 'suggets-apiwrite', strict_slash=True),
```
fangeugene: suggest -> request

fangeugene: Also, `suggets-apiwrite` is spelled wrong here and below, which is missing `-review`

main.py
```diff
@@ -100,6 +103,8 @@ def __call__(self, request, response, exception):
       RedirectRoute(r'/opr', OprHandler, 'opr', strict_slash=True),
       RedirectRoute(r'/predictions', PredictionsHandler, 'predictions', strict_slash=True),
       RedirectRoute(r'/record', RecordHandler, 'record', strict_slash=True),
+      RedirectRoute(r'/request/apiwrite/', SuggestApiWriteController, 'request-apiwrite', strict_slash=True),
+      RedirectRoute(r'/request/apiwrite/review', SuggestApiWriteReviewController, 'request-apiwrite-review', strict_slash=True),
```
fangeugene: this should still be suggest

PR #1667, opened 2016-10-24
Web notifications, Fixes #1182

### LINKED ISSUE #1182: Look Into Push Notifications for Browsers
http://updates.html5rocks.com/2015/03/push-notificatons-on-the-open-web

Looks like we'll just have to add a WEB ClientType, which will act just like android (it seems these APIs go though GCM, as well).

We can make GameDay prompt the user to display the same notifications that would be sent to mobile devices. This might also be a good opportunity to implement cross-device dismisses, as well.

fangeugene:
This PR enables push notifications on the web via Firebase Cloud Messaging. Currently only displays Ping and a bare-bones Match Score notifications when the website is closed or is NOT the tab in focus in the browser. It requires the user to enable notifications on the Account page.

Issues to be addressed in a later PR:
- Requires the user to be logged in for background notifications, but GAE login sessions are set to 2 weeks max.
- Support all notification types.
- Handle case where the website is the tab in focus.

Some images:
image

image

image

image


controllers/ajax_controller.py
```diff
@@ -14,10 +15,61 @@
...
+            MobileClient(
+                parent=ndb.Key(Account, user_id),
+                user_id=user_id,
+                messaging_id=fcm_token,
+                client_type=client_type,
+                device_uuid=uuid,
+                display_name=display_name).put()
+        else:
+            # Record already exists, update it
+            client = query.fetch(1)[0]
```
phil-lopreiato: Should we also update the user id here? This lets us account for the different users on one device, since tokens are tied to the browser and not the account? This way, we just keep the browser tied to the last user who logged in on it.

static/firebase-messaging-sw.js
```diff
@@ -0,0 +1,78 @@
...
+  // Check for login
+  fetch('/_/account/info', {credentials: 'include'})
+  .then(function (response) {
+    response.json().then(function (accountInfo) {
+      if (accountInfo.logged_in) {
+        console.log('[TBA FCM SW] Message:', payload);
+
+        buildAndShowNotification(self.registration, payload);
+      } else {
+        console.log("[TBA FCM SW] Not logged in! Deleting token...");
```
phil-lopreiato: I think we need to send this event to the TBA backend as well

static/javascript/tba_js/tba.js
```diff
@@ -153,3 +153,13 @@ $(document).ready(function(){
     $("#robot-image").html('<div class="thumbnail carousel team-media-carousel"><a href="' + url + '" target="_blank"><img src="' + url + '" alt="April Fools!"></a></div>');
   }
 });
+
+// Init Firebase
+var config = {
```
phil-lopreiato: We shouldn't check in the api keys, firebase domain, and storage bucket. Can we have some placeholders here, and then do some regex in a paver task that replaces them with the real values at deploy time?

static/javascript/tba_js/tba_fcm.js
```diff
@@ -0,0 +1,139 @@
...
+        if (accountInfo.user_id != window.localStorage.getItem('TBA_lastFCMTokenUserId')) {
+          console.log('[TBA FCM] Different user logged in. Can\'t use same token.');
+
+          messaging.requestPermission()
+          .then(function() {
+            return messaging.getToken();
+          })
+          .then(function(token) {
+            console.log('[TBA FCM] Deleting token:', token);
+            messaging.deleteToken(token)
```
phil-lopreiato: We probably should notify the backend here, as well

static/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
+  "short_name": "The Blue Alliance",
+  "name": "The Blue Alliance",
+  "start_url": "/",
+
+  "gcm_sender_id": "103953800507"
```
phil-lopreiato: I don't think this should be checked in either

nwalters512: `gcm_sender_id` is the same for all Firebase web clients, it shouldn't matter if this is checked in.

phil-lopreiato: Ah, you're right. This is the common one. The one above is the one we shouldn't check in

controllers/ajax_controller.py
fangeugene: Do you mean `user_id` or `uuid`?

`user_id` would never change for a given user, and all keys and queries are based off of it.
We shouldn't change `uuid` because it's unique per browser (unless someone clears the browser data). If it weren't  unique per browser, every time a user logs in and out, we need to generate a new UUID and then they will end up having 100s of MobileClients associated.

static/javascript/tba_js/tba.js
fangeugene: These aren't private keys (so we're not screwed that I've already committed and pushed this), but I guess it would be nice to not hardcode these into the repo if people want to fork it. But then by that argument any reference to "tbatv-prod-hrd" and "thebluealliance.com" should be replaced during deployment 😛.

I'll see what I can do.

static/manifest.json
fangeugene: :+1:

controllers/ajax_controller.py
phil-lopreiato: `user_id`. You were saying that the FCM token doesn't change if you log out and log back in with a different account. Therefore, I think we should always keep the `uuid` + `user_id` pair in sync with the most recent user to sign in on that browser.

static/firebase-messaging-sw.js
phil-lopreiato: You're right, the backend should be able to clean that up when the send fails. So probably not a huge deal, although we may need to revisit that in the future depending on how things shake out

controllers/ajax_controller.py
fangeugene: But we can't update `user_id` since that's part of the key.
Are you suggesting deleting the MobileClient and creating a new one with the updated user_id? That can be dangerous because then the only thing we're using to check validity is UUID.

static/firebase-messaging-sw.js
fangeugene: I think we definitely need to handle cleanup of webhooks, tokens, etc. in the backend. And this can be handled then as well.

static/javascript/tba_js/tba.js
fangeugene: I mean none of this is private because it's all client side javascript code. The validation comes from the pair of 1) GCM key on the server and 2) FCM token.

PR #1678, opened 2016-10-27
Send Emails to contact@ on Important Suggestions

phil-lopreiato:
For now, just offseason events and api key requests


controllers/suggestions/suggest_apiwrite_controller.py
```diff
@@ -1,10 +1,11 @@
 from consts.auth_type import AuthType
 from controllers.base_controller import LoggedInHandler
+from controllers.suggestions.suggestions_base_controller import SuggestionsBaseController
 from helpers.suggestions.suggestion_creator import SuggestionCreator
 from template_engine import jinja2_engine
 
 
-class SuggestApiWriteController(LoggedInHandler):
+class SuggestApiWriteController(SuggestionsBaseController):
```
fangeugene: If I understand correctly, SuggestionsBaseController is only useful to wrap the email sender. Would it make more sense to keep this LoggedInHandler and wrap the email sending in a helper?

controllers/suggestions/suggestions_base_controller.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, *args, **kw):
+        super(SuggestionsReviewBaseController, self).__init__(*args, **kw)
+        self.verify_permissions()
+
+    def verify_permissions(self):
+        for permission in self.REQUIRED_PERMISSIONS:
+            self._require_permission(permission)
+
+
+class SuggestionsBaseController(LoggedInHandler):
```
fangeugene: I think this is kind of pointless. A helper makes more sense.

phil-lopreiato: :+1:

phil-lopreiato:
Moved the email code to a helper

gregmarra:
This is awesome.

gregmarra:
Should we email when they are acted on too? That way several people won't
go to look.

phil-lopreiato:
Yeah, I think that's a good idea. We'll just have to make sure they stay in the same thread

PR #1689, opened 2016-10-30
Event filter, Search for nearby teams and events

fangeugene:
# Filter events by state/province
image

# Search nearby
image

image

image

image


bdaroz:
I should probably go into hiding for this but... from a UX standpoint.

`km` vs `mi`?

PR #1693, opened 2016-10-31
Keyless Eventwizard

phil-lopreiato:
Yay!

When a user is logged in:
 - The "Select Event" dropdown in the eventwizard will now be populated with events the user has trusted API tokens for
 - The id/secret boxes will disappear (but they'll still be there for when you manually input an event key)
 - The trusted API now validates that the requested event + current user have auth tokens and will count that as valid (with updated tests)

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -9,6 +9,17 @@ indexes:
 # automatically uploaded to the admin console when you next deploy
 # your application using appcfg.py.
 
+- kind: ApiAuthAccess
+  properties:
+  - name: event_list
+  - name: owner
+  - name: expiration
+
+- kind: ApiAuthAccess
+  properties:
+  - name: owner
+  - name: expiration
+
 - kind: Event
   properties:
   - name: __key__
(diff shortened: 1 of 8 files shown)


index.yaml
```diff
@@ -9,6 +9,17 @@ indexes:
 # automatically uploaded to the admin console when you next deploy
 # your application using appcfg.py.
 
+- kind: ApiAuthAccess
+  properties:
+  - name: event_list
+  - name: owner
+  - name: expiration
+
+- kind: ApiAuthAccess
```
fangeugene: Are both of these necessary? I don't understand indexes fully, but I would assume the above includes this one. To test, you can delete both indexes, restart the dev server, and then run through both ApiAuthAcces.query() calls.

phil-lopreiato: Maybe? This file was auto-generated, so I'll start from scratch and see what happens

phil-lopreiato: I ran it starting from no changes and from only the first index. After both times, the file ended up like what was committed. So I presume that's what we need, since I don't really understand indexes fully either

fangeugene: Ok then.

PR #1700, opened 2016-11-04
Cleaner Offseason Review Page

phil-lopreiato:
@gregmarra @fangeugene 
Does this seem clearer about where to put the key? I also added some validation :)
screenshot from 2016-11-04 13-05-57


fangeugene:
LGTM. @gregmarra does this confused you less?

gregmarra:
This is much better. The example is awesome.

Is there a browser native way to add hint text to each of these boxes? I
think we have that in the creation flows in the current admin panel.

phil-lopreiato:
I don't think the placeholder text will work here because it looks very similar to filled in text at a quick glance. I wanted it to be obvious which fields were blank and to be able to see the examples side by side so that the reviewer can compare and make sure the format is correct. 

I could mock it up though if you'd like to see it

fangeugene:
Merging this for now.

PR #1704, opened 2016-11-24
Moar Social Media. Snapchat, Twitch and Pinterest

technonerdz:
I saw that a lot of teams have snapchats and Twitch accounts. I also added pinterest while I was at it.

Screenshots:
suggest:
suggest

review:
reviewreal

on team page:
teampage


fangeugene:
So before this gets merged, we should have a discussion about where we draw the line for social media accounts. It seems ridiculous to support every type of social media account, so how do we decide what to implement and what not to?

bdaroz:
Referencing the various social media accounts of the teams does help "Make it LOUD", and I don't have any problem with Snapchat or Pinterest. Both are mainstream and serve a somewhat unique niche and target audience. As far as Twitch goes -- we have YouTube, and we all know the fun of posting match video there if there's even any hint of background music....

I think we may be able to draw two, somewhat blurry, lines in the sand. First, non-mainstream niche social media sites can probably be excluded, and second, If that social media avenue is simply parroting another social media account it gets redundant fast.

(Eg, Vine probably would have been ok when it got popular as it was fairly unique and mainstream, but after the Twitter acquisition and basically every Vine becoming a Tweet, I would have advocated for it's removal.)

technonerdz:
I agree with @bdaroz .

Maybe Twitch shouldn't be on team pages. If a team is hosting its own FRC event, it would be featured on GameDay anyway. If a team wants to do other random livestreams, Facebook Live and Youtube would be the best option because Twitch is not meant for that.

On the other end, I think Snapchat and even Pinterest should be there. Snapchat brings an exclusive look at what a team is doing.

What do you guys think?

jaredhasenklein:
I was just thinking about Snapchat today...we recently started using Snapchat and I agree that I think it helps spread the message. I don't see how Pinterest fits in but maybe that's just me...

Side note: I think it would be cool to have a team Facebook/Twitter/Instagram feed on a team's page.

themacman:
I'm in general agreence with the last 3 commenters. 

As far as Twitch goes, I'm not quite sure. I could eaisly see teams streaming build nights and such on Twitch Creative, but everyone else has some good points on why it isn't really a "social media" that teams use.

synth3tk:
I agree with:
- Not really feeling Pinterest. I'm sure there are some teams that use it, but are there enough where its inclusion is necessary?
- I can definitely see why Twitch should be included (@themacman brought up a great point about build nights), but I wouldn't be opposed to leaving it out.

As far as the "line in the sand", I think if it's popular, not redundant, and non-niche (market-wise), we should at least strongly consider it.

phil-lopreiato:
I think the line in the sand should be at something with a widget we can embed on the team's page. A long list of social media accounts starts to not look good with more than 4 or 5 entries and it gets hard to build a page to display them nicely. 

The intent behind adding social media support was to give teams a means to customize their TBA pages via their existing social media profiles, and that's not something that can happen unless we can embed some sort of content into the TBA page. The underlying assumption here (and my personal opinion) is that TBA shouldn't be simply a social media account aggregator, but to intelligently use the links we have to build better content[1]. If we can't use the information to display richer information[2], then I don't think the link is worthwhile in the clutter it adds.

Applying that litmus test to the proposed account types here, I'd say we don't need periscope (this can be removed and merged into twitter), pintrest, or snapchat.

[1] If we want TBA to be an aggregator, we'll have to come up with a more scalable way to present a large amount of social account connections, especially on mobile. 
[2] I'm still not quite sure how we want to actually go about doing this for the social sites and make it not look awful, but it's still the end goal to integrate

fangeugene:
Perhaps we should put out a poll for what social media teams use, and only support those that have > X% just to prune the obscure ones. I don't mind TBA being an aggregator, since the whole point is to provide more data and more ways for teams to #makeitloud. [This](https://cloud.githubusercontent.com/assets/1951283/20582525/e62e868e-b1af-11e6-8f87-174eff08bca7.png), however, is already bordering on clutter. Perhaps we should only show the big few, and then have it expand to show the others.

We also shouldn't be embedding 1st party widgets for every single type of social media, if any. David's right that it would look really bad.

bdaroz:
One thing to consider, before going down the road of adding all the social media posts to the team TBA page.... We could easily "make it redundant" instead of making it loud.... How many teams put up a FB event then link to it from Twitter, post a pic of the flyer linked to FB on instagram, etc, or post a video on Youtube, link to it from FB, twitter, instagram, etc?

Teams want to get their message out broadly - and (I believe) often that's done by not focusing on a single social media app/audience.

If we start embedding the world on TBA, how much content gets duped? Perhaps we should let teams/people "suggest" their "primary" social media accounts to embed? Maybe we limit it to one, or two, or we don't limit it at all, but we allow the teams to self-filter the embed, but still list all their accounts for those potential "followers" on those services.

jaredhasenklein:
Could we do a checkbox for each social media the team has and only show those in the feed? I don't know how doable that is but from a user experience, I think it would be very nice.

fangeugene:
I'm personally not a fan of displaying any feed on team pages. It just adds to the clutter. I feel like links are enough for discovery, and we can pull things like profile photos via APIs to make team pages feel more customized.

synth3tk:
I like the idea of having "featured" accounts displayed at first, then maybe providing an enhanced tooltip for "More accounts" that lists the whole burrito. We should then allow teams to choose their featured accounts, in case they focus heavily on two or three accounts but not the rest.

As we discuss it more, I'm definitely finding widget embeds of any kind less appealing. If we must, I'd back @jaredhasenklein's idea of giving teams the option of which one widget to use.

jaredhasenklein:
My question becomes how does control over that work? What if someone says they are from a team they are not on? Or two individuals from the same team put different choice? 

Wouldn't there need to be a way to verify "owenership"?

And yes...as hearing more and more embedding everything would probably be pretty clutured.

bdaroz:
To some extent we already deal with that as edits/updates are done via "suggestions". I admit, it's far from perfect, but if embedding feeds is decided as the way to go (and I'm not convinced), it's a better solution than just embedding everything.

That said, I'm also not sure letting the visitor select which feeds to follow would be very helpful. If the team's primary feed is via Instagram, for example, and they parrott links to twitter, a user who just sees the Twitter feed isn't seeing everything, or worse, the user who selects the rarely-used FB stream.

jaredhasenklein:
That is fair...maybe if there was a request form, there could be a box to certify that they are a member of the team, etc. to deter some people.

synth3tk:
That's definitely a discussion I think we should have at some point as we add more and more data about teams. I'd be a fan of giving them some control over what's displayed and featured. We just need to make sure we properly vet them, as Jared brought up. The only issue is the time and effort required on our end to maintain this system for thousands and thousands of teams. Again, something we should consider outside of this issue.

As for now, I think the suggestions system would be a crude-yet-functional way to handle it.

fangeugene:
Or just feature YouTube, Facebook, Twitter by default, and make the rest expandable. Should cover most use cases.

jaredhasenklein:
Just quickly on David's comment, I agree that teams having more control would be very nice in general. For example, we had a weird primary picture that got uploaded by mistake and had to contact Eugene via Facebook to change it.

fangeugene:
On that note , were planning to add suggestions for removal.

jaredhasenklein:
Nice

phil-lopreiato:
> Or just feature YouTube, Facebook, Twitter by default, and make the rest expandable. Should cover most use cases.

Seems reasonable to me

synth3tk:
Or suggestions for removal. That definitely works too. Do we have that in place for changing primaries photos yet? I get a few pings on the FRC Discord over that one, too.

jaredhasenklein:
Primary photo ends up making a big difference for me personally at least. I helped build on the [Messenger FRC chatbot](https://github.com/FRC-Bot/FRCBot-ChatBot) so that when requesting `picture <<year>> <<team>>` it responds with the primary photo - useful for scouting amoung other things.

technonerdz:
I am pretty busy this month and I have a bunch of things to do on other projects before kickoff. I understand that this is too much social media and it would look weird on some team pages. We should maybe keep this pull request on hold and see if we find a better solution in the future.

runneals:
Would it be worth it to remove the names entirely and replace them with clean looking graphics? If someone wants to really follow them, they would have to click on over anyway, right? This would also reduce real estate needed to display social media accounts.
@tsteward did a good job on our team's site:
Non-Hover
And below is one when you hover over it...
On Hover

PR #1707, opened 2016-12-04
Gameday URL Aliases

phil-lopreiato:
Allows us to have urls like /gameday/fun that automatically embed the FUN stream. The pull:
 - Tweaks the structure of the GD sitevar to store special webcasts and aliases
 - Adds an admin panel page to create/delete aliases & special webcasts
 - Make a request handler for /gameday/blah and /watch/blah to do URL transforms

screenshot from 2016-12-03 20-25-57


templates/admin/gameday_dashboard.html
```diff
@@ -0,0 +1,80 @@
...
+        <input type="hidden" name="webcast_key" value="{{ webcast.key_name }}" />
+        <button class="btn btn-danger" type="submit"><span class="glyphicon glyphicon-trash"></span> Delete</button>
+    </form></td>
+</tr>
+{% endfor %}
+
+<!-- Row to add another special webcast -->
+<tr>
+    <td><input type="text" class="form-control" name="webcast_name" form="add_webcast" placeholder="Display Name"/></td>
+    <td><input type="text" class="form-control" name="webcast_type" form="add_webcast" placeholder="Stream Type"/></td>
```
fangeugene: This should be a dropdown. Look at the existing stuff.

phil-lopreiato:
Updated with a few things:
 - Use dropdown for webcast types
 - Added event and team watch aliases (/watch/2016cmp and /watch/frc1124). Goes to blank gameday if something isn't valid
 - Put gameday controllers in their own file

screenshot from 2016-12-03 21-26-53

PR #1720, opened 2016-12-22
[GD2] Implemented remaining streaming providers

nwalters512:
More work on the march towards feature parity with GD1!

* Adds streaming support for HTML5 (HLS), Livestream, Iframe, Dacast, and RTMP streams
* Migrated HLS and RTMP streams to both use [video.js](http://videojs.com/)
* Removed support for MMS streams
* Added a nice "webcast not supported" message to handle the case where we accidentally set a bad type in the sitevar

Iframes were kind of weird to get working: I either ended up with weird double-escaping if I use the `safe` keyword in the template file, or escaped html entities without it. I ended up going with the latter and just regex-replacing the escaped entities on the client. @fangeugene want to see if you can figure this one out?

### DIFF
--- a/react/gameday2/components/embeds/EmbedIframe.js
+++ b/react/gameday2/components/embeds/EmbedIframe.js
@@ -1,0 +1,33 @@
+import React from 'react'
+import { webcastPropType } from '../../utils/webcastUtils'
+
+const EmbedIframe = (props) => {
+  const divStyle = {
+    width: '100%',
+    height: '100%',
+  }
+
+  let iframeMarkup = props.webcast.channel
+  iframeMarkup = iframeMarkup.replace(/&lt;/, '<')
+  iframeMarkup = iframeMarkup.replace(/&gt;/, '>')
+  const markup = {
+    __html: iframeMarkup,
+  }
+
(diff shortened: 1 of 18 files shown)


react/gameday2/components/embeds/EmbedIframe.js
```diff
@@ -0,0 +1,33 @@
...
+import { webcastPropType } from '../../utils/webcastUtils'
+
+const EmbedIframe = (props) => {
+  const divStyle = {
+    width: '100%',
+    height: '100%',
+  }
+
+  let iframeMarkup = props.webcast.channel
+  iframeMarkup = iframeMarkup.replace(/&lt;/, '<')
```
nwalters512: ¯\_(ツ)_/¯

phil-lopreiato:
Hey @nwalters512 can you merge in master here? I think we have some livestream events coming up

nwalters512:
Yeah, I can get on that this coming week. It's my school's hackathon this weekend so I'm kinda busy :)

fangeugene:
Build GameDay Lite for your hackathon.

PR #1728, opened 2017-01-04
Apiv3 WIP

fangeugene:
No authentication, no analytics tracking, only subset of endpoints work.

I want to merge this partial PR after I clean up caching and model_to_dict so we can play with it and do some benchmarks on it.

The idea behind caching for APIv3 is to cache DatabaseQueries in a way that minimizes work to convert the result of these queries into something JSON serializable for output, and then don't cache the JSON output beyond normal memcache and edge cache. This way, there's only one place we have to go to evict cache instead of 2 with APIv2.

An efficient way to do this is to cache not the full models in the DatabaseQueries, but instead dict-ified versions of the models. This is because object.property takes much longer than dict.get() in Python. In benchmarks run on the server, filtering a subset of properties from a list of 100+ matches takes on the order of hundreds of milliseconds, whereas filtering a subset of entries from a dict takes on the order of 10 milliseconds. Another advantage is that computed properties (like match.winning_alliance) can be cached.

For backwards compatibility, DatabaseQueries have been extended to cache both full models and dict-ified versions of the models. The rest of the site should transition to use the dict version and the full model version will be deprecated in the future.


fangeugene:
#yolomerge will finish in another PR

PR #1747, opened 2017-01-16
Base District Model

phil-lopreiato:
This is backend improvement towards apiv3 (a place to store district rankings).

Store one `District` model per year/region. This also gives us a persistent place to store stuff. Will deprecate the `DistrictType` constants and manage everything from the admin panel instead (this is more scalable).

This pull provides the base model and automated creation from the FMS Datafeed. The constants we'll need to parse event names that come back from elasticsearch need to be manually input. However, the code will keep the constants from the previous year, if it exists.

Migration path after merge:
1. Run some script that backfills these based on existing DCMPs and auto-fills the names from the existing constants file
2. Remove constants file, deprecate `Event.district_type_enum` and `DistrictTeam.district` and use the key links instead for rendering.
3. Profit

### DIFF
--- a/database/district_query.py
+++ b/database/district_query.py
@@ -3,6 +3,7 @@ from google.appengine.ext import ndb
 from consts.event_type import EventType
 from database.database_query import DatabaseQuery
 from database.dict_converters.district_converter import DistrictListConverter
+from models.district import District
 from models.event import Event
 
 
@@ -19,3 +20,16 @@ class DistrictListQuery(DatabaseQuery):
             Event.event_type_enum == EventType.DISTRICT_CMP).fetch_async(keys_only=True)
         events = yield ndb.get_multi_async(all_cmp_event_keys)
         raise ndb.Return(events)
+
+
+class DistrictsInYearQuery(DatabaseQuery):
+    CACHE_VERSION = 0
+    CACHE_KEY_FORMAT = "districts_in_year_{}"  # (year)
+    DICT_CONVERTER = None  # For now (TODO)
+
+    @ndb.tasklet
+    def _query_async(self):
+        year = self._query_args[0]
+        keys_future = yield District.query(District.year == year).fetch_async(keys_only=True)
+        districts = yield ndb.get_multi_async(keys_future)
+        raise ndb.Return(districts)
--- a/database/district_query.py
+++ b/database/district_query.py
@@ -33,3 +33,17 @@ class DistrictsInYearQuery(DatabaseQuery):
         district_keys = yield District.query(District.year == year).fetch_async(keys_only=True)
         districts = yield ndb.get_multi_async(district_keys)
         raise ndb.Return(districts)
+
+
+class DistrictHistoryQuery(DatabaseQuery):
+    CACHE_VERSION = 0
+    CACHE_KEY_FORMAT = "district_history_{}"  # (abbreviation)
+    DICT_CONVERTER = None  # For now (TODO)
+
+    @ndb.tasklet
+    def _query_async(self):
+        abbreviation = self._query_args[0]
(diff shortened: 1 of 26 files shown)


fangeugene:
Do we have to enter the elastic search name each year for the same district, or can it draw from previous years?

models/district.py
```diff
@@ -0,0 +1,24 @@
...
+class District(ndb.Model):
+    """
+    One instance of a district in a year. Here, we store info about a district and in-season data
+    (like district rankings)
+    """
+    year = ndb.IntegerProperty()
+    abbreviation = ndb.StringProperty()
+    display_name = ndb.StringProperty()  # This is what we'll show on the TBA site
+    elasticsearch_name = ndb.StringProperty()  # These names are in the event's name as returned by FRC Elasticsearch
+
```
fangeugene: This needs the constructor like

```
def __init__(self, *args, **kw):
        # store set of affected references referenced keys for cache clearing
        # keys must be model properties
        self._affected_references = {
            'year': set(),
        }
```

models/event.py
```diff
@@ -25,7 +26,8 @@ class Event(ndb.Model):
     short_name = ndb.StringProperty(indexed=False)  # Should not contain "Regional" or "Division", like "Hartford"
     event_short = ndb.StringProperty(required=True, indexed=False)  # Smaller abbreviation like "CT"
     year = ndb.IntegerProperty(required=True)
-    event_district_enum = ndb.IntegerProperty(default=DistrictType.NO_DISTRICT)
+    event_district_enum = ndb.IntegerProperty(default=DistrictType.NO_DISTRICT)  # Deprecated, use district_key instead
+    district_key = ndb.KeyProperty(kind=District)
```
fangeugene: Probably need to add `'district_key: set(),` to the constructor below.

phil-lopreiato:
> Do we have to enter the elastic search name each year for the same district, or can it draw from previous years?

It will be drawn from the previous year, see [here](https://github.com/the-blue-alliance/the-blue-alliance/pull/1747/files#diff-20738055d2c26cb94e864c58462bd267R22). For the initial migration, we can use constants from `DistrictType` before we deprecate it.

> Do you plan on adding cache clearing tests in this PR?

Yes, thanks for reminding me we had tests for that :)

fangeugene:
> It will be drawn from the previous year, see [here](https://github.com/the-blue-alliance/the-blue-alliance/pull/1747/files#diff-20738055d2c26cb94e864c58462bd267R22). For the initial migration, we can use constants from `DistrictType` before we deprecate it.

Awesome

models/event.py
phil-lopreiato: Eventually, yes, but I'll do that when I do the functional migration

models/district.py
phil-lopreiato: :+1:

database/district_query.py
```diff
@@ -19,3 +20,16 @@ def _query_async(self):
...
+
+class DistrictsInYearQuery(DatabaseQuery):
+    CACHE_VERSION = 0
+    CACHE_KEY_FORMAT = "districts_in_year_{}"  # (year)
+    DICT_CONVERTER = None  # For now (TODO)
+
+    @ndb.tasklet
+    def _query_async(self):
+        year = self._query_args[0]
+        keys_future = yield District.query(District.year == year).fetch_async(keys_only=True)
```
fangeugene: This is just `keys`, not `keys_future`. `yield` resolves futures.

fangeugene: This is just `keys`, not `keys_future`. `yield` resolves futures.

fangeugene:
LGTM, merge whenever.

PR #1755, opened 2017-01-25
Surrogates

fangeugene:
Save and display surrogates.
TODO: backfill surrogates for old events

New alliances format: 
```
{
  "red": {
    "teams": ["frc177", "frc195", "frc125"],
    "surrogates": ["frc177", "frc195"],
    "score": 25
  },
  "blue": {
    "teams": ["frc433", "frc254", "frc222"],
    "surrogates": ["frc433"],
    "score": 12
  }
}
```

image


phil-lopreiato:
Wooho! I think we can only backfill as far as the FRC API goes (so 2015 and 2016), so at least that shouldn't be too difficult

fangeugene:
What about backfill based on 3rd played match? Not sure how far back we can
go, or whether to do it for offseasons.

phil-lopreiato:
Oh, good point. If there is any variation in the number of matches played by some teams, then we know there are surrogates (which *should* always be the third match). We'd have to figure out how far back the surrogate policy goes (as far as we can?).

PR #1763, opened 2017-02-09
Auto-Generated Swagger Spec

phil-lopreiato:
**[WIP]**
Adds a paver task `paver swagger` that you can use to build the swagger spec file. This isn't tracked in git (so we can always ensure it's up to date).


phil-lopreiato:
Probably not worthwhile to do this

PR #1775, opened 2017-02-21
Gameday2 match ticker

fangeugene:
image

PR #1776, opened 2017-02-21
Gameday2 Live Updating Webcasts

fangeugene:
Special webcasts are still loaded through the template, live event webcasts are loaded through Firebase, so users don't need to refresh the page when webcasts get added/changed anymore!

Merge after #1775


nwalters512:
Any reason we can't do special webcasts over Firebase as well?

fangeugene:
Being lazy for now.

react/gameday2/reducers/webcastsById.js
```diff
@@ -9,37 +9,41 @@ const getWebcastsFromRawWebcasts = (webcasts) => {
...
-        name = `${name} ${index + 1}`
-      }
-      const id = getWebcastId(event.key, index)
+  if (specialWebcasts) {
+    specialWebcasts.forEach((webcast, index) => {
+      const id = getWebcastId(webcast.key_name, 0)
       webcastsById[id] = {
-        key: event.key,
-        num: index,
+        key: webcast.key_name,
```
fangeugene: Git sometimes isn't smart about diffs when you indent things...

PR #1781, opened 2017-02-24
Event Media Suggestions

phil-lopreiato:
screenshot from 2017-02-23 20-22-14

### DIFF
--- a/consts/account_permissions.py
+++ b/consts/account_permissions.py
@@ -3,6 +3,7 @@ class AccountPermissions(object):
     REVIEW_OFFSEASON_EVENTS = 2
     REVIEW_APIWRITE = 3
     REVIEW_DESIGNS = 4
+    REVIEW_EVENT_MEDIA = 5
 
     permissions = {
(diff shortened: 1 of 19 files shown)


fangeugene:
What does the suggestion page look like? What are our guidelines for what we will accept/reject?

phil-lopreiato:
The suggestion page looks just like the one for suggesting team media. Here's the review page
screenshot from 2017-02-24 10-10-28

fangeugene:
The suggestion page, not the review page.

phil-lopreiato:
screenshot from 2017-02-24 10-11-59

fangeugene:
Ah okay. I think we should be more clear about what we want media of, since I don't want it to just be a collection of everything people have taken. Unless we want to do that, in which case we need to change how we display things.

phil-lopreiato:
Updated Screenshots:
screenshot from 2017-02-24 11-15-53
screenshot from 2017-02-24 11-16-06
screenshot from 2017-02-24 11-20-03

PR #1782, opened 2017-02-24
Predict Upcoming Match Times

phil-lopreiato:
Adds a cronjob that calculates predicted times for the next 10 matches at each currently live event. The algorithms works like so:

1. Compute Cycle Time
   - This is the 30% percentile of all cycle times (biased 70/30 towards the schedule) from the same day, discarding outliers.

2. Assuming that cycle time, start with the last know match start time and predict the times for the next 10 matches.

Some data:
This was an event that ran ahead of schedule, for the most part:
2016ytr

This was an event that was very behind, for the most part:
2016nyny

This was an event that was all over the place:
2016necmp

Overall, the predicted times are rarely more than 3 minutes late when the event is running faster than anticipated and are always more (usually very) accurate for other cases.

### DIFF
--- a/controllers/cron_controller.py
+++ b/controllers/cron_controller.py
@@ -2,6 +2,7 @@ import datetime
 import logging
 import os
 import json
+import pytz
 
 from google.appengine.api import taskqueue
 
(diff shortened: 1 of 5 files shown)


fangeugene:
This looks fine. We'll see how it works. Curious as to what causes the up down zig zag of 2016nytr.

phil-lopreiato:
I think because the algorithm is very aggressive in removing/deemphasizing outliers, it may thrash a little if the first few matches are slow cycles. That'll cause it to have a small sample size of valid cycles until the event gets underway

PR #1790, opened 2017-03-01
Live webcasts

fangeugene:
If it's Twitch, we know if the webcast is live of not!

image


tsteward:
Can we do the samething for youtube?
http://stackoverflow.com/questions/32454238/how-to-check-if-youtube-channel-is-streaming-live
ustream?
http://stackoverflow.com/questions/5011750/detect-if-ustream-is-broadcasting-on-air
livestream?
https://livestream.com/developers/docs/api/#event (See 'isLive' field)

PR #1818, opened 2017-03-13
Improve event media examples

tsteward:
This changes the example youtube video on the event media page to be a video that is a good example of event media (instead of a robot reveal).

The placeholder text for the suggestion media url box was a chiefdelphi photo link. We currently don't support chiefdelphi photos for event media, so this is misleading. This changes to placeholder text to be a link for a youtube video.
capture


fangeugene:
I see what you did there.

PR #1819, opened 2017-03-13
add a horizontal split view to gameday

tsteward:
On some smaller (more square) screens when both the chat and social side bars are enabled, the videos are quite small in split view. This adds a horizontal split view for this situation.

This adds the horizontal split view as layout 2 and shifts the rest down. Will this be a problem for existing links to gameday? Should the new layout be 9 instead?

image
image
image


nwalters512:
Yeah, it should be 9. We should separate display order in the layout selection panel from the number of the view; number is essentially just an enum and shouldn't be used by users.

Maybe have a constant LAYOUT_DISPLAY_ORDER that's an array where each element is the number of the layout that should be shown at that index?

tsteward:
I made the new layout 9 and defined the order in a constant as you described.

Closes #1820

nwalters512:
Lgtm! :+1: :shipit:

phil-lopreiato:
It's live! Although it may be cached for a bit

PR #1841, opened 2017-03-18
Favicon with updated logo

jaredhasenklein:
Converted the new logo to a favicon. It doesn't have a transparent background like the previous one because there isn't a great configuration that I thought of to have transparency and maintain the color scheme.


fangeugene:
Can you try using the circular logo as well as just the lamp part (changed to blue with no background)? Also, adding screenshots would help.

jaredhasenklein:
I played in AI on PS so here are three different versions with previews of the icon in different contexts (there is a third party app that does a great job of that part).

https://drive.google.com/drive/folders/0B7KVTj--N5d4SUJIb0ZFT0Z6bTA

fangeugene:
That's awesome. I think I'm a fan of the circle. Do others have opinions?

brandondean:
I'm a fan of the circle.

nwalters512:
Yeah, the circle looks great!

jaredhasenklein:
Circle has been uploaded to the PR

PR #1842, opened 2017-03-18
Add event status to event table

tsteward:
This adds the status for each event to the event table to give an overview of how all of the events are progressing in one place.
screencapture-localhost-8088-1489855741573

### DIFF
--- a/templates/event_partials/event_table.html
+++ b/templates/event_partials/event_table.html
@@ -15,7 +15,9 @@
             <br>
             <small>{{event.city_state_country}}</small>
             <br>
-            {% if event.now and event.next_match %}
+          {% endif %}
+          {% if event.now %}
+            {% if event.next_match %}
               <b>Next Match: </b> {{event.next_match.verbose_name}}
             {% else %}
               {% if not event.alliance_selections and event.previous_match %}
--- a/templates/event_partials/event_table.html
+++ b/templates/event_partials/event_table.html
@@ -14,6 +14,14 @@
           {% if event.city_state_country %}
             <br>
             <small>{{event.city_state_country}}</small>
+            <br>
+            {% if event.now and event.next_match %}
+              <b>Next Match: </b> {{event.next_match.verbose_name}}
+            {% else %}
+              {% if not event.alliance_selections and event.previous_match %}
+                <b>Next: </b> Alliance Selections
+              {% endif %}
+            {% endif %}
           {% endif %}
         </td>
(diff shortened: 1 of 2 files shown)


tsteward:
The `event_table` template is also used in the event list page. I don't think we want also do this on that page because it is cached for longer cache expiration (I think). Would there be a way to make the `event_table` template only show details on main page.

templates/event_partials/event_table.html
```diff
@@ -14,6 +14,14 @@
           {% if event.city_state_country %}
             <br>
             <small>{{event.city_state_country}}</small>
+            <br>
+            {% if event.now and event.next_match %}
+              <b>Next Match: </b> {{event.next_match.verbose_name}}
+            {% else %}
+              {% if not event.alliance_selections and event.previous_match %}
```
fangeugene: The entire logic should be wrapped in event.now, not just the first clause.

fangeugene:
The cache is short on the event list.

templates/event_partials/event_table.html
```diff
@@ -14,6 +14,16 @@
           {% if event.city_state_country %}
             <br>
             <small>{{event.city_state_country}}</small>
+            <br>
```
fangeugene: This `<br>` should be wrapped inside the if statements below, before each `<b>`

tsteward: Yep just noticed

fangeugene:
This looks good. Do you have any other changes you'd like to make?

tsteward:
No

jaredhasenklein:
I just looked at the live home page for the first time since this change was made. I like that the information is made available however, I think that the formatting needs some work.

The event name and next match are the same font size whereas the location is 1.5 points smaller. This 
"Next Match" text is bold and given that it is the same size as the event name, I think it either needs to be unbolded and the event should be bolded OR the entire Next Match line should be reduced to the same size as the location.

Thoughts?

jaredhasenklein:
screenshot 315

@tsteward @fangeugene

fangeugene:
I wouldn't mind small and no bold.

tsteward:
@jaredhasenklein Do you want to go ahead and make the change?

jaredhasenklein:
Doing it now

jaredhasenklein:
Proposed in #1846

jaredhasenklein:
should be closed

PR #1846, opened 2017-03-18
Event table formatting

jaredhasenklein:
See #1842

### DIFF
--- a/templates/event_partials/event_table.html
+++ b/templates/event_partials/event_table.html
@@ -18,11 +18,11 @@
           {% if event.now %}
             {% if event.next_match %}
               <br>
-              <b>Next Match: </b> {{event.next_match.verbose_name}}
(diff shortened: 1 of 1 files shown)


fangeugene:
Can  you fix the tabbing? Use spaces.

jaredhasenklein:
Do you just mean the extra space?

fangeugene:
No, the indentation is off. It should be aligned with `<br>`

jaredhasenklein:
oops. yes fixed

jaredhasenklein:
One more thought on this topic:

We may want to list event status other than the next match or alliance selection. It would be helpful to know if the event is complete (final matches played), if the event is offline, or if matches haven't started yet. I'm definitely not a good person to code that but maybe someone wants to

PR #1855, opened 2017-03-22
Gracefully Handle Pre-Divisions CMP Registration Hackery

phil-lopreiato:
Currently, all teams attending championship show as attending Einstein. This adds a sitevar that allows the following:

 - skipping storing divisions (so they won't appear until there are actually teams)
 - overriding the Einstein names (so we can render them as FIRST Champinoship until divisions are announced)
- skip changing Einstein dates


fangeugene:
LGTM

jaredhasenklein:
@phil-lopreiato Now that divisions are released, it still says `FIRST Championship (Houston)`

Issue #1867, opened 2017-03-31
Mobile Version: "Watch Now" button is hidden 


sorianog:
The Mobile Version of the site seems to not display the "Watch Now" button for viewing an event's stream. Switching to the Desktop Version shows the button fine.

fangeugene:
It was purposely hidden in the past due to space reasons and because GameDay flat out didn't work on mobile. Now that it does, maybe we can bring the button back.

sorianog:
#bringthebuttonback Would you like some help with that?

fangeugene:
Yes please!

sorianog:
Where should I start? I'm new to the code base :/

sorianog:
In file ```event_details.html```, I see the following code that I think is related to this issue:

```javascript
          <div class="hidden-xs pull-right">
            {% if event.webcast %}
              {% if event.webcast_status == 'offline' %}
                <a class="btn btn-default btn-xs" href="{{event.gameday_url}}" target="_blank"><span class="glyphicon glyphicon-facetime-video"></span> Offline</a>
              {% else %}
                <a class="btn btn-success btn-xs" href="{{event.gameday_url}}" target="_blank"><span class="glyphicon glyphicon-facetime-video"></span> Watch Now</a>
              {% endif %}
            {% else %}
              <a class="btn btn-xs btn-default" href="/suggest/event/webcast?event_key={{event.key_name}}" target="_blank">Add webcast</a>
            {% endif %}
          </div>
```

What would need to be changed?

fangeugene:
Take a look at http://getbootstrap.com/css/#responsive-utilities

the `hidden-` and `visible-` classes determine what shows up on various screen sizes. In this case, `hidden-xs` is saying "don't show this element on extra small devices (phones)" If you remove that class, it will show up on phones.

sorianog:
Ah, that makes sense! No wonder the "xs" is there. I attempted to reproduce this issue when trying to spoof Android mobile-agent mode in Chrome but I was unsuccessful. When I visited an event on my Google Pixel though, I was able to reproduce this issue. How do you and the others test on mobile devices?

fangeugene:
We use responsive design, so different layouts are determined by screen size, not user agent. Resizing a window on desktop is a good way to test for "mobile."

sorianog:
Interesting. So when the browser width is <768px, that is when the button disappears (tested locally and saw that the docs say that too). Good to know.

PR #1868, opened 2017-04-01
#1867 "Watch Now" visibility fix for Extra Small Devices (<768px)

sorianog:
Removed "hidden-xs" property from event_details.html to allow for the stream button to display on mobile devices where browser width <768px

See: http://getbootstrap.com/css/#responsive-utilities

PR for #1867

### DIFF
--- a/templates_jinja2/event_details.html
+++ b/templates_jinja2/event_details.html
@@ -72,7 +72,7 @@
     <div class="col-sm-12">
       <div class="panel panel-default">
         <div class="panel-heading">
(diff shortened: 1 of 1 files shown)


gregmarra:
Thanks @sorianog ! In the future, it'd be best if you could upload a screenshot of how this looks, to confirm it looks as expected.

sorianog:
@gregmarra here is the screenshot verifying the fix! 
tba-hidden-xs-fix

gregmarra:
Great!

PR #1870, opened 2017-04-03
Fix README setup section numbering

andrewda:
The numbering in the README's setup section is a bit off thanks to a recent GitHub Markdown update. This should fix that.

### Before:
Before

### After:
After


phil-lopreiato:
Thanks for the change!

PR #1873, opened 2017-04-04
Fix OPR page header

andrewda:
This is a minor styling fix to the header of the OPR page.

### Before
before

### After
after


phil-lopreiato:
LGTM :+1:

PR #1876, opened 2017-04-05
Add team media suggestion API endpoint

### LINKED ISSUE #1037: [concept] Authenticated API for writing Suggestions
Allow people to write Media suggestions, Match Video suggestions via API.

In particular, how would we enable someone to build an app that let you log in to TBA, log in to Chief Delphi (or maybe this is an example of where we really do need imgur support), and upload photos straight from your phone. Someone could snap a photo of every robot in a pit and blast it to TBA right from their phone.

andrewda:
This will add an API endpoint for suggesting team media. A developer would make a POST request to `/api/v3/team/<team>/suggest/media/<year>` (subject to change) with the POST data `{ media_url: "some media link" }` in order to queue a suggestion. Should resolve #1037.

I've been playing with this quite a bit, and from my tests it should be working :)

### DIFF
--- a/apiv3_main.py
+++ b/apiv3_main.py
@@ -69,7 +69,7 @@ app = webapp2.WSGIApplication([
         atc.ApiTeamYearMediaController, methods=['GET', 'OPTIONS']),
     # Team Media Suggestions
     webapp2.Route(r'/api/v3/team/<team_key:>/suggest/media/<year:([0-9]+)>',
-        asgc.ApiSuggestTeamMediaController, methods=['POST', 'GET', 'OPTIONS']),
+        asgc.ApiSuggestTeamMediaController, methods=['POST', 'OPTIONS']),
     # Event List
     webapp2.Route(r'/api/v3/events/<year:([0-9]+)>',
         aec.ApiEventListController, methods=['GET', 'OPTIONS']),
--- a/apiv3_main.py
+++ b/apiv3_main.py
@@ -7,6 +7,7 @@ from controllers.apiv3 import api_district_controller as adc
 from controllers.apiv3 import api_event_controller as aec
 from controllers.apiv3 import api_match_controller as amc
 from controllers.apiv3 import api_team_controller as atc
+from controllers.apiv3 import api_suggest_controller as asgc
 
 # Ensure that APIv3 routes include OPTIONS method for CORS preflight compatibility
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests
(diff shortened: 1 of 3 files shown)


controllers/apiv3/api_base_controller.py
```diff
@@ -122,9 +122,21 @@ def _validate_tba_auth_key(self):
...
+                self.auth_owner = auth.owner
                 self.auth_description = auth.description
                 logging.info("Auth owner: {}, X-TBA-Auth-Key: {}".format(self.auth_owner, x_tba_auth_key))
             else:
                 self._errors = json.dumps({"Error": "X-TBA-Auth-Key is invalid. Please get an access key at http://www.thebluealliance.com/account."})
                 self.abort(400)
+
+class ApiSuggestBaseController(ApiBaseController):
+    def post(self, *args, **kw):
+        super(ApiSuggestBaseController, self).get(*args, **kw)
```
phil-lopreiato: So, we don't want the outputs of this to be cached - both in `memcache` on App Engine and with the `Cache-Control` headers in the response. This will happen by default here - take a look at [`CacheableHandler`](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/controllers/base_controller.py#L22). 

If you want to continue sharing the same code as the `GET` handler, add a class variable flag to control when we call `super(ApiBaseController, self).get(*args, **kw)`, or duplicate that code here (but without the `super` call).

@fangeugene any preferences as to how we handle this?

apiv3_main.py
```diff
@@ -66,6 +67,9 @@
         atc.ApiTeamYearMatchesController, methods=['GET', 'OPTIONS']),
     webapp2.Route(r'/api/v3/team/<team_key:>/media/<year:([0-9]+)>',
         atc.ApiTeamYearMediaController, methods=['GET', 'OPTIONS']),
+    # Team Media Suggestions
+    webapp2.Route(r'/api/v3/team/<team_key:>/suggest/media/<year:([0-9]+)>',
```
SUPERCILEX: @andrewda this should be changed to `api/v3/suggest/...`, right?

controllers/apiv3/api_base_controller.py
andrewda: Would it be better to add a `get_no_cache` method in `CacheableHandler`?

phil-lopreiato: I think I'd rather add a `post` method to `ApiBaseController` and duplicate the contents of `get`, just without the `super` call. You'd still have to call `_render` in the subclass, though. [See here](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/controllers/base_controller.py#L74)

andrewda:
Alright, those issues should be fixed!

SUPERCILEX:
@andrewda Thanks for this! My [work on Robot Scouter](https://github.com/SUPERCILEX/Robot-Scouter/pull/107) is basically done, just waiting on this PR.

SUPERCILEX:
@fangeugene Can you please take a look at this? I have the start of Worlds as a deadline for `v1.1` of Robot Scouter and the major PR for that is waiting on this one.

fangeugene:
This has been deployed on production. Keep an eye out for funny caching issues, since ApiBaseController extends CacheableHandler, but I think it will be fine because caching only happens for GET requests, not POST requests.

SUPERCILEX:
@fangeugene Thank you!!!!!!!!!

PR #1878, opened 2017-04-06
Added Facebook event field to offseason event form

### LINKED ISSUE #1875: Add 'facebook event' to Offseason Request form

verycumbersome:
This should fix #1875.


models/event.py
```diff
@@ -47,6 +47,7 @@ class Event(ndb.Model):
     facebook_eid = ndb.StringProperty(indexed=False)  # from Facebook
```
fangeugene: `facebook_eid` is already present for this -- there's no need to add `facebook_event`. EID = Event ID.

PR #1882, opened 2017-04-12
Return error code 404 if team doesn't exist or isn't competing this year instead of 400 for suggesting media

SUPERCILEX:
I don't think 400 really makes sense in this case since the team doesn't exist or isn't competing. @fangeugene @phil-lopreiato what do you guys think?

### DIFF
--- a/controllers/apiv3/api_base_controller.py
+++ b/controllers/apiv3/api_base_controller.py
@@ -80,7 +80,7 @@ class ApiBaseController(CacheableHandler):
         self._validate_tba_auth_key()
         self._errors = ValidationHelper.validate_request(self)
         if self._errors:
-            self.abort(400)
+            self.abort(404)
 
         self._track_call(*args, **kw)
         super(ApiBaseController, self).get(*args, **kw)
@@ -91,7 +91,7 @@ class ApiBaseController(CacheableHandler):
         self._validate_tba_auth_key()
         self._errors = ValidationHelper.validate_request(self)
         if self._errors:
-            self.abort(400)
+            self.abort(404)
 
         rendered = self._render(*args, **kw)
         self._track_call(*args, **kw)
--- a/controllers/apiv3/api_suggest_controller.py
+++ b/controllers/apiv3/api_suggest_controller.py
@@ -27,7 +27,7 @@ class ApiSuggestTeamMediaController(ApiBaseController):
                     "success": True
                 }
             else:
-                message ={
+                message = {
                     "success": False,
                     "message": status
                 }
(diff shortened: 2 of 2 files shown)


andrewda:
I think we should be as consistent as possible throughout the API. Currently, using an invalid team in `/v3/team/<team_id>` will result in a 400 with the message `{'Errors': [{'team_id': 'team id <team_id> does not exist'}]}` (e.g. https://www.thebluealliance.com/api/v3/team/frc2)

phil-lopreiato:
I'd prefer throwing a 404 in this case as well. That seems to make more sense

SUPERCILEX:
@andrewda I agree with @phil-lopreiato so I think we shouldn't return a 400 when the team doesn't exist and return a 404 instead (which is also consistent with `v2` and just makes more sense).

controllers/apiv3/api_base_controller.py
```diff
@@ -91,7 +91,7 @@ def post(self, *args, **kw):
         self._validate_tba_auth_key()
         self._errors = ValidationHelper.validate_request(self)
         if self._errors:
-            self.abort(400)
+            self.abort(404)
```
SUPERCILEX: @phil-lopreiato @fangeugene [`validate_request`](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/helpers/validation_helper.py#L36-L98) is all about checking whether or not a key exists so this should be fine.

andrewda:
Yep! Looks good!

controllers/apiv3/api_base_controller.py
fangeugene: I believe the two `validate_request` should be 400's

SUPERCILEX: @fangeugene What's your reasoning? It doesn't make sense to me because a `400` means "bad request". For example, 400 is returned [here](https://github.com/SUPERCILEX/the-blue-alliance/blob/fae50437048384a6a12b8929c2f4d95e41d491e4/controllers/apiv3/api_base_controller.py#L132) where it makes sense because the auth header is missing, hence a "bad request".

However, if a team, event, etc. doesn't exist, it's not a bad request, but instead a `404` which means "Not Found Error". That makes way more sense because it's what's actually happening: the team, event, etc. wasn't found or doesn't exist.

fangeugene: Oh right, how about 401 then?

SUPERCILEX: @fangeugene That makes sense for [another PR](https://github.com/the-blue-alliance/the-blue-alliance/pull/1884), but not for this one. Have a look at the [`validate_request`](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/helpers/validation_helper.py#L36-L98) method, it's validating whethor or not something exists hence returing a `404` if it fails.

fangeugene: Ohh I forgot what validate_request does. Yes 404 makes sense here.

SUPERCILEX: @fangeugene @phil-lopreiato Awesome, are we good to merge!? 😄

SUPERCILEX:
@fangeugene Thanks! Time to go update my Robot Scouter code 😄

Issue #1893, opened 2017-04-17
Include districts in search.


CarlColglazier:
I often use the district pages, but they are rather difficult to get to at the moment. Currently, the way I get to the North Carolina district page is searching "North Carolina", clicking the [first result](https://www.thebluealliance.com/event/2017nccmp) and then clicking [*North Carolina District*](North Carolina District) at the top of the page.

It would be useful to make district pages more accessible by making districts searchable (only events and teams are searchable at the moment).

fangeugene:
Implemented in #1896

PR #1896, opened 2017-04-17
Add districts to typeahead search.

CarlColglazier:
image

The district search works in a manner similar to event searching, except there is only one searchable entry per district. For example, the search for "North Carolina" leads to the [events list](https://www.thebluealliance.com/events/nc) for the current year. 

Implements #1893 

Edit: typo


fangeugene:
This is great.

PR #1897, opened 2017-04-17
Store Divisional Mappings

### LINKED ISSUE #1895: Implement Event Parent/Child Relationships
Now that divisions are coming to more places (thanks MSC), we should formalize their support to better show how teams qualified for the "Einstein" event and where division-winning teams go.

phil-lopreiato:
Adds keys for `parent_event` and `divisions` to the event model.

Works for Einstein(s) and MSC this year.

image

image

Closes #1895

### DIFF
--- a/database/dict_converters/event_converter.py
+++ b/database/dict_converters/event_converter.py
@@ -5,7 +5,7 @@ from database.dict_converters.district_converter import DistrictConverter
 
 class EventConverter(ConverterBase):
     SUBVERSIONS = {  # Increment every time a change to the dict is made
-        3: 4,
+        3: 5,
     }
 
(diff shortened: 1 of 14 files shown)


database/dict_converters/event_converter.py
```diff
@@ -30,9 +30,11 @@ def eventConverter_v3(cls, event):
             'event_code': event.event_short,
             'event_type': event.event_type_enum,
             'event_type_string': event.event_type_str,
+            'parent_event': event.parent_event.id() if event.parent_event else None,
```
fangeugene: I think these should be `parent_event_key` and `division_keys` for consistency.

PR #1910, opened 2017-04-24
Added Chairmans Honorable Mention to get a banner.

MC42:


phil-lopreiato:
It's not quite this simple, sadly. Prior recipients off the honorable mention didn't receive a banner, so we need to make a new award type `CHAIRMANS_FINALIST` that does. And we'll need to ensure the awards from `2017cmptx` get correctly attributed with the new type.

MC42:
Well damn, I was so excited any everything...

phil-lopreiato:
Sorry to be the bearer of bad news 😢

PR #1918, opened 2017-05-03
Link to FOC events page

jaredhasenklein:
Adds direct link to event page for FOC on templates/index_lhc_offseason.html


dracco1993:
I like the idea of linking this to the actual event, however I think you might want to make one small change.  We may actually want to put the link inside the `b` tag here, as with it just inside the `h5` there's a much larger click area (shown in green).  See the two images below:

Inside the `h5`:


Inside the `b`:
ss 2017-05-03 at 01 12 56

jaredhasenklein:
You are completly right. I actually didn't even intend to hyperlink the dates so I fixed it in the patch. 

If I remember correctly, just the event name is usually hyperlinked on the home page.

dracco1993:
Yeah, if you look at something like [/events](https://www.thebluealliance.com/events), it's just the name of the event that's linked.

PR #1919, opened 2017-05-03
First pass at adding a Chairman's Finalist Award

### LINKED ISSUE #1909: Blue Banner for CCA Finalist

dracco1993:
This is a WIP, but I'd like to get some input from others on what I've got so far.

I still need to 

> "[ensure the awards from `2017cmptx` get correctly attributed with the new type](https://github.com/the-blue-alliance/the-blue-alliance/pull/1910/#issuecomment-296501750)"

Fixes #1909

Edit:
This may actually contain all of the necessary code as it is right now?  I've not tested it against real data or anything, but all of the tests I've added are passing.


dracco1993:
Note:
Both `2017cmptx` and `2017cmpmo` should be attributed with the new award.

helpers/award_helper.py
```diff
@@ -25,7 +25,7 @@
 AWARD_MATCHING_STRINGS = [
     (AwardType.CHAIRMANS, (["chairman"], ["hon", "finalist"])),
     (AwardType.CHAIRMANS_HONORABLE_MENTION, (["chairman", "hon", "mention"], [])),
-    (AwardType.CHAIRMANS_HONORABLE_MENTION, (["chairman", "finalist"], [])),
+    (AwardType.CHAIRMANS_FINALIST, (["chairman", "finalist"], ["hon", "mention"])),
```
phil-lopreiato: Refresh my memory - why does the `["hon", "mention"]` need to be here?

fangeugene: It means it doesn't match those. Probably a non-issue since we're probably not going to be re-importing old awards.

dracco1993: It's the blacklist for searching, so I figured that if "hon" and "mention" were going to be for `CHAIRMANS_HONORABLE_MENTION`, you wouldn't want them to match for `CHAIRMANS_FINALIST`.
I'm totally open to changing this, I was just trying to make sure you'd get the right results when searching.

phil-lopreiato: Nope, that sounds right. I just didn't remember what the structure of this was.

fangeugene:
I think this works. @phil-lopreiato any comments? @dracco1993 any changes?

phil-lopreiato:
LGTM 👍

dracco1993:
I've still not been able to confirm it works, other than the tests, but if you're cool with it, I've got nothing else to change.  👍

phil-lopreiato:
I tested it locally on my machine, and everything worked!

PR #1929, opened 2017-05-08
Ignore matches without a score

### LINKED ISSUE #1223: Event Wizard - FMS Results input after Quarters returns frcundefined error

SUPERCILEX:
Fixes #1223

### DIFF
--- a/static/javascript/tba_js/eventwizard_apiwrite.js
+++ b/static/javascript/tba_js/eventwizard_apiwrite.js
@@ -205,11 +205,11 @@ $('#results_file').change(function(){
         $('#results_preview').html("<tr><th>Time</th><th>Match</th><th>TBA Key</th><th>Red 1</th><th>Red 2</th><th>Red 3</th><th>Blue 1</th><th>Blue 2</th><th>Blue 3</th><th>Red Score</th><th>Blue Score</th></tr>");
         var good_matches = 0;
         var last_match_type = null;
-        for(var i=0; i<matches.length; i++){
+        for(var i=0; i<matches.length; i++) {
             var match = matches[i];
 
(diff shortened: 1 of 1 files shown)


dracco1993:
@SUPERCILEX Is @phil-lopreiato's comment still relevant?

> Actually, time shouldn't be a mandatory field here. Just check for 'Red Score' and 'Blue Score'

SUPERCILEX:
@dracco1993 I ignored that because the code has been fine like that for over a year and I don't want to break stuff. @phil-lopreiato do you still want to move forward with that change?

phil-lopreiato:
Ummm... I don't remember. I guess this depends on a few things...

 1. Can you reproduce the bug outlined in https://github.com/the-blue-alliance/the-blue-alliance/issues/1223? You can use the FMS report attached to test. Note that the case outlined is when a future competition level isn't fully determined, so there are missing team numbers and scores (so you'll have to remove parts of this report).
[NYSU2017_MatchResultsReport_Playoffs.xlsx](https://github.com/the-blue-alliance/the-blue-alliance/files/984927/NYSU2017_MatchResultsReport_Playoffs.xlsx)

  2. If the bug still exists, can you verify that this patch actually resolves them?

SUPERCILEX:
@phil-lopreiato I have an AP test coming up, but yeah I'll make sure the bug is still around soon™.

SUPERCILEX:
Google I/O just happened so I don't see myself having much time to play around with this. Dunno if you want to close the other one too?

PR #1951, opened 2017-07-07
Standardize test format

andrewda:
This makes the test format more standard (and fixes a bunch of styling inconsistencies). Tests methods use snake_case, while non-test methods use camelCase. Also, `setUp()` and `tearDown()` are moved to the top of the test class.

Let me know if there's anything you'd like me to add/change or if I missed anything!


andrewda:
Tests passing so I'm un-WIPing this.

andrewda:
That last commit is bound to break something with ~1000 additions and ~1000 deletions...

But it's so beautiful now 😄

andrewda:
I've never been so happy that a test passed in my life. 😃

phil-lopreiato:
Holy giant diff batman! Tests pass, so I guess it's good...

PR #1953, opened 2017-07-07
Fix iffy location tests

andrewda:
Sometimes the location tests are iffy and make the entire test fail.


phil-lopreiato:
Ooh, this is a good idea to try and fix the flakiness 👍

PR #1960, opened 2017-07-16
Add FIRST Global link to Upcoming Events block

jaredhasenklein:
Since this isn't really an event, I'm not sure what the best way to do this is but I thought given that the event is happening right now, this could be a quick and easy fix.


phil-lopreiato:
Done in a cleaner way in 8393deb4c936aa4f2ba2232db71af4ca5cf15ae3

PR #1975, opened 2017-08-24
Add short url routes

andrewda:
This begins to work on some of the mentioned features in #1933.

Examples:
`thebluealliance.com/254` => `thebluealliance.com/team/254`
`thebluealliance.com/2521` => `thebluealliance.com/team/2521`
`thebluealliance.com/frc254` => `thebluealliance.com/team/254`
`thebluealliance.com/frc2521` => `thebluealliance.com/team/2521`
`thebluealliance.com/2017nytr` => `thebluealliance.com/event/2017nytr`
`thebluealliance.com/2017mndu2` => `thebluealliance.com/event/2017mndu2`

### TODO:

- [x] ~~Figure out how to cache (any ideas, @fangeugene? does redirect like this need caching?)~~
- [X] Allow `/frc####` => `/team/####`
- [x] Add tests

### DIFF
--- a/main.py
+++ b/main.py
@@ -33,7 +33,7 @@ from controllers.nearby_controller import NearbyController
 from controllers.nightbot_controller import NightbotTeamNextmatchHandler, NightbotTeamStatuskHandler
 from controllers.notification_controller import UserNotificationBroadcast
 from controllers.district_controller import DistrictDetail
-from controllers.short_controller import ShortTeamHandler, ShortEventHandler
+from controllers.short_controller import ShortEventHandler, ShortTeamHandler
 from controllers.suggestions.suggest_apiwrite_controller import SuggestApiWriteController
 from controllers.suggestions.suggest_apiwrite_review_controller import \
       SuggestApiWriteReviewController
--- a/main.py
+++ b/main.py
@@ -176,8 +176,8 @@ app = webapp2.WSGIApplication([
       RedirectRoute(r'/_/typeahead/<search_key>', TypeaheadHandler, 'ajax-typeahead', strict_slash=True),
       RedirectRoute(r'/_/webcast/<event_key>/<webcast_number>', WebcastHandler, 'ajax-webcast', strict_slash=True),
       RedirectRoute(r'/_/yt/playlist/videos', YouTubePlaylistHandler, 'ajex-yt-playlist', strict_slash=True),
-      RedirectRoute(r'/<:(frc)?><team_number:[0-9]{1,5}>', ShortTeamHandler, 'short-team-canonical'),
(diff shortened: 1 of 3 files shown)


fangeugene:
I don't think caching is fundamentally possible here.

andrewda:
Alright, guess that makes it easy.

andrewda:
Alright, should be good now 👍

main.py
```diff
@@ -175,6 +176,8 @@ def __call__(self, request, response, exception):
       RedirectRoute(r'/_/typeahead/<search_key>', TypeaheadHandler, 'ajax-typeahead', strict_slash=True),
       RedirectRoute(r'/_/webcast/<event_key>/<webcast_number>', WebcastHandler, 'ajax-webcast', strict_slash=True),
       RedirectRoute(r'/_/yt/playlist/videos', YouTubePlaylistHandler, 'ajex-yt-playlist', strict_slash=True),
+      RedirectRoute(r'/<:(frc)?><team_number:[0-9]{1,5}>', ShortTeamHandler, 'short-team-canonical'),
+      RedirectRoute(r'/<event_key:[0-9]{4}[a-z]{1,6}[0-9]?>', ShortEventHandler, 'short-event-detail'),
```
phil-lopreiato: What was your rationale for enforcing a length on event codes? _Technically_ they can be unlimited in length (like how FIRST uses the entire division name at champs), it's just TBA convention to keep them short.

fangeugene: There shouldn't be a length on event codes. we make no guarantees for that.

andrewda: Hmm, what regex do you guys suggest here? I was thinking something like

```python
[0-9]{4}[a-z0-9]+
```

But that would trigger any value that has 4 digits followed by any amount of text. Maybe that's fine, I'm not really sure.

phil-lopreiato: Here's how we validate event keys in the code: https://github.com/the-blue-alliance/the-blue-alliance/blob/master/models/event.py#L423

andrewda: Another idea - maybe keep these regexs pretty loose (such as what I wrote above) and only redirect if the Team/Event actually exists. Otherwise show a 404.

dracco1993: Yeah, I think keeping it loose then 404'ing if the team/event/etc. is invalid isn't a bad idea.

fangeugene: Well redirecting to an invalid page will 404 automatically, so I don't think it's necessary to do that logic here.

dracco1993: That's what I was saying.  Just let the router send people wherever, then let the normal error handling mechanisms do their thing.  👍

andrewda:
Alright, that look good?

PR #1977, opened 2017-08-26
Lint Python Code in Pulll Requests

phil-lopreiato:
This lints the entirety of changed files in a given pull. This isn't ideal, but I'm not entirely sure how to extract only the changed lines. Anyone with some good git-fu?


andrewda:
I think this is probably what we want (linting all changed files). If you delete the only reference to a variable, for example, you'd expect the linter to yell at you for an unused variable, but it wouldn't if you only lint the changed lines.

phil-lopreiato:
True, maybe this is alright then.

bdaroz:
I tossed something to this regard in slack b/c I wasn't at my desk, but yes. It'll also generally lead to false positives and negatives. If you want to quality gate before you push to production, two checks are worthwhile: 
1. does the overall issue count rise between deployed builds, 2. Are there issues/new issues in the changed lines.
Either one would be worth aborting automatic deployment.

phil-lopreiato:
I'm actually only running lint on pull requests for now. But there is likely a way to find python syntax errors (as opposed to style errors) and run those always

PR #1980, opened 2017-08-26
Clean Up Travis Builds

phil-lopreiato:


ops/travis/travis-install.sh
```diff
@@ -0,0 +1,15 @@
...
+    exit 0
+fi
+
+pip install -r travis_requirements.txt
+paver install_libs
+wget https://storage.googleapis.com/appengine-sdks/featured/google_appengine_1.9.40.zip -nv
+unzip -q google_appengine_1.9.40.zip -d $HOME
+rm google_appengine_1.9.40.zip
+npm install
+npm install -g gulp-cli uglify-js uglifycss less
```
bdaroz: Q: If you do the `npm install -g` before the `npm-install` won't `npm` not have to install those packages twice?

phil-lopreiato: `npm install` will put them in the local `node_modules` directory while the latter will install them somewhere else. So i don't think the order makes a difference

PR #1988, opened 2017-08-28
FIRST HoF page

fangeugene:
image


fangeugene:
The url is `/first-hof`. Any comments?

I'm also linking to the page from the main Insights page.

andrewda:
> The url is `/first-hof`. Any comments?

Why not just `/hall-of-fame`, and redirect there from `/hof`? I think "first" is unnecessary. Maybe also a little message describing how teams get in the hall of fame?

Issue #1390, opened 2016-03-04
Delete bad team media


fangeugene:
@gregmarra or I should go through the contact mail and look for delete media requests now that #1370 is done

PR #1989, opened 2017-08-28
Remove 404ing team media

andrewda:
Should help out with #1390. If media is 404ing, remove it.


andrewda:
Should I also disallow users from submitting 404ing media?

requirements.txt
```diff
@@ -3,4 +3,5 @@
 jinja2==2.6
 numpy==1.6.1
 Paver==1.2.4
-pep8==1.7.0
\ No newline at end of file
+pep8==1.7.0
+httplib2==0.10.3
```
fangeugene: Can we use one of the existing http request libraries?

andrewda: What are we using at the moment?

fangeugene: https://github.com/the-blue-alliance/the-blue-alliance/blob/master/datafeeds/datafeed_fms_api.py#L186

The `context` thing is a bit funny, but it allows us to make async calls.

controllers/cron_controller.py
```diff
@@ -666,6 +689,29 @@ def get(self):
             logging.exception(e)
 
 
+class CheckMediaDo(webapp.RequestHandler):
+    """
+    Remove unavailable media
+    """
+    def get(self, media_key):
```
fangeugene: Everything between `.get()` and `.put()` should be in a transaction (https://cloud.google.com/appengine/docs/standard/python/ndb/transactions). Right now, this runs the risk of losing data (although the risk is really small). If between `.get()` and `.put()`, another process edits the Media object (like an admin adding a tag), `.put()` will overwrite the update.

controllers/cron_controller.py
```diff
@@ -666,6 +689,29 @@ def get(self):
...
+        elif media.media_type_enum == MediaType.YOUTUBE_VIDEO:
+            url = "https://www.youtube.com/oembed?url={}&format=json".format(media.youtube_url_link)
+        elif media.media_type_enum == MediaType.EXTERNAL_LINK:
+            url = media.external_link
+
+        if not WebsiteHelper.exists(url):
+            media.key.delete()
+        else:
+            media.checked = datetime.datetime.now()
+            media.put()
```
fangeugene: Add the following comment here:

`# .put() is being used instead of MediaManipulator to avoid cache busting.`

helpers/website_helper.py
```diff
@@ -18,3 +21,17 @@ def format_url(cls, website_url):
...
+        Verify that a given URL exists (returns a non-404 status code)
+        """
+        context = ndb.get_context()
+
+        if not website_url:
+            return False
+
+        result = context.urlfetch(website_url).get_result()
+
+        return int(result.status_code) != 404
```
fangeugene: Is the status code not already an int?

controllers/cron_controller.py
```diff
@@ -666,6 +689,29 @@ def get(self):
...
+        media = ndb.Key(Media, media_key).get()
+
+        if media.is_image:
+            url = media.view_image_url
+        elif media.is_social:
+            url = media.social_profile_url
+        elif media.media_type_enum == MediaType.YOUTUBE_VIDEO:
+            url = "https://www.youtube.com/oembed?url={}&format=json".format(media.youtube_url_link)
+        elif media.media_type_enum == MediaType.EXTERNAL_LINK:
+            url = media.external_link
```
fangeugene: This needs an `else: return` case. If we introduce new media types that don't get caught by this we don't want them getting accidentally deleted.

controllers/cron_controller.py
```diff
@@ -666,6 +689,29 @@ def get(self):
...
+
+        if not WebsiteHelper.exists(url):
+            media.key.delete()
```
fangeugene: This should be `MediaManipulator.delete(media)` so caches get busted.

helpers/website_helper.py
andrewda: Good point, I'm guessing you're right. For some reason I thought it was a string, but I'll check.

andrewda:
@fangeugene combing through some old PRs and wondering if you think anything still needs to be updated here. I'm gonna try to pull in the past year of changes and make sure everything's still working.

PR #2017, opened 2017-11-03
[vagrant bootstrap] Fix bug preventing provisioning multiple times

MC42:
The first time this runs, the template gets renamed. So it fails the second time


phil-lopreiato:
This really should only `cp` if the target file doesn't exist (I think there's a flag for that), but don't worry about that for now, I'll merge this to unblock you

PR #2024, opened 2017-11-13
[Trusted API] Support surrogates/dqs

fangeugene:
Backwards compatible; `surrogates` and `dqs` are optional.


fangeugene:
Already did (see above)

phil-lopreiato:
Reading is hard 😛

PR #2034, opened 2017-11-23
Remove mixed content warning in GameDay

technonerdz:
Load video.js over https to remove the mixed content warning on the top right in chrome and prevent it from getting blocked.


phil-lopreiato:
Thanks @technonerdz!

PR #2036, opened 2017-11-23
Use minified video.js scripts

### LINKED ISSUE #2035: [GameDay] Use minified video.js scripts
(see [PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fwww.thebluealliance.com%2Fgameday))

Using the minified video.js scripts will save over 100kB on the request. Minified versions are already on the Cloudflare CDN, so just using `.min.js` instead of `.js` will be enough.

andrewda:
## Description
Uses minified video.js scripts instead of large original scripts. Fixes #2035.

## Motivation and Context
Large reduction in GameDay page size (~133 kB).

## How Has This Been Tested?
Tested locally.

## Types of changes

- [ ] Breaking change (fix or feature that would chance API specifications or require data migrations)

### DIFF
--- a/templates/gameday2.html
+++ b/templates/gameday2.html
@@ -36,8 +36,8 @@
 {% block login_modal %}{% endblock %}
 
 {% block main_javascript %}
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/video.js/5.10.2/video.js"></script>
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
Nice catch, @andrewda

PR #2040, opened 2017-11-29
Update api_v3.json to match Swagger spec

dc74089:
Update the `api_v3.json` file to comply with Swagger spec and match data actually returned.
## Description

In the `api_v3.json` file:

- Update the lines that reference `WLT_Record` in the `Team_Event_Status_playoff` object to the correct syntax.
- Move the `required` block from the root of `Event_District_Points` into the inner `additionalProperties` object to match Swagger spec and correct the formatting on the `/apidocs/v3` webpage.
- Add descriptions to the children of `Event_District_Points` as required by Swagger spec. 
- Change the `Match.video` object to fix incorrect syntax and more correctly reflect what is currently returned by the API. 

## Motivation and Context

Fixes syntax errors that prevent auto-generating an API client using swagger-codegen, and fixes some formatting on the `/apidocs/v3` webpage. 
## How Has This Been Tested?

Validated syntax using the official [Swagger Editor](https://swagger.io/swagger-editor/), and confirmed that the new `api_v3.json` file matches what it returned by the API in its current state. 
If appropriate
## Types of changes


phil-lopreiato:
Thanks for the fix! We should add a linter (like this? https://github.com/whitlockjc/swagger-lint) for the swagger files into our CI rig to make sure the files are always valid

PR #2051, opened 2018-01-02
Don't track API calls with errors. Fixes #2050

### LINKED ISSUE #2050: Don't track API calls that have 4XX errors in GAnalytics
Throws off actual usage stats and maybe inflates APIv2 usage stats.

fangeugene:
## Motivation and Context
Throws off actual usage stats and maybe inflates APIv2 usage stats compared to APIv3.

Note: API statistics may decrease for 2018 vs. 2017 because we won't be logging "bad" calls anymore.

## How Has This Been Tested?
Tried valid and invalid API calls (v2 and v3) and verified that only valid calls called the tracking function.

## Types of changes

### DIFF
--- a/controllers/api/api_base_controller.py
+++ b/controllers/api/api_base_controller.py
@@ -84,10 +84,12 @@ class ApiBaseController(CacheableHandler):
         if self._errors:
             self.abort(400)
 
-        self._track_call(*args, **kw)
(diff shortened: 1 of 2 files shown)


phil-lopreiato:
Yeah.... This is definitely something we should be doing

PR #2052, opened 2018-01-02
Update build season landing page for 2018

fangeugene:
Still needs:

- [x] Updated game animation
- [x] Updated encrypted manual password


gregmarra:
These could be sitevars for 2019 if we wanted to be fancy. Shouldn't require a code push to update.

phil-lopreiato:
I think that's a good idea

phil-lopreiato:
@gregmarra https://github.com/the-blue-alliance/the-blue-alliance/pull/2055 :)

PR #2055, opened 2018-01-06
Parameterize Landing Pages

phil-lopreiato:
## Description
Do some hacky stuff around the main landing controller to make everything controlled by a sitevar

## Motivation and Context
Change the landing page without needing to redeploy the site

## How Has This Been Tested?
Run locally

## Types of changes


controllers/main_controller.py
```diff
@@ -220,6 +223,40 @@ def _render(self, *args, **kw):
         return template.render(path, self.template_values)
 
 
+class MainLandingHandler(CacheableHandler):
```
phil-lopreiato: This is where the magic happens. It's horribly hacky, but gets the job done

PR #2070, opened 2018-02-06
Opens search results in same tab

jaredhasenklein:
On the Search page, results open in a new tab
## Description

Added the `linktarget="_parent"` attribute per the documentation from [devs.google](https://developers.google.com/custom-search/docs/element)

### DIFF
--- a/templates/search.html
+++ b/templates/search.html
@@ -20,7 +20,7 @@
           s.parentNode.insertBefore(gcse, s);
         })();
       </script>
-      <gcse:search></gcse:search>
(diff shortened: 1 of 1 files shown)


fangeugene:
Why `_parent` and not `_self`?

jaredhasenklein:
I found a few Stack Exchange and other things saying that `_parent` is the way to do it - I'm not sure if there's any practical difference here.

jaredhasenklein:
I found another few saying `_self` is actually superior - I'll adjust

PR #2088, opened 2018-02-12
Update team badges to use v3 API and other cleanup

legoktm:
My goal was to stay as close to the original TBA team badges as
possible. Because there is no single equivalent v3 API call, this now
needs to make multiple API calls and is twice as slow due to that
(~1sec).

You will need to create a `config.php` file with the X-TBA-Auth-Key for
making API requests.

The generated teamoverlays and gradients directories are now commited to
git to keep the local tree clean when testing.

## Motivation and Context
The TBA badges that ChiefDelphi users love stopped working. I saw someone mention it it in the chat, realized it was PHP, and decided to fix it.

Sadly, I couldn't find an open issue related to this.

## How Has This Been Tested?

I have PHP 7.1 and php-gd installed locally (Fedora 27). Inside the tbadotnet directory, I ran `php -S 127.0.0.1:8002` to start the built-in development server. Open <http://127.0.0.1:8002/teambadge_gen.php?team=604&start=000000&stop=C0C0C0&debug=1&regen=1> in your browser to generate the image. I tested with teams 604, 254, 973, 8, and 2767. (Though teams with more than 5 events like 2767 works in "best effort" mode).

Theoretically this should work with PHP 5.6 and later, but I think functionality will mostly depend on what version of the gd library is used.

604_tba_2017
604_tba_2018

## Types of changes

### DIFF
--- a/utils/tbadotnet/.gitignore
+++ b/utils/tbadotnet/.gitignore
@@ -1,0 +1,3 @@
+/config.php
+/gradients/from*
+/teamoverlays/team*
--- a/utils/tbadotnet/config.php.example
+++ b/utils/tbadotnet/config.php.example
@@ -1,0 +1,5 @@
+<?php
+
+// X-TBA-Auth-Key
+$key = '...';
+
--- a/utils/tbadotnet/tba_gd_lib.php
+++ b/utils/tbadotnet/tba_gd_lib.php
@@ -1,4 +1,4 @@
-<?
+<?php
 
 /**
  * @param    $hex string        6-digit hexadecimal color
@@ -128,4 +128,12 @@ function imageStringoutline(&$im,$size,$x,$y,&$col,&$outlinecol,$text,$width) {
     // Draw the main text
     $text2 = imageString($im,$size,$x,$y,$text,$col);
 }
-?>
+
+function tba_api_request($path) {
+	global $key;
+	// TODO: Use curl instead?
+	$json = file_get_contents(
+		"https://www.thebluealliance.com/api/v3/$path?X-TBA-Auth-Key=$key"
+	);
+	return json_decode($json, true);
+}
--- a/utils/tbadotnet/teambadge_gen.php
+++ b/utils/tbadotnet/teambadge_gen.php
@@ -1,4 +1,6 @@
-<?PHP
+<?php
+
+include __DIR__ . '/config.php';
 
 include("tba_gd_lib.php");
 error_reporting(E_ALL ^ E_NOTICE ^ E_STRICT);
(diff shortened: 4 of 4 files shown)


fangeugene:
This is awesome. Some questions for @gregmarra and @phil-lopreiato:
1) Do we still have the capability to update whatever server is hosting the old PHP code? http://thebluealliance.net/tbatv/teambadge_gen.php?team=604&start=000000&stop=C0C0C0 is still clearly doing something...
2) Even if we can update the PHP server, do we want to? Would it make more sense to re-implement this badge on App Engine or as a JS/iframe widget? ChiefDelphi probably doesn't allow the embedding of JS/iframe widgets, so serving images via App Engine might be the way to go.
3) Is there enough demand for this? Serving images for CD would be a lot of bandwidth. The JS/iframe widget for team websites should be cheap.

gregmarra:
I likely can update the old server. I would prefer to migrate away from it
and generate these in app engine.

Images are easy to cache, I don’t think we need to worry about bandwidth.

legoktm:
GAE supports PHP (<https://cloud.google.com/appengine/docs/php/>) but I think in the long term Python is probably better for maintenance...maybe with Pillow or something.

That said, I think it would be cool to see these come back for this season to fit in with the 8-bit retro theme, so maybe updating the old server would be OK for now? :-)

gregmarra:
I’ll take a look later this week at php on the old server.

This is a simple enough thing that we should port to Python, not host PHP
on GAE.

fangeugene:
Hosting PHP on GAE would be a headache because it would need to be a separate environment from our Python one.

If we do host it on the old server, define a cleaner URL pattern that will be future proof and set up redirects with the .com domain.

gregmarra:
Thinking out loud:

tba.com/team/<number>/banner/<gradient>/<gradient>

fangeugene:
Or if grouping them together makes more sense: `/teambanner/<number>/<gradient>/<gradient>`
Also "badge" vs "banner"

legoktm:
For the record, URLs like <http://thebluealliance.net/tbatv/teambadge/gen/0177/ccc/ffcc00/teambadge.png> currently redirect.

gregmarra:
It appears that the master account to the tbadotnet server no longer has root access. I'm not able to modify files either using cPanel's file manager tool or after ssh'ing into the server. I spent about 30 minutes trying to debug this, but wasn't successful.

I think migrating to Python on App Engine to generate these is the best long term strategy.

jaredhasenklein:
This seems like it should be closed given that Chief signatures are not a thing anymore. Unless anyone know some alternative applications. 

As an aside, the links that were working as of February 2018 no longer work for anyone wondering.

PR #2090, opened 2018-02-13
2018 Real-Time Scoring on Event Pages (WIP)

fangeugene:
This PR replaces the "Previous/Upcoming matches" at the top of an event with a live-updating widget that shows previous match results, upcoming match times, and ongoing match data from GDCV.

## Description
A the widget is a React component that listens to Firebase and displays previous match results, upcoming match times, and ongoing match data.

## Motivation and Context
We (will) have GDCV! This makes use of it in an interesting way.

This provides a way to follow matches live without watching the webcast. Potential use cases:
- Want to follow event on mobile with limited bandwidth
- Twitch blocked by a firewall
- In-pit screen at competition

## How Has This Been Tested?
Works using dev TBA Firebase with emulated data 

Video (outdated): https://www.youtube.com/watch?v=brMztvJjItk

image

## Types of changes


dracco1993:
It's beautiful.   😲

templates_jinja2/event_details.html
```diff
@@ -106,19 +106,21 @@ <h3 class="panel-title">Event in progress</h3>
...
-        <div class="col-sm-6">
-          <h4>Upcoming Matches</h4>
-          {{ mtm.upcoming_match_table(matches_upcoming) }}
-        </div>
-        {% endif %}
-        {% if matches_recent %}
-        <div class="col-sm-6">
-          <h4>Recent Matches</h4>
-          {{ mtm.recent_match_table(matches_recent) }}
+        <div id="liveevent-content">
```
dracco1993: What're your thoughts on having a graceful degradation back to the old view if firebase doesn't have the necessary data?  Or is that not a concern?  🤔

fangeugene: If there is no javascript, the old view won't get replaced. If there is, it will get replaced. Firebase will have the necessary data to render last/upcoming matches, so it will look like the old view but live updating.

dracco1993: Fair enough.  👍

fangeugene:
Going to merge this once tests pass to ensure people get the updated page for Week 0.

PR #2142, opened 2018-03-01
[WIP] Add advanced match search

andrewda:
## Description

This adds a new Advanced Match Search page, based off the Advanced Team Search. It currently allows users to search for matches based on year and teams on one alliance and/or the other. More search queries to come.

- [ ] Search by event
- [ ] Search by match level (quals, quarters, semi, finals)
- [ ] Search by score?
- [ ] Search by DQ?

@fangeugene @phil-lopreiato any other search query ideas?

## Motivation and Context

I think something along these lines was asked for on Slack, and I thought it was a good idea.

## How Has This Been Tested?

Not much so far, hence WIP.

screen shot 2018-03-01 at 12 02 03 am
screen shot 2018-03-01 at 12 02 14 am
screen shot 2018-03-01 at 12 02 32 am
screen shot 2018-03-01 at 12 03 09 am

## Types of changes


fangeugene:
Perhaps take a look at Computed Properties instead of building a separate search index?  Might be easier to not have to worry about keeping everything in sync.

https://cloud.google.com/appengine/docs/standard/python/ndb/entity-property-reference#computed

andrewda:
Yea I tried that but wasn't able to search based on the computer values (though I might have been doing something wrong, will look again in a bit).

fangeugene:
I think the whole point of computed properties is that they're indexed.
Otherwise why not just make them instance methods/properties like we
already have.

andrewda:
screen shot 2018-03-01 at 9 52 13 pm

screen shot 2018-03-01 at 9 52 49 pm

@fangeugene anything else you'd like to see here/have changed?

fangeugene:
Will take a close look this weekend.

controllers/advanced_search_controller.py
```diff
@@ -204,4 +204,208 @@ def _render(self):
...
-        return jinja2_engine.render('advanced_search.html', self.template_values)
+        return jinja2_engine.render('advanced_team_search.html', self.template_values)
+
+
+class AdvancedMatchSearchController(CacheableHandler):
+    VALID_YEARS = list(reversed(range(1992, tba_config.MAX_YEAR + 1)))
+
+    VALID_COMP_LEVELS = [
+        ('qm', 'Qualifications'),
+        ('qf', 'Quarterfinals'),
```
fangeugene: `'ef': 'Octofinals'` is a thing.

controllers/admin/admin_cron_controller.py
```diff
@@ -534,3 +535,28 @@ class AdminUpdateTeamSearchIndexDo(LoggedInHandler):
...
+        for match_key in match_keys:
+            taskqueue.add(
+                queue_name='search-index-update',
+                url='/tasks/do/update_match_search_index/' + match_key.id(),
+                method='GET')
+
+
+class AdminUpdateMatchSearchIndexDo(LoggedInHandler):
+    def get(self, match_key):
+        team = Match.get_by_id(match_key)
```
fangeugene: Should this be `match` instead of `team`?

helpers/search_helper.py
```diff
@@ -55,6 +56,31 @@ def update_team_location_index(cls, team):
     def remove_team_location_index(cls, team):
         search.Index(name=cls.TEAM_LOCATION_INDEX).delete(team.key.id())
 
+    @classmethod
+    def update_match_number_index(cls, match):
```
fangeugene: A task that enqueues a call to this should be added to `MatchManipulator.postUpdateHook()` to keep the search index in sync with this. Add to the queue `search-index-update`.

controllers/admin/admin_cron_controller.py
```diff
@@ -534,3 +535,28 @@ class AdminUpdateTeamSearchIndexDo(LoggedInHandler):
     def get(self, team_key):
         team = Team.get_by_id(team_key)
         SearchHelper.update_team_awards_index(team)
+
+
+class AdminUpdateAllMatchSearchIndexEnqueue(LoggedInHandler):
+    def get(self):
+        taskqueue.add(
+            queue_name='match-search-index-update',
```
fangeugene: Should be `search-index-update`

helpers/search_helper.py
```diff
@@ -55,6 +56,31 @@ def update_team_location_index(cls, team):
...
+                search.TextField(name='team3', value=match.team3),
+                search.TextField(name='team4', value=match.team4),
+                search.TextField(name='team5', value=match.team5),
+                search.TextField(name='team6', value=match.team6),
+            ]
+            search.Index(name="matchNumber").put(
+                search.Document(doc_id=match.key.id(), fields=fields))
+
+    @classmethod
+    def remove_match_number_index(cls, match):
```
fangeugene: Similar to above, a task that enqueues a call to this should be added to MatchManipulator.postDeleteHook() to keep the search index in sync with this. Add to the queue search-index-update.

models/match.py
```diff
@@ -95,6 +95,17 @@ class Match(ndb.Model):
     push_sent = ndb.BooleanProperty()  # has an upcoming match notification been sent for this match? None counts as False
     tiebreak_match_key = ndb.KeyProperty(kind='Match')  # Points to a match that was played to tiebreak this one
 
+    event_key = ndb.ComputedProperty(lambda self: self.event_key_name)
```
fangeugene: These all shouldn't need to be ComputedProperties since you're building the separate search index. `@property` should work fine.

PR #2143, opened 2018-03-01
Changed 2017 GameDay OG image to 2018 season image.

MC42:
## Description
I changed the GameDay meta image to be the 2018 rather than the 2017 one.

## Motivation and Context
This change is a visual style change to ensure consistency of year.

## How Has This Been Tested?
The only change is to the file, and given identical name and encoding, there's no difference for the OpenGraph tag.

## Types of changes


fangeugene:
Can you find one with more robots/action on the field?

PR #2146, opened 2018-03-01
Fix twitch chat to not prompt on first message

fangeugene:
## Motivation and Context
Old embed had a weird alert when a user tries to send their first message.

## Types of changes


fangeugene:
Actually, this is a non-issue that only affects mods.

PR #2171, opened 2018-03-04
Fix "Currently competing at" panel on team page

dracco1993:
## Description
I changed the amount of screen each section had to work with, taking 1/12 from the left-hand side container and giving it to the right-hand side container.

## Motivation and Context
There's plenty of screen reality to work with, so nothing should be overflowing.  😁 

## How Has This Been Tested?
Locally only.  See screenshots blow.  👇 

Before:
Before

After:
After

## Types of changes

### DIFF
--- a/templates_jinja2/team_partials/team_info.html
+++ b/templates_jinja2/team_partials/team_info.html
@@ -13,7 +13,7 @@
 </div>
 
 <div class="row">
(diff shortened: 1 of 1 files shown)


dracco1993:
Thanks @fangeugene!

PR #2191, opened 2018-03-09
Fixed an unmet dependency error that caused provisioning to fail

tweirtx:
## Description
Added an apt dependency that was causing dev environment setup to fail.

## Motivation and Context
This change is required to set up a new development environment correctly. It solves a mysql_config environment error on first-time setup.

## How Has This Been Tested?
Issue was discovered trying to set up a development environment, applied fixes, reran setup commands and did not get the error. This was on an Ubuntu 16.04 system.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/ops/dev/Dockerfile
+++ b/ops/dev/Dockerfile
@@ -20,7 +20,8 @@ RUN apt-get update && apt-get install -y \
   libssl-dev \
   tmux \
   vim \
-  openssh-server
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
Thanks for finding the fix here!

PR #2195, opened 2018-03-09
Fix model of /team/{team_key}/robots to match API return

rakusan2:
I have:
* Fixed the model of /team/{team_key}/robots to match API return
* Added min and max to year parameter
* Added min to page_num parameter
* Changed { "type": "number", "format": "integer" } to { "type": "integer" }

## Description
I changed the endpoint to better show what is returned
The others are for autogenerated clients and for the online API test to not accept invalid values 

## Motivation and Context
My own autogenerated [client](https://github.com/rakusan2/tba-client-generator)

## How Has This Been Tested?
I have tested it on the online Swagger Editor

## Types of changes

### DIFF
--- a/static/swagger/api_v3.json
+++ b/static/swagger/api_v3.json
@@ -2,10 +2,10 @@
   "swagger": "2.0",
   "info": {
     "description": "# Overview \n\n Information and statistics about FIRST Robotics Competition teams and events. \n\n# Authentication \n\nAll endpoints require an Auth Key to be passed in the header `X-TBA-Auth-Key`. If you do not have an auth key yet, you can obtain one from your [Account Page](/account). \n\n A `User-Agent` header may need to be set to prevent a 403 Unauthorized error.",
-    "version": "3.03.0",
+    "version": "3.03.1",
     "title": "The Blue Alliance API v3",
     "x-version-info": "Versions of the API follow the format X.Y.Z, with X being a major version, Y denoting the minor version, and Z the point release. Changes to the spec or API that result in a major version change will significantly impact implementations. Changes to the minor version indicate paths and/or fields in existing models may be added, and any breaking changes will be confined to paths or models listed as in-development in the prior minor version. Changes to the point release indicate no format or structure changes to the paths or models, but updated or clarified documentation. Note that changes to paths and models refer to the actual result from the API, not the documentation.",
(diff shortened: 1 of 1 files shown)


static/swagger/api_v3.json
```diff
@@ -62,16 +62,17 @@
...
     },
     "year": {
       "name": "year",
       "in": "path",
       "description": "Competition Year (or Season). Must be 4 digits.",
       "required": true,
-      "type": "number",
-      "format": "integer"
+      "type": "integer",
+      "minimum":1997,
```
andrewda: space before value (and below)

static/swagger/api_v3.json
```diff
@@ -62,16 +62,17 @@
...
+      "maximum":2020
```
andrewda: Not sure I like hard-coding the value of something like this that a) is not the current year and b) we might forget to update in the future.

If it's the current year, fine, we'll update it with everything else that needs to be updated year-to-year, but setting it to several years in the future puts me off.

static/swagger/api_v3.json
```diff
@@ -62,16 +62,17 @@
       "in": "path",
       "description": "Page number of results to return, zero-indexed",
       "required": true,
-      "type": "number",
-      "format": "integer"
+      "type": "integer",
+      "minimum":0
```
andrewda: space before value

static/swagger/api_v3.json
rakusan2: I can change the max to 9999 just to ensure that it will always be a 4 digit number

static/swagger/api_v3.json
```diff
@@ -62,16 +62,17 @@
...
     "year": {
       "name": "year",
       "in": "path",
       "description": "Competition Year (or Season). Must be 4 digits.",
       "required": true,
-      "type": "number",
-      "format": "integer"
+      "type": "integer",
+      "minimum": 1997,
+      "maximum": 9999
```
phil-lopreiato: Hmmm, I guess we do assume this is a 4 digit number in a lot of places, but this still feels clunky

static/swagger/api_v3.json
```diff
@@ -62,16 +62,17 @@
...
     },
     "year": {
       "name": "year",
       "in": "path",
       "description": "Competition Year (or Season). Must be 4 digits.",
       "required": true,
-      "type": "number",
-      "format": "integer"
+      "type": "integer",
+      "minimum": 1997,
```
phil-lopreiato: We go back to 1992: https://www.thebluealliance.com/events/1992

static/swagger/api_v3.json
rakusan2: It only stops people from inputting a 5 digit number on the site and has zero effect on the actual request.
Also since the site is not showing the minimum I can also lower the minimum to 1000 in order for it to show a red border on an incorrect number of digits only if you want

phil-lopreiato: I don't think the UI being accurate is all that important of a concern, if people put in an invalid year it'll simply return no events. I think it'll be fine to just remove the minimum and maximum values here

rakusan2:
I have removed the min and max

PR #2225, opened 2018-04-02
Minor cleanup to team detail pages.

gregmarra:
## Description
Standardize references to teams as "Team 1234" on team details page. IMO, we shouldn't use #'s to refer to teams, and we should capitalize "Team" when referring to a specific team. If there is any official style guide from FIRST I would love to follow that!

## Motivation and Context
Consistency.

## How Has This Been Tested?
Tested on local dev machine

## Types of changes
- [ x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/templates_jinja2/team_details.html
+++ b/templates_jinja2/team_details.html
@@ -62,14 +62,14 @@
           <hr>
           <h2 id="event-results">Event Results</h2>
           {% if season_wlt %}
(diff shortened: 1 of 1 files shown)


pordonj:
IIRC There was one year where FIRST's thing was referring to your team in a specific way (might have been part of Dean's Homework/Make It Loud one year). If I also remember correctly it was that you were supposed to always say "FIRST Team xyz" - but that probably makes more sense when talking/presenting info to people "outside the fold".

PR #2237, opened 2018-04-08
Show event names in match video suggestion review

gregmarra:
## Description
Adds event names in header of match video suggestion review.

## Motivation and Context
@bdaroz asked for it to make approving videos easier.

## Failure Case
If the start of the `suggestion.target_key` is not an Event, creating the Event object will fail and the page will 500. There is no way to create a malformed suggestion other than writing a bad request, but if a malicious suggestion were created, it would break the page. I have not added error handling for this case.

## How Has This Been Tested?
Tested on local machine.

image

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)


gregmarra:
Sorry for the extra commit I had to revert.

PR #2269, opened 2018-04-29
Adds property safety to event/{event_key}/insights endpoint

ZachOrr:
Modifies the Event Insights endpoint (`event/{event_key}/insights`) to add a year-independent model to represent the `qual` and `playoff` fields for event insights. This should be a non-breaking change for API consumers

## Motivation and Context

If we're going to have Event Insights models for every year, we don't need to have models that duplicate functionality (ex: `Event_Insights_2017` and `Event_Insights_2018` used to be the same model)

## How Has This Been Tested?

Only in the Swagger editor and it *looks* right - let me know if there's code changes to be made

screen shot 2018-04-29 at 10 48 36 am

## Types of changes

### DIFF
--- a/static/swagger/api_v3.json
+++ b/static/swagger/api_v3.json
@@ -2122,8 +2122,7 @@
           "200": {
             "description": "Successful response",
             "schema": {
-              "type": "object",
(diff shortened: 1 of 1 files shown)


ZachOrr:
Just now realizing we never documented 2018 😅

PR #2272, opened 2018-04-30
Add warning to timeseries about accuracy.

MC42:
As requested in Slack, a warning blurb below the timeseries chart as to the accuracy of the data following multiple Chief Delphi threads about it.

## Motivation and Context
Teams were confusing the time series chart as a definitive source of data and minor bugs in the OCR were causing 200 point jumps in either direction (OCR'd a 1 as a 3).  This hopefully will clarify that it's an experimental feaure.

## How Has This Been Tested?
I used inspect element to check that it displayed properly and implemented it in the exact location in the jinja template creates it in.

## Types of changes

### DIFF
--- a/templates_jinja2/match_timeseries.html
+++ b/templates_jinja2/match_timeseries.html
@@ -26,7 +26,7 @@
     <div class="col-xs-12">
       {% if timeseries_data %}
       <canvas id="timeseries-chart" style="width: 500px; height: 100px;">{{timeseries_data}}</canvas>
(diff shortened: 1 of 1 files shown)


fangeugene:
Screenshot?

MC42:
image
Oops, sorry; here ya go.

bdaroz:
One thought... the graph is the problem, the Detailed Results are good. It's not exactly clear as to which the text refers to.

bdaroz:
Perhaps: `The above graph is unofficial, and is subject to signifiant error or missing data. Do not rely on the above data for any purpose.`

Issue #2277, opened 2018-05-02
Migrate to Cloud Endpoints v2


phil-lopreiato:
Cloud Endpoints v1 is deprecated.

https://cloud.google.com/endpoints/docs/frameworks/python/migrating

Tried to do in https://github.com/the-blue-alliance/the-blue-alliance/pull/2044 but reverted in https://github.com/the-blue-alliance/the-blue-alliance/pull/2045

phil-lopreiato:
Doesn't work yet

phil-lopreiato:
Found https://stackoverflow.com/questions/44228959/google-python-api-tries-to-import-the-deprecated-oauth2client-contrib-multistore

Trying in 287a479

phil-lopreiato:
Yay that worked!

PR #2278, opened 2018-05-03
Migrate to Cloud Endpoints v2 (yet again)

phil-lopreiato:
## Description
Followed migration guide:
https://cloud.google.com/endpoints/docs/frameworks/python/migrating

I think the problem last time was that we didn't deploy a new GAE version. So let's try that now

Closes https://github.com/the-blue-alliance/the-blue-alliance/issues/2277

## Motivation and Context
Google will deprecate this in August....

## How Has This Been Tested?
yolo

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -16,10 +16,10 @@ libraries:
   version: "1.5"
 - name: numpy
   version: "1.6.1"
-- name: endpoints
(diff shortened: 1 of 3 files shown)


phil-lopreiato:
Brace yourselves, here goes nothing

PR #2286, opened 2018-05-20
Removed trailing commas that broke JSON

tweirtx:
## Description

There were trailing commas in the JSON that made it invalid
## Motivation and Context

The invalid JSON broke the build system for the codegen libraries.
## How Has This Been Tested?

Linted JSON on jsonlint.com, fixed all errors it gave me. JSON is now valid.
## Types of changes

### DIFF
--- a/ops/dev/Dockerfile
+++ b/ops/dev/Dockerfile
@@ -21,8 +21,7 @@ RUN apt-get update && apt-get install -y \
   tmux \
   vim \
   openssh-server \
-  libmysqlclient-dev \
-  language-pack-en-base
+  libmysqlclient-dev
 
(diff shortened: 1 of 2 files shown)


fletch3555:
Confirmed fixed.

Before:
```
$ curl -s 'https://www.thebluealliance.com/swagger/api_v3.json' | jq -r '.info.version'
parse error: Expected another key-value pair at line 4103, column 9
```

After:
```
$ curl -s 'https://raw.githubusercontent.com/tweirtx/the-blue-alliance/json-fix/static/swagger/api_v3.json' | jq -r '.info.version'
3.03.1
```

PR #2305, opened 2018-09-15
Add variant to mobile API messages to fix IntegerField being serialized as strings

ZachOrr:
## Description
In the mobile API messages, all values being set in `IntegerField` classes were being serialized and sent to clients as strings. This adds a specific `variant` field to the `IntegerField` instances that forces them to be serialized as integers (example of this [here](https://stackoverflow.com/a/30737750/537341))

## Motivation and Context
This fixes incorrect functionality in the mobile API

## How Has This Been Tested?
Deployed my changes to a GCP and consumed the modified endpoint functionality to ensure codes were being sent back as integers, as opposed to strings

## Types of changes

### DIFF
--- a/models/mobile_api_messages.py
+++ b/models/mobile_api_messages.py
@@ -4,7 +4,7 @@ from protorpc import messages
 
 
 class BaseResponse(messages.Message):
-    code = messages.IntegerField(1, default=200)
+    code = messages.IntegerField(1, default=200, variant=messages.Variant.INT32)
     message = messages.StringField(2, default='')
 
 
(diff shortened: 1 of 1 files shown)


fangeugene:
Can we be sure this doesn't break Android?

ZachOrr:
This seems fine to me - everywhere I'm looking on Android, they're treating that `model_type` or `code` as an Int

https://github.com/the-blue-alliance/the-blue-alliance-android/blob/5c94e11e5640a08a3547c16646efec3f018835c8/libTba/src/main/java/com/appspot/tbatv_prod_hrd/model/ModelsMobileApiMessagesBaseResponse.java
https://github.com/the-blue-alliance/the-blue-alliance-android/blob/5c94e11e5640a08a3547c16646efec3f018835c8/libTba/src/main/java/com/appspot/tbatv_prod_hrd/model/ModelsMobileApiMessagesFavoriteMessage.java

cc @phil-lopreiato @nwalters512

ZachOrr:
Ah yeah - the quick way to verify this change (server-side, not Android side) is to just generate the `tbaMobilev9openapi.json` before and after this change and notice all of those `IntegerField`s go from strings to ints in the specs

tweirtx:
I'll test later with my Android off of this branch to verify

ZachOrr:
Bumping this

ZachOrr:
We can always roll back if something breaks. Thanks!

PR #2306, opened 2018-09-22
These tests seem to randomly fail so I deleted them

tweirtx:
## Description

I've noticed that the timezone tests are extremely inconsistent about whether or not they work and constantly fail builds for PRs that are actually perfectly fine. This pull request removes those tests so that we can see if it's the PR's code itself that's making it fail instead of having to check if errant timezone tests screwed everything up again.

## Motivation and Context

Pull request builds are failing with no reason other than these tests acting up.

## How Has This Been Tested?

A test run with no changed code except for this passed.

image

## Types of changes

### DIFF
--- a/tests/test_event_get_timezone_id.py
+++ b/tests/test_event_get_timezone_id.py
@@ -1,137 +1,137 @@
-import unittest2
-
-from google.appengine.ext import ndb
-from google.appengine.ext import testbed
-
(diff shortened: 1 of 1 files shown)


fangeugene:
Can you just comment them out instead, so we can go back to try to fix them in the future? I tried briefly to debug it, but I couldn't figure out why they're failing.

tweirtx:
there we go, the entire file is now commented out

PR #2312, opened 2018-09-29
Add Eventwizard offseason permission

phil-lopreiato:
## Description
Add a new permission, `EVENTWIZARD`, that we can grant to allow mods to import FMS reports for offseason events

## Motivation and Context
I'm tired of doing it myself

## How Has This Been Tested?
Added unit tests

### DIFF
--- a/consts/account_permissions.py
+++ b/consts/account_permissions.py
@@ -4,6 +4,7 @@ class AccountPermissions(object):
     REVIEW_APIWRITE = 3
     REVIEW_DESIGNS = 4
     REVIEW_EVENT_MEDIA = 5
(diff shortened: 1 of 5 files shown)


gregmarra:
Could the name be more specific? “Offseason event admin”?

The name doesn’t actually say what you get.

phil-lopreiato:
Renamed the permission to `OFFSEASON_EVENTWIZARD` and made sure it was only valid for the current year's events. Still need to add unit tests...

phil-lopreiato:
Fleshed out unit tests for the new cases, this should be good to go

PR #2334, opened 2018-12-01
Add Ping endpoint to myTBA

ZachOrr:
Adds a Ping endpoint to myTBA to dispatch a test notification for a user.

## Motivation and Context
This is a next-step in https://github.com/the-blue-alliance/the-blue-alliance-ios/pull/345 which helps users debug information about push notifications. Allowing users to kickoff a push will help us test end-to-end that things are working as expected.

## How Has This Been Tested?
It has not ⚠️  There doesn't seem to be any `NotificationHelper` tests, and I don't believe there's documentation on how to get a push setup locally. This seems fairly low-risk, since it's adding new functionality and not modifying existing functionality.

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

PR #2335, opened 2018-12-04
Remove auth from ping endpoint

ZachOrr:
We can't seem to get auth working for the ping endpoint added in https://github.com/the-blue-alliance/the-blue-alliance/pull/2334 - this removes that authentication


ZachOrr:
I'm fairly certain this is because we haven't updated the OpenAPI spec to include the ping endpoint. Will PR to add a step to deploy to create/deploy/update this.

PR #2337, opened 2018-12-04
Fix UpcomingMatch notification to send predicted_time

### LINKED ISSUE #2254: Webhook API docs have wrong description
The documentation for the webhook notification "upcoming match" has an incorrect description of the field `predicted_time`.  The docs say that `predicted_time` is the "time that the match is predicted to start at based on the times that scores were posted for previous events, in UNIX UTC."  However in the code for the notifications, namely [this line](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/notifications/upcoming_match.py#L33), `predicted_time` is set to the same value as `scheduled_time`.

ZachOrr:
Fixes #2254 to properly send `predicted_time` in the `UpcomingMatchNotification`. Also updates the documentation to better clarify when these fields will or won't have data.

I updated the tests in `UpcomingMatchNotification` as well, but we never predict times for matches in our tests, so the `predicted_time` isn't really tested and will always be `None`. If anyone feels strongly about adding some way to add predictions to matches in `MatchTestCreator`, I'm more than willing to do that work.

### DIFF
--- a/notifications/upcoming_match.py
+++ b/notifications/upcoming_match.py
@@ -28,11 +28,15 @@ class UpcomingMatchNotification(BaseNotification):
         data['message_data']['event_name'] = self.event.name
         data['message_data']['match_key'] = self.match.key_name
         data['message_data']['team_keys'] = self.match.team_key_names
+
         if self.match.time:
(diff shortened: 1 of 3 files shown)


ZachOrr:
@phil-lopreiato This should be good to go :v:

PR #2339, opened 2018-12-07
Automatically deploy endpoints config

ZachOrr:
During the deploy phase in Travis, check if our endpoints configuration has changed, and deploy the new configuration if necessary.

We shouldn't need to specify the `ENDPOINTS_SERVICE_VERSION` - endpoints will default to using the latest deployed version https://github.com/cloudendpoints/endpoints-management-python/pull/69

Note: this does move the `should_deploy` check to later, after setting up Python and gcloud and whatnot.

## Motivation and Context
Automate something that's currently manual and requires keys

### DIFF
--- a/controllers/apiv3/api_admin_controller.py
+++ b/controllers/apiv3/api_admin_controller.py
@@ -30,12 +30,14 @@ class ApiAdminSetBuildInfo(ApiAdminController):
         commit_time = data.get('commit_time', '')
         deploy_time = data.get('deploy_time', '')
         travis_job = data.get('travis_job', '')
+        endpoints_sha = data.get('endpoints_sha', '')
 
         web_info = {
             'current_commit': current_commit_sha,
             'commit_time': commit_time,
             'deploy_time': deploy_time,
             'travis_job': travis_job,
+            'endpoints_sha': endpoints_sha,
         }
 
         status_sitevar = Sitevar.get_or_insert('apistatus', values_json='{}')
--- a/ops/travis/should-deploy.sh
+++ b/ops/travis/should-deploy.sh
@@ -11,8 +11,9 @@ update_build_info() {
     commit_time="$(git show -s --format=%ci HEAD)"
     deploy_time="$(date)"
     travis_job="$TRAVIS_BUILD_ID"
+    endpoints_sha=$(sha1sum tbaMobilev9openapi.json | awk '{ echo $1 }')
 
-    http $SET_BUILD_STATUS X-TBA-Auth-Key:$APIv3_KEY current_commit="$current_commit" commit_time="$commit_time" deploy_time="$deploy_time" travis_job="$travis_job"
+    http $SET_BUILD_STATUS X-TBA-Auth-Key:$APIv3_KEY current_commit="$current_commit" commit_time="$commit_time" deploy_time="$deploy_time" travis_job="$travis_job" endpoints_sha="$endpoints_sha"
 }
 
 check_killswitch() {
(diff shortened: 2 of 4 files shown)


controllers/apiv3/api_admin_controller.py
```diff
@@ -30,12 +30,14 @@ def _render(self):
         commit_time = data.get('commit_time', '')
         deploy_time = data.get('deploy_time', '')
         travis_job = data.get('travis_job', '')
+        endpoints_sha = data.get('endpoints_sha', '')
```
phil-lopreiato: Do I need to see this manually for the first deploy?

ops/travis/travis-deploy.sh
```diff
@@ -23,8 +23,6 @@ with_python27() {
     bash -c "source $HOME/virtualenv/python2.7/bin/activate; $1"
 }
 
-should_deploy
```
phil-lopreiato: I still think this should be the first thing - we should maintain a way to killswitch the entire continuous pipeline if we need to

ops/travis/should-deploy.sh
```diff
@@ -45,6 +45,31 @@ check_overwrite_old() {
     fi
 }
 
+update_endpoints_config_info() {
```
phil-lopreiato: This won't work. There's a whitelist of expected keys and you need to pass them all up together. See https://github.com/the-blue-alliance/the-blue-alliance/blob/master/controllers/apiv3/api_admin_controller.py#L28

ZachOrr:
Alright, maybe something like this? The idea was to delete that `tbaMobilev9openapi.json` after generating it, but it requires storing/passing around that file's sha for updating the build info, which seems weird. I figure keeping it around isn't a problem, so it's easier to just keep it around and let it get uploaded with the web deploy.

ops/travis/should-deploy.sh
```diff
@@ -11,8 +11,9 @@ update_build_info() {
     commit_time="$(git show -s --format=%ci HEAD)"
     deploy_time="$(date)"
     travis_job="$TRAVIS_BUILD_ID"
+    endpoints_sha=$(sha1sum tbaMobilev9openapi.json | awk '{ echo $1 }')
 
-    http $SET_BUILD_STATUS X-TBA-Auth-Key:$APIv3_KEY current_commit="$current_commit" commit_time="$commit_time" deploy_time="$deploy_time" travis_job="$travis_job"
+    http $SET_BUILD_STATUS X-TBA-Auth-Key:$APIv3_KEY current_commit="$current_commit" commit_time="$commit_time" deploy_time="$deploy_time" travis_job="$travis_job" endpoints_sha="$endpoints_sha"
```
phil-lopreiato: You're still going to need to have the server read this value here: https://github.com/the-blue-alliance/the-blue-alliance/blob/master/controllers/apiv3/api_admin_controller.py#L32

ops/travis/travis-deploy.sh
```diff
@@ -36,6 +36,12 @@ with_python27 "$BOOTSTRAP --usage-reporting=false --command-completion=false --p
 
 PATH=$PATH:$INSTALL/$NAME/bin/
 
+# Check if we need to deploy our Endpoints config - cleanup afterwards so it's not in our web deploy
```
phil-lopreiato: Can we do this after the code deploy?

That'll solve our push-safety problem, since by that point, the server will be expecting the `endpoints_sha` param to be sent to the apistatus update endpoint

ZachOrr: Uhh yeah we can! Good call.

PR #2340, opened 2018-12-09
Attempt to fix web deploys

ZachOrr:
Web deploys have been stuck since https://github.com/the-blue-alliance/the-blue-alliance/commit/893056fe0f3a17ab9107d9363c18a0f4c0fbeaaf - it looks like this is a problem others using Travis have fixed via this `BOTO_CONFIG` hack

As noted in Slack, it looks like what changed is we upgraded from Cloud SDK 226.0.0 -> Cloud SDK 227.0.0 and this broke


phil-lopreiato:
Let's give it a try!

PR #2341, opened 2018-12-11
Fix crash when generating endpoints configuration

ZachOrr:
https://github.com/the-blue-alliance/the-blue-alliance/commit/19034851d81767de972dd6c93a41e03a0a779768 broke deploys (well, only the last half of deploys - we're still successfully deploying to the web), because we're running in to an error when generating the endpoints configuration. That `ANDROID_AUDIENCE: True` line is the problem - the endpoints config script complains about a missing issuer

```
TypeError: Missing issuer 
```

I'm not entirely sure what it's looking for here, since from what I can tell, that `ANDROID_AUDIENCE` value isn't going to be one of the values in our issuers. Additionally, I can't find any documentation about what that line is actually doing.

Finally, I added a step to the `LINT` job that generates our endpoints configuration (and cleaned up the formatting), since it's fairly cheap, and we should probably make sure that works before merging PRs 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_install:
 install:
 - travis_retry ./ops/travis/travis-install.sh $JOB
 - export PYTHONPATH=${PYTHONPATH}:${HOME}/google_appengine
+- export ENDPOINTS_GAE_SDK=$(gcloud info --format="value(installation.sdk_root)")
 before_script:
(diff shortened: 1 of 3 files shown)


ZachOrr:
I'm working on getting that generation step working - the libs should be installed for this to be able to be run at this stage, I think it's just a path problem. I could be wrong though 🤷‍♀️

PR #2347, opened 2018-12-17
[modtools] Implement Team Media Management

phil-lopreiato:
Give team media moderators the ability to remove/make preferred media.

## Description
Does a few things:
 - skeleton out team mod dashboard
 - hook up removing references/managing preferred-ness

This does not yet deal with social media or the suggestion queue. It should probably also implement a modal confirmation before doing writes. And we should also have logging for admin actions...

image


gregmarra:
We can log admin actions as immediately accepted suggestions.

PR #2349, opened 2018-12-17
Build/deploy React Native bundle with web app

ZachOrr:
For https://github.com/the-blue-alliance/the-blue-alliance-ios/issues/138 - looking to build/host the React Native bundle used to supplement some views in the the iOS app on the web


phil-lopreiato:
I'm happy to merge this, but do you think that this is the right approach (as opposed to putting all the RN stuff in this repo)? This way, there's no guarantee that RN changes get deployed in a timely manor if there are no www commits.

fangeugene:
Is a separate repo that deploys to something like firebase hosting sound like a better option?

ZachOrr:
If we want to host on Firebase, we have this PR ready to merge which will deploy to Firebase when there's changes in the RN branch https://github.com/the-blue-alliance/the-blue-alliance-react/pull/7

@phil-lopreiato We pull the RN code in to the iOS repo so we can deploy a copy locally with the iOS app, so it shouldn't live in the web repo

phil-lopreiato:
Woah hey, we should probably just merge that other PR...

ZachOrr:
We will have to setup the service worker stuff, but also all of the RN service code in the iOS app interfaces with Firebase Storage so that's all set

Just waiting on a decision on this one - feel free to close this PR if we wanna go with the other PR

phil-lopreiato:
I think the other PR is the right approach, although I think there's a bit of work on my end to set up the right keys for the prod instance

PR #2350, opened 2018-12-18
[modtools] Basic suggestion queue support

phil-lopreiato:
A mostly-working implementation for the suggestion queue

## Description
Render pending suggestions for the mod dashboard. This does some horribly hacky things like posting to the main review controllers (and thankfully they all expect the same format of data input). Which means it _does not_ yet do permissions - you need to have the global review permission for this to work (for now)

image

### DIFF
--- a/controllers/team_admin_controller.py
+++ b/controllers/team_admin_controller.py
@@ -16,6 +16,16 @@ from template_engine import jinja2_engine
 
 class TeamAdminDashboard(LoggedInHandler):
     ALLOWED_SUGGESTION_TYPES = ["media", "social-media", "robot"]
+    SUGGESTION_NAMES = {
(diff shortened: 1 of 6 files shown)


controllers/team_admin_controller.py
```diff
@@ -16,6 +16,16 @@
 
 class TeamAdminDashboard(LoggedInHandler):
     ALLOWED_SUGGESTION_TYPES = ["media", "social-media", "robot"]
+    SUGGESTION_NAMES = {
+        "media": "Media",
+        "social-media": "Social Media",
+        "robot": "Robot CAD",
+    }
+    SUGGESTION_REVIEW_URL = {
```
phil-lopreiato: ugh

PR #2355, opened 2018-12-24
TBA Notification Service, using Firebase HTTPv1 API (Ping and Verification)

ZachOrr:
Throwing this up so people can start to get eyes on it. This is very much a first step, and not a complete system. Reviews are appreciated.

## Description
This introduces a new service - The Blue Alliance Notification Service (or TBANS). TBANS manages dispatching notifications to clients and webhooks, and managing the lifecycle of notifications (handling retries, handling failures, etc.).

Up for discussion is if TBANS should manage client subscriptions, or if that should be managed by another piece of the codebase. (These subscriptions refer to Firebase topics - which will be the new system for sending notifications to multiple clients, as opposed to using an array of Firebase tokens, which isn’t supported anymore. Refer to the [Topic Messaging](https://firebase.google.com/docs/cloud-messaging/js/topic-messaging) section of the FCM documentation)

For now, TBANS only manages the immediate dispatching of ping and webhook verifications. Other notifications will be built out, but I need to think more about the subscription stuff mentioned above.

Upcoming Match, Update Favorites, and Update Subscriptions notifications have been included in this PR (along with their tests), although they’re not used yet. This is to give a sense of how to use notifications with the new objects.

## Motivation and Context
The old push notification system (built on the legacy Firebase API) won’t work for iOS clients moving forward. The notification payloads specified by the current notification objects aren’t structured in a way that iOS is able to make sense of them and handle them. I’m unsure about if the web will or won’t support the current notification payloads.

Additionally, the legacy notification API hasn’t been receiving updates (it’s missing some new notification features in iOS), and it doesn’t seem to support data notification on iOS (I have no idea why this is - regular title/body notifications work fine, I can’t get a data message to deliver).

Since our push notification and webhook stack are so tightly coupled, TBANS will now manage dispatching messages to webhooks as well.

By splitting notifications in to it’s own service, other TBA services shouldn’t need to be concerned with constructing notification objects or sending them - only calling TBANS.

Moving forward - this NotificationHelper will exist and act as a router to figure out if we should dispatch the old notifications for Android, or the new notifications for other platforms. Once we can get Android moved over to the new notification system, we should be able to remove this code. But for now, not breaking Android is a priority.

## How Has This Been Tested?
Unit tests, along with manual testing with webhook verification in a test instance, and pings via the ping endpoint in the mobile API

## Screenshots
Here's a Ping sent via the mobile API
img_0717


deploy_requirements.txt
```diff
@@ -7,7 +7,7 @@ beautifulsoup4
 GoogleAppEngineCloudStorageClient
 
 # Cloud Endpoints Stuff
-oauth2client==3.0.0
+oauth2client==4.1.2
```
ZachOrr: We probably don't need to bump this all the way up - however, the service worker stuff I'm using to generate an OAuth token for dispatching notifications isn't available in that `3.0.0`

helpers/notification_helper.py
```diff
@@ -155,17 +155,9 @@ def send_broadcast(cls, client_types, title, message, url, app_version=''):
 
     @classmethod
     def send_ping(cls, client):
-        if client.client_type == ClientType.WEBHOOK:
-            keys = {client.client_type: [(client.messaging_id, client.secret)]}
+        if client.client_type == ClientType.OS_ANDROID:
```
ZachOrr: Moving forward - this `NotificationHelper` will exist and act as a router to figure out if we should dispatch the old notifications for Android, or the new notifications for other platforms. Once we can get Android moved over to the new notification system, we should be able to remove this code. But for now, not breaking Android is a priority.

helpers/tbans_helper.py
```diff
@@ -0,0 +1,45 @@
...
+class TBANSHelper(object):
+
+    """
+    Helper class for sending push notifications.
+    Methods here should call the TBANS API to dispatch/queue notifications
+    """
+
+    @staticmethod
+    def _create_service():
+        return TBANSService.Stub(transport.HttpTransport('https://tbatv-prod-hrd.appspot.com/tbans/tbans'))
```
ZachOrr: Is there a way we can grab this domain/host (`tbatv-prod-hrd.appspot.com`) dynamically? The idea is that we want this to work seamlessly in a dev container, but also in other people's prod instances

tbans/controllers/fcm/fcm.py
```diff
@@ -0,0 +1,100 @@
...
+        'INTERNAL': INTERNAL_ERROR,
+        'INVALID_ARGUMENT': 'invalid-argument',
+        'QUOTA_EXCEEDED': 'message-rate-exceeded',
+        'SENDER_ID_MISMATCH': 'mismatched-credential',
+        'UNAVAILABLE': 'server-unavailable',
+        'UNREGISTERED': 'registration-token-not-registered',
+    }
+
+    def __init__(self, project_id='tbatv-prod-hrd'):
+        self._FCM_URL = 'https://fcm.googleapis.com/v1/projects/{project_id}/messages:send'.format(project_id=project_id)
```
ZachOrr: Another spot where we need to figure out if we can pull that `project_id` dynamically

tbans/controllers/fcm/fcm.py
```diff
@@ -0,0 +1,100 @@
...
+            self._handle_fcm_error(error)
+        except Exception, error:
+            msg = 'Failed to call messaging API: {}'.format(error)
+            raise ApiCallError(self.INTERNAL_ERROR, msg, error)
+
+    # From https://github.com/firebase/firebase-admin-python/blob/932cf17a6f222c627dbd1502658f3eb338077250/firebase_admin/messaging.py
+
+    # Retrieve a valid access token that can be used to authorize requests.
+    def _get_access_token(self):
+        credentials = ServiceAccountCredentials.from_json_keyfile_name('zach-tba-dev-firebase-adminsdk-355dn-5fe528f891.json', scopes=['https://www.googleapis.com/auth/firebase.messaging'])
```
ZachOrr: We'll need to figure out the best way to do this on prod

tbans/models/notifications/upcoming_match.py
```diff
@@ -0,0 +1,65 @@
...
+    @staticmethod
+    def _type():
+        return NotificationType.UPCOMING_MATCH
+
+    @property
+    def notification_payload(self):
+        # Title: Midwest Regional Finals 1 Match 2 Starting Soon
+        # Body: 111, 200, 1984 will play 112, 201, 1985
+        title = '{} {} Starting Soon'.format(self.event.display_name, self.match.verbose_name)
+        # TODO: There has to be a standard way to remove this `frc` prefix
```
ZachOrr: Some questions for someone more familiar with the codebase to advise on please

.gitignore
```diff
@@ -90,3 +90,5 @@ test_keys.json
 ###########
 .vagrant/
 dev-datastore/
+
+zach-tba-dev-firebase-adminsdk-355dn-5fe528f891.json
```
gregmarra: Don't forget to remove this zach-specific line, if it is zach specific.

tbans/controllers/fcm/fcm.py
gregmarra: Worst case these should at least be in constants somewhere common.

tbans/models/messages/webhook.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    JSON for WebhookMessage will look like...
+    {
+        'message_data': {...},
+        'message_type': ...
+    }
+    """
+    def json_string(self):
+        json_dict = {
+            'message_type': NotificationType.type_names[type(self.notification)._type()]
```
ZachOrr: Design question - it seems like we should send this `message_type` automatically in the data payloads for push notifications, right? That's probably going to be our only way to identify what type of notification is coming from the server, and how to respond to it in clients.

ZachOrr:
It's also worth asking - what's going to be the cost of spinning this up in a new service?

ZachOrr:
Hey y'all - this is ready for a review. I need to write a little more documentation about how to test things manually/locally and how notifications are structured and stuff, add tests for the TBANS helper (I'm unsure how to mock that for the moment)/TBANSService, and then add some logging. But code-wise, this should be good.

ZachOrr:
@bdaroz All of the notifications in this PR (just pings and verification) don't break existing webhook functionality. Moving forward, my goal is to make sure we don't break webhooks or have to bump the webhook version, unless we have a good reason.

bdaroz:
Ok, I just saw the webhook docs get changed so I wasn't sure.

ZachOrr:
Ah yeah - I rolled back that doc change. Right now we are sending a webcast object, it’s just undocumented.

controllers/account_controller.py
```diff
@@ -14,7 +14,7 @@
 from consts.auth_type import AuthType
 from consts.client_type import ClientType
 from consts.model_type import ModelType
-from consts.notification_type import NotificationType
+from tbans.consts.notification_type import NotificationType
```
ZachOrr: As a heads up - this has moved

tbans/tbans.yaml
```diff
@@ -0,0 +1,17 @@
...
+threadsafe: true
+
+builtins:
+- appstats: on
+- deferred: on
+
+handlers:
+- url: .*
+  script: tbans_service.app
+  # login: admin
```
ZachOrr: This needs to be fixed - I keep getting 302's when hitting this from the mobile API

.travis.yml
```diff
@@ -12,6 +12,7 @@ cache:
 env:
   matrix:
   - JOB=PYUNIT
+  - JOB=TBANS_TESTS
```
ZachOrr: Unless someone can help me get the tests running via the Python script that runs our tests :)

.travis.yml
```diff
@@ -23,7 +24,7 @@ before_install:
 - nvm install 8.0.0
 install:
 - travis_retry ./ops/travis/travis-install.sh $JOB
-- export PYTHONPATH=${PYTHONPATH}:${HOME}/google_appengine
+- export PYTHONPATH=${PYTHONPATH}:${HOME}/google_appengine:${HOME}/google_appengine/lib/protorpc-1.0
```
ZachOrr: If we can figure out how to get everything from that `lib` folder on the path, that would be hella

tbans/tbans_service.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class TBANSService(remote.Service):
+    """ The Blue Alliance Notification Service.... Service """
+
+    @remote.method(PingRequest, TBANSResponse)
+    def ping(self, request):
+        """ Immediately dispatch a Ping to either FCM or a webhook """
+        if request.fcm and request.webhook:
+            return TBANSResponse(code=400, message='Cannot ping both FCM and webhook')
```
ZachOrr: I'm not familiar enough with RPC responses - this feels like we should be able to set a status code for the request as opposed to just setting a code in an object, since this will 200 and we'll lose some clarity around what's going on in logs/debugging

ZachOrr:
Removing WIP - this is ready for review. I'll need help figuring out why the tests are failing.

phil-lopreiato:
Let's give it a whirl

PR #2049, opened 2018-01-02
Decommission APIv2

fangeugene:
## Description
Decommissions APIv2 after January 1, 2018.

## Motivation and Context
APIv3 is the primary API.

## How Has This Been Tested?
Trying to access any APIv2 endpoint returns:
```
{"Error": "As of Jan. 1 2018, v2 of the TBA API has been decommissioned. Please see http://www.thebluealliance.com/apidocs for more info."}
```

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would change API specifications or require data migrations)

One of the most breaking changes possible 😆

PR #2359, opened 2018-12-29
Nuke APIv2

brandondean:
## Description
Starts with Eugene's #2049, but actually has a few deletes. If there's some sweaty palms about this nuking than #2049 can be merged first, this rebased, and then merged later when there's more confidence.

## Types of changes


fangeugene:
Don't need this anymore, we didnt port apiv2 to py3

PR #2364, opened 2018-12-31
Hotfix the CD media links

dracco1993:
## Description
Hotfix the CD images, for now.

## Motivation and Context
CD media is currently entirely foobar'ed on the site, and we're not sure yet when things will be back up, or when the permanent migration/solution will be available.  In the meantime, we can link the images from the Wayback Machine.

## How Has This Been Tested?
Locally, everything is working as intended.  See screenshots below.
I used the following page for testing:
https://www.thebluealliance.com/team/2056/2014

## Screenshots:
Before the fix:

After the fix:

## Types of changes

### DIFF
--- a/templates_jinja2/media_partials/cdphotothread_partial.html
+++ b/templates_jinja2/media_partials/cdphotothread_partial.html
@@ -1,8 +1,8 @@
 <div class="thumbnail cdphotothread-thumbnail">
-  <a href="{{media.image_direct_url}}" class="gallery">
-    <span style="background-image:url('{{media.image_direct_url_med}}')"></span>
+  <a href="https://web.archive.org/web/{{media.image_direct_url}}" class="gallery">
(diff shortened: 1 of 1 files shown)


bdaroz:
Rather than hack the HTML in that one spot, as IIRC the images are used in more than the team detail page (event team list comes to mind), I would suggest editing this line: https://github.com/the-blue-alliance/the-blue-alliance/blob/fe43a0ce3b1bf2f74de945765f4e04f37eb6112d/models/media.py#L100

And to get the tests working again you'll likely need to comment out this: https://github.com/the-blue-alliance/the-blue-alliance/blob/1324e9e5b7c4ab21315bd00a768112991bada108/tests/suggestions/test_media_url_parse.py#L33

dracco1993:
@bdaroz Yeeeeeah.  It was 2 AM when I made that fix.  😅  I'll get a more proper fix in.  Lol

Issue #2366, opened 2019-01-01
Return media image url in apiv3


phil-lopreiato:
We want the apps to not have to construct media urls on their own - we should augment the existing apiv3 response to include the ones generated server-side. 

We'll want to pass `media.view_image_url` and `media.image_direct_url` into the output of `MediaConverter`

fangeugene:
Closed in #2368

PR #2368, opened 2019-01-01
Add view_url and direct_url to MediaConverter

LeoCodes21:
## Description
This PR updates the MediaConverter class and adds 2 new fields to the API-facing Media model: `view_url` and `direct_url`. These fields provide pre-formatted URLs for accessing media.

## Motivation and Context
Currently, applications using TBA's API to retrieve media data are required to do the URL generation work. This PR allows for easier access to proper, server-generated media URLs that can be used by consuming applications. 
This PR addresses issue #2366.

## How Has This Been Tested?
Inserted media from multiple sources into my local installation to ensure that my changes did not cause any issues with MediaConverter. These sources were Imgur, YouTube and Instagram. Note that Chief Delphi was unavailable at the time of testing, but I do not expect any issues to come up.

No screenshots available.

## Types of changes


static/swagger/api_v3.json
```diff
@@ -2,10 +2,10 @@
   "swagger": "2.0",
   "info": {
     "description": "# Overview \n\n Information and statistics about FIRST Robotics Competition teams and events. \n\n# Authentication \n\nAll endpoints require an Auth Key to be passed in the header `X-TBA-Auth-Key`. If you do not have an auth key yet, you can obtain one from your [Account Page](/account). \n\n A `User-Agent` header may need to be set to prevent a 403 Unauthorized error.",
-    "version": "3.03.1",
+    "version": "3.03.2",
```
bdaroz: This shouldn't bump the point release, but the minor release instead, as this does add additional fields to the models. (See `x-version-info` for rationale)

static/swagger/api_v3.json
```diff
@@ -5741,6 +5741,14 @@
         "preferred": {
           "type": "boolean",
           "description": "True if the media is of high quality."
+        },
+        "image_direct_url": {
+          "type": "string",
+          "description": "Direct URL to the media's image, if there is one."
```
bdaroz: How does this work with YouTube videos? Do they have `image_direct_url`s, and I presume the `view_image_url` drops to the YT page.

phil-lopreiato: I think that youtube videos should actually have this set (to the thumbnail image). See https://github.com/the-blue-alliance/the-blue-alliance-android/blob/5c94e11e5640a08a3547c16646efec3f018835c8/android/src/main/java/com/thebluealliance/androidclient/types/MediaType.java#L57

LeoCodes21: YouTube videos don't have `image_direct_url`s and `view_image_url`s. Technically they're not images, and they have different fields for URLs: `youtube_url_link` for the link to the video and `youtube_url` for the embed URL.

phil-lopreiato: That should be okay - we can add those properties for youtube (the rendering code should do its own checks on the media type first anyway)

static/swagger/api_v3.json
LeoCodes21: Resolved in the latest commit.

static/swagger/api_v3.json
bdaroz: Rather than adding a bunch of properties for YT (or maybe later ones) and ugly-ing the model up, can we make the fields more "generically named" and integrate them?

LeoCodes21: So, to be clear, should YouTube media have the image properties set? I don't see any way to get YouTube thumbnails in the code. So maybe we should generalize the names of the fields so that we can comfortably support YouTube links?

LeoCodes21: Oops. Ninja'd by @bdaroz

database/dict_converters/media_converter.py
```diff
@@ -20,9 +20,18 @@ def mediasConverter_v3(cls, medias):
...
+        dict = {
             'type': media.slug_name,
             'foreign_key': media.foreign_key,
             'details': media.details if media.details else {},
             'preferred': True if media.preferred_references != [] else False,
+            'view_url': None,
+            'direct_url': None,
         }
+        if media.slug_name == "youtube":
+            dict["direct_url"] = media.youtube_url
```
phil-lopreiato: For images, we actually want this to be the thumbnail url, see https://github.com/the-blue-alliance/the-blue-alliance-android/blob/5c94e11e5640a08a3547c16646efec3f018835c8/android/src/main/java/com/thebluealliance/androidclient/types/MediaType.java#L57

So `"http://img.youtube.com/vi/{foreignKey}/hqdefault.jpg"`

LeoCodes21: Resolved in the latest commit.

PR #2370, opened 2019-01-02
Fix 404ing Tasks from Admin Interface & location_helper API check

bdaroz:
First, remove the insights task enqueuing items from the Admin Task interface -- they have long since been removed from the backend. Second, the `location_helper` will attempt to geolocate events even if there is no Google API key present. In the past this was ok, a small number of requests were permitted without a key, but this is no longer the case.

## Motivation and Context
Both are dev instance quality-of-life items.

## How Has This Been Tested?
Tested on local dev instance.

## Types of changes

### DIFF
--- a/helpers/location_helper.py
+++ b/helpers/location_helper.py
@@ -500,7 +500,7 @@ class LocationHelper(object):
             google_secrets = Sitevar.get_by_id("google.secrets")
             google_api_key = None
             if google_secrets is None:
-                logging.warning("Missing sitevar: google.api_key. API calls rate limited by IP and may be over rate limit.")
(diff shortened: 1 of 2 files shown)


brandondean:
Does it make sense to add some curl requests to documentation for when you need to fire their replacements locally?

phil-lopreiato:
Actually, this was broken when these tasks go moved to a separate module, the new base url is `/backend-tasks-b2/math/enqueue/insights/`, if you'd rather keep the admin page around.

Otherwise, I just end looking at the urlmap when I can't remember a url :stuck_out_tongue:

PR #2390, opened 2019-02-09
[team-admin] Allow editing robot names

phil-lopreiato:
## Description
FIRST stopped allowing people to set robot names in 2019, so we can do that ourselves

image


phil-lopreiato:
Added cache warning
image

Issue #885, opened 2014-02-10
[gameday2] support x-out hiding of videos


gregmarra:
Support hiding videos by allowing them to be x-ed out in the onHover overlay.

fangeugene:
Gameday2 shipped, this is not an issue.

PR #2392, opened 2019-02-14
Use HTTPS for YouTube image URLs

bherbst:
## Description
The direct image URLs for YouTube videos are currently exposed as HTTP links instead of HTTPS links. This changes switches them to use HTTPS.

## Motivation and Context
The Android app would like to use the media URLs exposed via the API instead of generating those URLs ourselves (see [#885](https://github.com/the-blue-alliance/the-blue-alliance-android/issues/885)). However, as of Android 9 (Pie), cleartext traffic is disabled by default. While we could enable it here, there is no compelling reason to do so instead of just getting HTTPS links to begin with.

## How Has This Been Tested?
I set up a dev environment with this change, and observed no regression on the team pages under media (which makes sense, as it does not appear as though the web site is consuming this- it is exposed for other clients).

The HTTP versions of these URLs redirect to the HTTPS versions anyway.

## Types of changes

### DIFF
--- a/database/dict_converters/media_converter.py
+++ b/database/dict_converters/media_converter.py
@@ -29,7 +29,7 @@ class MediaConverter(ConverterBase):
             'direct_url': None,
         }
         if media.slug_name == "youtube":
-            dict["direct_url"] = "http://img.youtube.com/vi/{}/hqdefault.jpg".format(media.foreign_key)
+            dict["direct_url"] = "https://img.youtube.com/vi/{}/hqdefault.jpg".format(media.foreign_key)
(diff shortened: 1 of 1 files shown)


bherbst:
@fangeugene thanks! I didn't *think* I had to bump that since the structure of the dict wasn't changing, but I also wasn't convinced.

fangeugene:
LGTM. @phil-lopreiato ?

ZachOrr:
🤔 I'm surprised this isn't breaking iOS. IIRC we have ATS on. Although - the YouTube library we use is pretty buggy, so maybe it is

PR #2393, opened 2019-02-14
Move ChiefDelphi web archive URLs to a direct image URL

bherbst:
## Description
Switches ChiefDelphi direct image URLs to using a direct image link on web.archive.org.

## Motivation and Context
The Android app would like to use the media URLs exposed via the API instead of generating those URLs ourselves (see [#885](https://github.com/the-blue-alliance/the-blue-alliance-android/issues/885)).

However, the existing CD media links exposed to the app include the internet archive's iframe, and end up not rendering correctly in the app (I'm really not sure how web's `background-image` CSS attributes do it properly).

The internet archive supports flags to disable this and return an image directly (the `im_` flag). Unfortunately this can only be used with a specific entry in the archive (e.g. a timestamp), but using a timestamp of "0" gives you the most recent version.

**Old:** https://web.archive.org/web/https://www.chiefdelphi.com/media/img/a88/a880fa0d65c6b49ddb93323bc7d2e901_l.jpg

**New:**
[https://web.archive.org/web/**0im_**/https://www.chiefdelphi.com/media/img/a88/a880fa0d65c6b49ddb93323bc7d2e901_l.jpg](https://web.archive.org/web/0im_/https://www.chiefdelphi.com/media/img/a88/a880fa0d65c6b49ddb93323bc7d2e901_l.jpg)

More on these types of flags:
 * https://en.wikipedia.org/wiki/Help:Using_the_Wayback_Machine#Specific_archive_copy
 * https://webapps.stackexchange.com/questions/40911/web-archive-links-without-header

This is basically a tweak to #2364 , so I'm hoping a more robust solution comes at some point.

## How Has This Been Tested?
I mostly swapped out the URLs in Chrome's dev tools to confirm they still showed up correctly on TBA web, but I also made this change in a local dev environment and confirmed the URLs were coming through as I expected.

The existing unit tests for these URLs were taken out in #2364 .

## Types of changes

### DIFF
--- a/models/media.py
+++ b/models/media.py
@@ -97,7 +97,7 @@ class Media(ndb.Model):
     # URL renderers
     @property
     def cdphotothread_image_url(self):
-        return 'https://web.archive.org/web/https://www.chiefdelphi.com/media/img/{}'.format(self.details['image_partial'])
+        return 'https://web.archive.org/web/0im_/https://www.chiefdelphi.com/media/img/{}'.format(self.details['image_partial'])
 
     @property
     def cdphotothread_image_url_med(self):


fangeugene:
@bdaroz what's the status on the migration mapping? Should we just do that instead of keeping around the archive workaround?

phil-lopreiato:
I'm fine with making this change until we get around to running a full migration - this seems non-threatening enough

PR #2395, opened 2019-02-15
Update rankings, match breakdowns, and event insights for 2019

fangeugene:
The mock data from the API doesn't have realistic data, so numbers don't necessarily add up.
image

image

## Types of changes


gregmarra:
One suggestion would be to show RP dots in the boxes that also state the RP
numbers to further associate the dots in the small scoreboxes with RP's.

fangeugene:
@gregmarra  like this? 
image

Issue #2399, opened 2019-02-17
APIv3 returning Awards with null `awardee` and `team_key`


ZachOrr:
There's a handful of event award endpoints that are returning null for both the `awardee` and `team_key` - they look to mostly be around rookie awards

`https://www.thebluealliance.com/api/v3/event/2018miwmi/awards` (Rookie Inspiration)
`https://www.thebluealliance.com/api/v3/event/2017qcmo/awards` (Rookie Inspiration)
`https://www.thebluealliance.com/api/v3/event/2018mosl/awards` (Rookie All Star)

This is crashing iOS, since we expect an Award to be given to [either a team or an individual](https://github.com/the-blue-alliance/the-blue-alliance-ios/blob/master/the-blue-alliance-ios/Core%20Data/Award/AwardRecipient.swift#L61).

The docs seem unclear about if returning an Award with no `awardee` or `team_key` is expected behavior (or should we just not return the Award)

<img width="1031" alt="screen shot 2019-02-17 at 12 05 05 pm" src="https://user-images.githubusercontent.com/516458/52916362-70381c80-32ac-11e9-8c0a-daf07443b126.png">

dracco1993:
It seems, according to the FIRST API, that it is not guaranteed that either of those fields will be not `null`:
```
{
  "awardId": 643,
  "teamId": null,
  "eventId": 27921,
  "eventDivisionId": null,
  "eventCode": "MIWMI",
  "name": "Rookie Inspiration Award sponsored by National Instruments",
  "series": 1,
  "teamNumber": null,
  "schoolName": null,
  "fullTeamName": "",
  "person": null
}
```
from https://frc-api.firstinspires.org/v2.0/2018/awards/miwmi

That being said, I _do_ think we still want these surfaced these in the API, as the fact that the award was available, but not given out is information that we'd lose if we just omitted it form the list, since not all events are eligible to give out these awards.

Both of the awards listed, and I believe those are the only two awards that could fall into this category, have this information as part of their [descriptions](https://www.firstinspires.org/resource-library/frc/awards-based-on-team-attributes):
> As there are often far fewer rookie teams than veteran teams present at events, Judges have the option of not presenting this award if they feel no rookie team competing meets the criteria.

ZachOrr:
Going to close this - we've got an open PR to address the documentation, but this is expected behavior.

PR #2400, opened 2019-02-17
Update Award_Recipient docs to list that awardee and team_key may be null

ZachOrr:
I'm open to fixing the wording on this one - just want to make sure we call out that both fields may be null in some cases, since the current docs don't say that.

### DIFF
--- a/static/swagger/api_v3.json
+++ b/static/swagger/api_v3.json
@@ -5728,7 +5728,7 @@
             "instagram-image",
             "external-link",
             "avatar"
-          ] 
+          ]
         },
         "foreign_key": {
           "type": "string",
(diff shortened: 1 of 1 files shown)


static/swagger/api_v3.json
```diff
@@ -5728,7 +5728,7 @@
             "instagram-image",
             "external-link",
             "avatar"
-          ]
```
ZachOrr: wat

ZachOrr:
Related to #2399

static/swagger/api_v3.json
phil-lopreiato: trailing whitespace is bad?

PR #2410, opened 2019-02-24
Return None for Team motto

ZachOrr:
We're moving away from supporting Team mottos - this will (hopefully) catch all places in APIv3 that we return a Team motto and return None instead.

This shouldn't affect APIv2 - which will keep returning mottos

Note: Motto stuff still exists in the bootstrap script, the model itself, visible on some admin page, and the parsers. Eventually hopefully when we kill motto altogether at the end of the season we can remove it everywhere.


FletcherS7:
I hold the opinion that the Team Motto's should be moved into myTBA team admin instead of being removed.

gregmarra:
@ZachOrr  this breaks API tests that expect team mottos to be returned :)

ZachOrr:
This was meant to be a draft PR but I clicked the wrong button 🙃 I haven't gotten back to this yet. I still need to update this branch to null out APIv2 motto responses.

database/dict_converters/team_converter.py
```diff
@@ -3,7 +3,7 @@
 
 class TeamConverter(ConverterBase):
     SUBVERSIONS = {  # Increment every time a change to the dict is made
-        3: 4,
+        3: 5,
```
ZachOrr: @bdaroz Do we need to bump this?

bdaroz: I don't believe so because the dict isn't changing -- we're still holding spot for the value -- just nulling it.

Also, the version number for the APIv3 Swagger spec (and changelog in it) should be bumped too.

fangeugene: You do need to bump it.

ZachOrr: Can you point me in the right direction for those things?

fangeugene: You already bumped it. Unless you're referring to Swagger in case @bdaroz chime in.

ZachOrr: Yep - talking about the Swagger version/changelog stuff

ZachOrr: Paging @bdaroz on this one

bdaroz: At an event this weekend. Will reply after.

ZachOrr:
This one should be good to go - we should try to communicate this change (via mailing list probably?) before we merge this.

If it looks good, we can discuss a timeline for these changes and I can send an email.

static/swagger/api_v3.json
```diff
@@ -2,10 +2,10 @@
...
...vent` model to correct type. Update `Media`.`type` enum. --- 3.0.3 - 3.0.4: Correct syntax in `Team_Event_Status_playoff` object, include descriptions in `Event_District_Points` and fix the path in `Team_Event_Status_playoff.properties.record` in order to more closely match Swagger spec. Fix syntax of `Match.video` to be correct and accurately reflect what is returned by the API. --- 3.0.2 - 3.0.3: `extra_stats` and `extra_stats_info` added to `Event_Ranking` model. Corrected match video property claiming to use the `Media` model, which it doesn't. Added `first_event_code` to the `Event` model. Added `/team/{team_key}/media/tag/{media_tag}` and year endpoints. Added `dq_team_keys` to `Match_alliance` model.  --- 3.0.1 - 3.0.2: `Team_Event_Status` and `Event_Ranking` model documentation changed to reflect actual API return. Changes involved W-L-T and ranking properties. --- 3.0.0 - 3.0.1: Descriptions updated to clarify some terms and improve grammar. The `district` property on the `Event` object now points to the `District_List` model, which was identical. `Required` fields on fully documented models now represent fields that may not be `null`. `District_Ranking` model documented."
```
bdaroz: To be consistent with the other change (L3517) we should say "null" not "None" here, the latter being a python-ism for nulls.

ZachOrr: Fixed!

bdaroz:
API Stuff LGTM. :shipit:

PR #2412, opened 2019-02-27
Add event week to myTBA

jaredhasenklein:
Add the event week to the date block on myTBA.

## Description
Uses the macro at `templates_jinja2/event_partials/event_macros.html` to add the event date with the week. Replaces the code for the event date with `{{ em.showEventDates(event) }}`. 

## Motivation and Context
The event week is now listed next to the date in other places, so it should be listed here, too.

## How Has This Been Tested?
I _think_ I did this right but I'm not positive.

image

## Types of changes

### DIFF
--- a/templates/mytba_live.html
+++ b/templates/mytba_live.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% import 'event_partials/event_macros.html' as em %}
 
 {% block title %}myTBA Dashboard - The Blue Alliance{% endblock %}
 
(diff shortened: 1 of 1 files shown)


fangeugene:
Had to revert. It's not a jinja2 template so we can't use macros.

gregmarra:
@jaredhasenklein , do you have a local dev environment set up where you can test these locally? I can walk you through getting local development working.

I put up https://github.com/the-blue-alliance/the-blue-alliance/pull/2413 as an example of a jinja1 -> jinja2 template conversion, which you could do at the same time as this to enable jinja macros.

jaredhasenklein:
@gregmarra I tried to get one going a little while ago and was having issues with it and never got around to finishing set up. Will look in to when not on my way to a Regional

gregmarra:
Feel free to ping me on Slack some time and we can try to walk through it live.

PR #2422, opened 2019-03-02
Make async webhook calls in TBANS

ZachOrr:
## Description

Change requests to webhooks via TBANS to be async instead of sync, since we don't really care about the server's response content or response code. Surfacing those things via TBANS felt nice for monitoring/logging, but I now realize we shouldn't need to care about those things, and the original code was misguided.

## Motivation and Context

We're seeing errors when calling TBANS that webhook request are timing out, because clients aren't handling requests in time (default timeout for `urlfetch` is 5s). Our protorpc request timeout to TBANS is also 5s. After digging through source, there doesn't appear to be a way to bump the protorpc request timeout.

Instead of bumping the response time down for clients, we can make calls to webhooks asynchronously and ignore their responses, since we don't care. This will speed up TBANS, meaning we'll always return from a webhook verification call in under 5s, and should fix the errors we're seeing.

The `ping` endpoint gets this functionality as well, which is nice, since we're also seeing some 500s there too with the same error.

## How Has This Been Tested?

Tested locally with a webhook that takes 5s+ to respond to a request

## Types of changes

### DIFF
--- a/tbans/models/messages/fcm_message.py
+++ b/tbans/models/messages/fcm_message.py
@@ -136,7 +136,7 @@ class FCMMessage(Message):
             response = urlfetch.fetch(
                 url=self._fcm_url,
                 payload=message_json,
-                method='POST',
+                method=urlfetch.POST,
                 headers=headers
             )
             return FCMMessage._transform_fcm_response(response)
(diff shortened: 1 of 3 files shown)


ZachOrr:
This should *also* fix the errors we're seeing where servers are returning us content with unicode and we're erroring because of it 🙇

ZachOrr:
We can also bump the timeout for webhooks to be higher than 5s if we want - Phil's original code didn't have an explicit timeout so I don't know what's appropriate.

fangeugene:
Is this good to go?

ZachOrr:
Yes! For now. Let’s get this merged and get feedback from users on what feels right for next steps (no action being an option too)

PR #2432, opened 2019-03-06
TBANS Subscription API

### LINKED ISSUE #1215: GCM Improvements
# io2015

Move myTBA notifications to a more pub/sub architecture 
https://developers.google.com/cloud-messaging/topic-messaging

Post notifications to relevant topics (which users are subscribed too), and GCM handles the distribution

ZachOrr:
## Description

This PR adds subscription management API endpoints to the TBANS service and consumes them via the client API. Closes #1215

## Motivation and Context

As part of our migration from the FCM legacy HTTP API to the new HTTPv1 API, we need to start [managing subscriptions](https://firebase.google.com/docs/cloud-messaging/manage-topics) with Firebase using their [Instance ID API](https://developers.google.com/instance-id/reference/server#get_information_about_app_instances). When mobile clients hit the client API, we'll dispatch subscription update calls to TBANS.

When TBANS gets a call to update subscriptions for a token or for an entire user, we queue two tasks. The first task is a pull task that describes the work that needs to be done. The second task is a push task to inform a worker that there's new work to be done for a given tag. A subscription worker gets notified via a push task that there's work to be done, and pulls all applicable tasks off the pull queue. We pull as many tasks as we can and flatten them down in to a single task, since multiple tasks for the same action (sub/unsub, or update) can be queued before a worker gets to it. We then take action on the task. If the task needs to get retried, we'll tell the push task to try again in a few seconds, and we'll keep the single flattened pull task in the queue. Otherwise, we'll delete our processed pull queue task.

TBANS also makes sure we only subscribe users to "valid" topics. If we know a topic will not get any messages (ex: some notifications for an event aren't applicable after the even is over), then we won't subscribe users to these topics. This is to avoid hitting the hard 2000 topic/device limit FCM imposes.

## How Has This Been Tested?

Along with a bunch of unit tests, this has been tested end-to-end using a dev instance of TBA and a dev instance of TBA for iOS to ensure tasks are getting dispatched and picked up properly.

## Types of changes

### DIFF
--- a/tbans/models/subscriptions/subscription_topic.py
+++ b/tbans/models/subscriptions/subscription_topic.py
@@ -138,11 +138,11 @@ class SubscriptionTopic:
 
         from consts.model_type import ModelType
         if model_type == ModelType.EVENT:
-            return ['event_{}_{}'.format(model_key, _notification_type_name(nt)) for nt in notification_types]
+            return ['{}_{}'.format(model_key, _notification_type_name(nt)) for nt in notification_types]
         elif model_type == ModelType.TEAM:
-            return ['team_{}_{}'.format(model_key, _notification_type_name(nt)) for nt in notification_types]
+            return ['{}_{}'.format(model_key, _notification_type_name(nt)) for nt in notification_types]
         elif model_type == ModelType.MATCH:
-            return ['match_{}_{}'.format(model_key, _notification_type_name(nt)) for nt in notification_types]
+            return ['{}_{}'.format(model_key, _notification_type_name(nt)) for nt in notification_types]
         else:
             return []
 
--- a/tbans/models/subscriptions/subscription_topic.py
+++ b/tbans/models/subscriptions/subscription_topic.py
@@ -1,0 +1,262 @@
+# Note: Be VERY careful modifying these.
+# Modifications to these will probably mean we'll have to run some kind of migration.
+
+
+class SubscriptionTopic:
+    """
+    Helper methods for getting topics for notification types.
+    """
+    @staticmethod
+    def user_subscription_topics(user_id):
+        """
+        Get all topics for a user_id and all all Subscription objects for a user_id.
+        This includes base topics, user topics, and subscription topics.
+
+        Args:
+            user_id (string): User ID to get user topics/Subscriptions for.
+
+        Returns:
+            list (string): List of valid topics for a user.
+        """
+        from tbans.utils.validation_utils import validate_is_string
+        validate_is_string(user_id=user_id)
+
+        from models.subscription import Subscription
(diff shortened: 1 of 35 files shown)


tbans/workers/subscription_worker.py
```diff
@@ -0,0 +1,72 @@
...
+        # If we don't have any tasks to handle - that's fine.
+        if not tasks:
+            return
+
+        # Purge all but one task - we'll re-queue this single task if we fail, as opposed to re-queueing all tasks.
+        task = tasks[0]
+        tasks.remove(task)
+        subscription_queue.delete_tasks(tasks)
+
+        # Wrap everything in a try/catch - if we fail we need to un-lease our tasks
```
ZachOrr: Confirm this logic seems right please

tbans/workers/subscription_worker.py
```diff
@@ -0,0 +1,72 @@
+import webapp2
+import logging
+
+
+class SubscriptionWorkerHandler(webapp2.RequestHandler):
+
+    @staticmethod
+    def task_queue():
+        from google.appengine.api import taskqueue
```
gregmarra: This just in time importing is a different style than we've used elsewhere. The advantage is less overhead if we never call this method?

ZachOrr: Yeah - it probably matters less for things outside of the TBANSService class. The idea is that - classes that need to call out to TBANS shouldn't need to import anything else. So doing the inline imports makes this possible. I'm just doing the inline imports everywhere (even when it's not necessarily needed because we're not exposing those classes) so it's all similar.

For example, right now when we're spinning up the client API, we have to pull in GDCV, since we just do huge imports at the top of the file. You shouldn't *need* to pull in GDCV to hit the test endpoint in the client API, but you do.

gregmarra: That makes sense. It'd be nice if Python did its own lazy importing to avoid this.

tbans/models/subscriptions/subscription_topic.py
```diff
@@ -0,0 +1,262 @@
...
+        user_notifications = [NotificationType.UPDATE_FAVORITES, NotificationType.UPDATE_SUBSCRIPTION]
+        return ['{}_{}'.format(user_id, _notification_type_name(nt)) for nt in user_notifications]
+
+    @staticmethod
+    def _subscription_topics(subscriptions):
+        """
+        Get all valid topics for all Subscription objects.
+
+        Args:
+            subscriptoins (list, Subscription): List of Subscriptions to get topics for.
```
gregmarra: typo

ZachOrr:
Hold on this one - making some changes to try to use firebase-admin

PR #2436, opened 2019-03-07
Add 2019 livescore display

andrewda:
## Description
Adds the GameDay livescore display for 2019.

## Screenshots
image

image

## Types of changes


fangeugene:
Is that `0` supposed to be the time? It should be in the bar above the scores if it is...

Also suggestion: put cargo on top of hatch, make them have the same diameter, and make the hatch stroke thicker (see https://github.com/the-blue-alliance/the-blue-alliance/blob/master/templates_jinja2/match_partials/match_breakdown/match_breakdown_2019.html#L7)

andrewda:
I was just using bad fake data 😛 Here's what it looks like with match time/match number (and cargo on top):

image

PR #2438, opened 2019-03-07
Defer Stackdriver Tracing

fangeugene:
Puts the API call to StackDriver tracing in a deferred task to not slow down the main response.

## Types of changes


fangeugene:
Before/After (at around  9:20)

image

PR #2441, opened 2019-03-09
Separate service for internal mobile API

fangeugene:
- Duplicates the current internal mobile API at `/mobileapi/*` under a separate service.
- Adds a `test` endpoint for easier testing.

## Motivation and Context
We are trying to move everything to its own service for better separation of concerns for code and ops. This allows us to test & use the existing internal mobile API with both the old and new service until everything is migrated to point to the new one.

## How Has This Been Tested?
Visiting `/_ah/api/tbaMobile/v9/test` and `/mobileapi/tbaMobile/v9/test` both work in local dev.

## Types of changes

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -77,7 +77,7 @@ handlers:
 - url: /api/.*
   script: api_main.app
 - url: /_ah/api/.* # Endpoints for internal mobile API
-  script: mobile_main.app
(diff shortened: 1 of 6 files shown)


ZachOrr:
This LGTM - moving all of the existing models seems like a bigger arc of work. But this is good standalone.

PR #2456, opened 2019-03-13
Add helper method to log/return 400s in TBANS

ZachOrr:
The motivation behind this is if we're sending invalid request to TBANS, the logs look all 200. In the case we start sending invalid requests (maybe someone changes the helper methods), the logs will at least start showing up as red.

### DIFF
--- a/tbans/tbans_service.py
+++ b/tbans/tbans_service.py
@@ -35,6 +35,12 @@ class TBANSService(remote.Service):
 
         return app_identity.get_application_id() == incoming_app_id
 
+    def _application_error(self, message):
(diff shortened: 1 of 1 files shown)


tbans/tbans_service.py
```diff
@@ -35,14 +35,20 @@ def _authenticated(self):
 
         return app_identity.get_application_id() == incoming_app_id
 
+    def _application_error(self, message):
+        """ Helper method to log and return a 400 TBANSResponse """
+        # TODO: Monitor these
```
phil-lopreiato: lol

PR #2472, opened 2019-03-20
Modify "Add Video" buttons

jaredhasenklein:
Updates the URL tied to three different buttons on the site.

## Description

- Changes the </img> button to link to `/add-data` on both `index_buildseason` and `index_competitionseason`.
- Changes the </img> button on `event` pages to go to the same URL as the </img> button on top of the `event` page.

## Motivation and Context
Currently, the Add button on the homepage just links to the Contact page. However, once on the contact page, there is a section about adding data (including match videos) which links to the Add Data page. So, it seems more appropriate to link there. As of right now, there is no generic page to add match video for an arbitrary event -- if leaving the `match_key` or `event_key` parameter blank at `/suggest/event/video?match_key=xxx` or `/suggest/event/video?event_key=xxx`, the user is redirected back to `index`. I think since there is no generic "Add Video" page, the Add Data page is the closest thing we have.

Also, currently on an event page there are two CTAs for adding video. There is a green button on the top which links to the video suggestion page for the event and there is a plain button on the bottom which links to the contact page. I'm not sure why there are two distinctive buttons to begin with but this change will make them both link to the same place. Perhaps we want to make them more uniform or remove one altogether but leaving that open to see if anyone has context on why we have two in the first place.

## Types of changes

### DIFF
--- a/templates/index_buildseason.html
+++ b/templates/index_buildseason.html
@@ -50,7 +50,7 @@
         {% endwith %}
         <div>
           <a class="btn btn-default" href="/webcasts"><span class="glyphicon glyphicon-info-sign"></span> Add Webcasts</a>
-          <a class="btn btn-default" href="/contact"><span class="glyphicon glyphicon-upload"></span> Add YouTube Videos</a>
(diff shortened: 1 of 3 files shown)


fangeugene:
Oof, I can't believe these weren't already linking to `add-data`

PR #2476, opened 2019-03-22
Make sure webhook url, name, and secret are non-None

### LINKED ISSUE #2474: Add web-side validation for non-empty webhook information
url/secret both shouldn't be empty when adding a webhook - we should add some validation to the website to make sure this is true.

ZachOrr:
## Description

This change to the webhook controller ensures that name, url, and secret are all non-None for webhooks. Closes #2474

## Motivation and Context

TBANS was assuming webhook secrets were non-None, which doesn't appear to be enforced via during webhook creation. We discussed via Slack generating secrets for users that don't specify secrets so they have a fallback for doing checksums. The next step is to migrate webhook users that don't have secrets set right now and add the validation back in to TBANS.

This also surfaces secrets in the mobile clients UI, since we don't show it right now. There's a lot of diffs, but mostly it's fixing formatting in the HTML/un-nesting the returns in the webhook controller.

## How Has This Been Tested?

Tested in dev

Screen Shot 2019-03-22 at 12 57 49 PM
Screen Shot 2019-03-22 at 12 57 39 PM
Screen Shot 2019-03-22 at 1 02 34 PM
Screen Shot 2019-03-22 at 1 13 47 PM

## Types of changes

### DIFF
--- a/controllers/base_controller.py
+++ b/controllers/base_controller.py
@@ -260,3 +260,13 @@ class LoggedInHandler(webapp2.RequestHandler):
             TeamAdminAccess.expiration > now).fetch()
         if not existing_access:
             return self.abort(403)
+
+    def _require_user(self):
+        current_user_account_id = self.user_bundle.account.key.id()
+
+        target_account_id = self.request.get('account_id')
+        if not target_account_id:
+            return self.abort(403)
+
+        if target_account_id != current_user_account_id:
+            return self.abort(403)
--- a/controllers/base_controller.py
+++ b/controllers/base_controller.py
@@ -261,7 +261,7 @@ class LoggedInHandler(webapp2.RequestHandler):
         if not existing_access:
             return self.abort(403)
 
-    def _require_user(self):
+    def _require_request_user_is_bundle_user(self):
         current_user_account_id = self.user_bundle.account.key.id()
 
         target_account_id = self.request.get('account_id')
--- a/controllers/webhook_controller.py
+++ b/controllers/webhook_controller.py
@@ -24,13 +24,13 @@ class WebhookAdd(LoggedInHandler):
         self._require_request_user_is_bundle_user()
 
         # Name and URL must be non-None
-        url = self.request.get('url')
-        name = self.request.get('name')
+        url = self.request.get('url', None)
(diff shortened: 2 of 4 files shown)


controllers/base_controller.py
```diff
@@ -260,3 +260,13 @@ def _require_team_admin_access(self, team_number, redirect_url=None):
             TeamAdminAccess.expiration > now).fetch()
         if not existing_access:
             return self.abort(403)
+
+    def _require_user(self):
```
ZachOrr: We do this check in `UserNotificationBroadcast` as well, but that's the only other place. I'll probably migrate that controller as well. It's also pretty poorly named since it only manages pings.

controllers/webhook_controller.py
```diff
@@ -16,56 +14,69 @@ class WebhookAdd(LoggedInHandler):
...
+
         path = os.path.join(os.path.dirname(__file__), '../templates/webhook_add.html')
         self.response.out.write(template.render(path, self.template_values))
 
     def post(self):
         self._require_registration()
+        self._require_user()
+
+        # Name and URL must be non-None
+        url = self.request.get('url')
```
ZachOrr: Should we have get defaults in case these aren't supplied for some reason, or should we just let them throw errors?

controllers/base_controller.py
gregmarra: It's not 100% obvious from the name what this function is intended to do.

controllers/webhook_controller.py
gregmarra: If you set the default to None, you'll redirect to the user friendlier error state.

templates_jinja2/account_overview.html
```diff
@@ -287,7 +287,7 @@ <h2>Connected Devices</h2>
         <tr>
           <th>Type</th>
           <th>Name</th>
-          <th>ID</td>
+          <th>ID</th>
```
gregmarra: yikes!

gregmarra:
I'm not an expert on the webhooks code, but this looks generally correct to me.

ZachOrr:
I think this is good to go

PR #2493, opened 2019-04-13
Fix advanced insights chart ranges to correctly produce upper bound.

### LINKED ISSUE #2491: [advanced insights] graph cut off
https://www.facebook.com/groups/moardata/permalink/2293302634023352/

newtonwolfe:
Fixes #2491.

### DIFF
--- a/templates_jinja2/event_insights.html
+++ b/templates_jinja2/event_insights.html
@@ -196,10 +196,10 @@
         minQualScore = Math.min(minQualScore,
                                 predictions[key]['blue_mean']
                                 - 2*Math.sqrt(predictions[key]['blue_var']));
-        maxQualScore = Math.min(maxQualScore,
+        maxQualScore = Math.max(maxQualScore,
                                 predictions[key]['red_mean']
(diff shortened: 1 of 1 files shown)


gregmarra:
Really looks like that makes more sense! 

@newtonwolfe , did you test this somehow?

@fangeugene  I'm going to let you review this, because you wrote advanced insights.

newtonwolfe:
@gregmarra Tested on 2019incmp on my local TBA container last night, and it worked then. 
Unfortunately there aren't currently any events where the logic gets hit, since it requires a specific set up of predicted but unplayed matches.^1 If there's a tool to pull in part of an event (or modify / mock up an event) I'm happy to build an example, but I don't see an easy way to do that.

^1: We need the maximum 95th percentile prediction to be in a higher "bucket" of 50 points than the max actual score. This happened a lot in 2017 and 2018 but with the smaller scores this year, it only really happens when no or very few matches have been played, or when there aren't any matches with more than 100 points.

fangeugene:
Nice catch!

PR #2504, opened 2019-04-23
Fix #2379.

### LINKED ISSUE #2379: Possible XSS through JSON API
Through the JSON API there is embedded HTML which is rendered directly on the page. This could possibly allow for unintentional XSS through embedded script tags. **NOTE:** this does appear to impact the website, but may impact people who use the API.

## Expected Behavior
JSON response should not contain HTML tags.

## Current Behavior
JSON response returns HTML tags which are directly rendered on TBA.
`"overall_status_str": "Team 254 was <b>Rank 3/42</b> with a record of <b>6-4-0</b> in quals, competed in the playoffs as the <b>1st Pick</b> of <b>Alliance 1</b>, and was <b>eliminated in the Finals</b> with a playoff record of <b>5-2-0</b>."`

## Possible Solution
Do text formatting client-side without HTML tags. You could use custom tags that are replaced with valid HTML in JavaScript. I.E: `<b>` maps to `b_s` and `</b>` maps to `b_e`

## Steps to Reproduce (for bugs)
1. Send get request to `https://www.thebluealliance.com/api/v3/team/frc254/event/2018cc/status`
2. See `alliance_status_str` for example

## Context
I was trying to get team status data for an event from the website and noticed this.

## Your Environment

* Version used: v3
* Environment name and version (e.g. Chrome 39, node.js 5.4): Chrome 71.0.3578.98, Insomnia 6.3.2
* Operating System and version (desktop or mobile): Windows 10 Desktop

brennonbrimhall:
Fixes #2379.

## Description
Makes formatting optional; calls to the API remove HTML tags.

## Motivation and Context
First issue/PR.  See #2379 for context.

## How Has This Been Tested?
I've made calls to myself in my local environment, and run paver test.

## Types of changes

### DIFF
--- a/api/apiv3/api_event_controller.py
+++ b/api/apiv3/api_event_controller.py
@@ -147,9 +147,9 @@ class ApiEventTeamsStatusesController(ApiBaseController):
             team_key = event_team.team.id()
             if status:
                 status.update({
-                    'alliance_status_str': EventTeamStatusHelper.generate_team_at_event_alliance_status_string(team_key, status),
-                    'playoff_status_str': EventTeamStatusHelper.generate_team_at_event_playoff_status_string(team_key, status),
-                    'overall_status_str': EventTeamStatusHelper.generate_team_at_event_status_string(team_key, status),
+                    'alliance_status_str': EventTeamStatusHelper.generate_team_at_event_alliance_status_string(team_key, status, formatting=False),
+                    'playoff_status_str': EventTeamStatusHelper.generate_team_at_event_playoff_status_string(team_key, status, formatting=False),
+                    'overall_status_str': EventTeamStatusHelper.generate_team_at_event_status_string(team_key, status, formatting=False),
                 })
             statuses[team_key] = status
         return json.dumps(statuses, ensure_ascii=True, indent=2, sort_keys=True)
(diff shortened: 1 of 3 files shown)


brennonbrimhall:
Hold off on this for a second.

brennonbrimhall:
OK, we should be good now.

fangeugene:
is #2379 actually an issue? Do we really want to fix it this way?

brennonbrimhall:
Is it an XSS vector against TBA?  No.  Is it an XSS vector against a client, not really.  However, I personally don't think we should be sending out HTML in the response.

ZachOrr:
#2379 is not a real issue. Having HTML tags in our responses is a weird part of our API. I believe we should just remove them altogether. I believe the Android app depends on these tags for formatting in the Team@Event view. We should make sure we're not breaking clients with this change.

brennonbrimhall:
> I believe we should just remove them altogether

Remove what, exactly?

brennonbrimhall:
One other thing worth mentioning - I'm not seeing anywhere where we specify the formatting of the status strings.  If we do want to return HTML here, it might be worth updating that in the documentation.

ZachOrr:
My comment was referring to the HTML tags in our JSON responses.

PR #2512, opened 2019-05-01
Bootstrap dqs.

brennonbrimhall:
When a local instance is created for development purposes, make sure to
pull down disqualification status too.

## Description

Minor addition to bootstrap.py.

## Motivation and Context

See above.

## How Has This Been Tested?

Tested in my local instance with `2019cur`.

N/A

## Types of changes


ZachOrr:
Can we add `score` while we're at it?

fangeugene:
@ZachOrr I think it's already included?

ZachOrr:
I guess I didn't see it in the script - it seems hard to imagine we'd go this long without putting score on match objects when bootstrapping (and I can't test right now) so I'll defer

PR #2513, opened 2019-05-01
Fix #2507.

### LINKED ISSUE #2507: Return all teams DQ'd during for elim DQs
## Expected Behavior

We should return that all teams were DQ'd if a team receives a DQ during an elimination/playoff match.

## Current Behavior

Only the alliance captain is DQ'd. Keep in mind - this behavior is true for 2019 but I'm not entirely sure how we should treat other years.

See https://www.thebluealliance.com/match/2019cur_qf4m1

brennonbrimhall:
Correctly indicate disqualification status for eliminations.

## Description

See diff.  It's pretty minor.

## Motivation and Context

We currently do not correctly indicate that elimination alliances are
disqualified together.  We currently only mark the middle team as
disqualified.

See #2507.

## Types of changes

### DIFF
--- a/database/match_query.py
+++ b/database/match_query.py
@@ -7,7 +7,7 @@ from models.match import Match
 
 
 class MatchQuery(DatabaseQuery):
-    CACHE_VERSION = 1
+    CACHE_VERSION = 2
     CACHE_KEY_FORMAT = 'match_{}'  # (match_key)
     DICT_CONVERTER = MatchConverter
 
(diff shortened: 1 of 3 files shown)


brennonbrimhall:
Looks like some of the tests were expecting 1 team in dqs for elims matches.  This obviously breaks them.

fangeugene:
This looks great! You should roll all the `CACHE_VERSION`s here though: https://github.com/the-blue-alliance/the-blue-alliance/blob/master/database/match_query.py

PR #2514, opened 2019-05-04
Migrate TBANS to use Firebase Admin SDK

ZachOrr:
This moves from using our own internal models/code to send to FCM to using the Firebase Admin Python SDK.

- `ssl` library and `GAE_USE_SOCKETS_HTTPLIB` are needed for authentication for the Firebase lib
  - https://google-auth.readthedocs.io/en/latest/user-guide.html#the-app-engine-standard-environment
- Removed a lot of the README since it feels a little too specific. I have plans to replace with something more helpful. The design doc works for as documentation for now
- `PlatformPayloadPriority` -> `PlatformPriority`, added some helper methods for validation/DRYing code
- `PlatformPayloadType` -> `PlatformType`, added some helper methods for validation/DRYing code
- `PlatformPayload` -> `PlatformConfig`, simplified code since we no longer have to create JSON blobs
- Removed FCM and IID error constants since we get them in the Firebase library
- Updated `Notification` to match the new design in the design doc
- Removed the unnecessary notification models like `NotificationPayload`, `Payload`, since we get this functionality in the Firebase lib
- Updated ping and verification notification
- Removed currently unused notifications (myTBA notifications)
- Simplified `FCMRequest`, removed response objects
- Removed all Instance ID code (subscription models, request/response objects) since we won't be hitting IID for now
- Renamed `NotificationRequest` -> `Request`, updated to reflect new Request design
- Simplified sending in `WebhookRequest`
- Updated `TBANSService` to consume new notification/request models
- Added initialization of the Firebase library when importing TBANS - could be moved
- Removed auth utils since we're not longer manually authenticating with Firebase
- Updated tests


tbans/__init__.py
```diff
@@ -0,0 +1,7 @@
+import tba_config
+
+# Don't init Firebase in tests
+if not tba_config.TESTING:
+    import firebase_admin
+    # Authenticate with Firebase
+    firebase_admin.initialize_app()
```
ZachOrr: This probably needs to move

tbans/tbans_service.py
```diff
@@ -12,20 +12,19 @@
...
+            self._firebase_app = firebase_admin.initialize_app(name='tbans')
 
     def _validate_authentication(self):
         import tba_config
         # Allow all requests in debug mode
         if tba_config.DEBUG:
             return
 
-        # Ignore auth check during tests
-        if self.testing:
```
ZachOrr: `tba_config.DEBUG` catches this case doh

tests/tbans_tests/requests/test_webhook_request.py
```diff
@@ -60,23 +60,30 @@ def test_str(self):
...
-        message = WebhookRequest(MockNotification(webhook_payload={'data': 'value'}), 'https://www.thebluealliance.com/', 'secret')
+        message = WebhookRequest(MockNotification(webhook_message_data={'data': 'value'}), 'https://www.thebluealliance.com/', 'secret')
         response = message.send()
-        self.assertEqual(response.status_code, 200)
-        self.assertIsNone(response.content)
+        result = response.get_result()
+        self.assertEqual(result.status_code, 200)
+        self.assertEqual(result.content, 'Some content here')
+
+    def test_send_fail(self):
```
ZachOrr: It turns out we weren't hitting our validations in `_stub_webhook_request` in our `test_send` since the changes introduced in https://github.com/the-blue-alliance/the-blue-alliance/pull/2422 since we were never waiting to get the content. This adds a test that will fail if we don't hit our validations, which makes sure were hitting the validations in our other tests.

ZachOrr:
This looks good testing end-to-end with a device and with a webhook

IMG_0313
Screen Shot 2019-05-04 at 1 06 05 PM

tbans/requests/fcm_request.py
```diff
@@ -0,0 +1,99 @@
...
+        Returns:
+            string: ID for the sent message.
+        """
+        from firebase_admin import messaging
+        return messaging.send(self._fcm_message(), app=self._app)
+
+    def _fcm_message(self):
+        platform_config = self.notification.platform_config
+
+        from tbans.consts.fcm.platform_type import PlatformType
```
ZachOrr: We could move this in to `Notification` - just namespacing these methods is important. We can't have properties called `android_config` for example, since that's what subclasses override. We could have `_android_config` which will do this business of checking for an override, falling back to the `platform_config`, etc. Using notification configurations is a FCM request specific thing though, so part of me feels like it's okay here.

tbans/consts/fcm/platform_priority.py
```diff
@@ -0,0 +1,53 @@
+class PlatformPriority:
```
phil-lopreiato: why can't things be consistent ever :(

tests/tbans_tests/requests/test_webhook_request.py
phil-lopreiato: <3

PR #2516, opened 2019-05-11
Unpack num_played in 2015 advancement template

ZachOrr:
Phil's change in https://github.com/the-blue-alliance/the-blue-alliance/commit/a04fe4ad9c687ced4ec125a7f8f9b7dc4d0ae147 added a `num_played` field to the results of `generatePlayoffAdvancement2015`, but we weren't unpacking that value in the templates. This updates the templates to unpack that value and fix the 500s for 2015 events.


phil-lopreiato:
Sorry I broke it :(

PR #2517, opened 2019-05-11
Fix reference to MatchHelper.generateBracket to point to PlayoffAdvan…

ZachOrr:
This got moved/renamed in https://github.com/the-blue-alliance/the-blue-alliance/pull/2510/files but didn't get updated 🙊


phil-lopreiato:
<3

PR #2519, opened 2019-05-16
Update spec to OpenAPI

tweirtx:
## Description

Converted the api_v3.json from Swagger (2.0) to OpenAPI (3.0).
## Motivation and Context

OpenAPI is making leaps and bounds ahead of Swagger, including with Codegen. The prior spec was never parsed by the old codegen correctly due to issues that the upgrade has fixed.
## How Has This Been Tested?

Passed automated tests, client library generation, and autogenerated API documentation checks.
## Types of changes


bdaroz:
I know when I did the Swagger 2.0 version there was _a lot_ of duplicated descriptions I refactored with `#ref` tags to cut down the file considerably. Did support for many of those get dropped in Swagger 3.0?

tweirtx:
Unsure, this was autogenerated so it's entirely possible that the generator doesn't do refs even though OpenAPI can

tweirtx:
At a glance, it doesn't look possible to do the last 2. However, I will do the first one once I track down where I've made a circular reference that's breaking everything.

bdaroz:
Disappointing about the last one, it was supposedly going to be one of the big pushes for v3 at the time we did v2.

tweirtx:
It's entirely possible I'm just missing it, I'm trying to make it match the v2 one first before I try to add anything new.

tweirtx:
So the swagger spec linter in CI keeps failing even though everything else I have is saying it's fine. I'm gonna replace the linter with something else

tweirtx:
After further research, anyOf support is a thing in the new standards. I would like to potentially revisit at a later date, however.

bdaroz:
My thoughts on this come down to breaking changes in generated code. If the migration from the v2 to v3 spec will cause a breaking change in generated code, then I'd strongly suggest we do the `anyOf` migration at the same time as I'm sure it would be a breaking change when we do finally implement it.

If, however, we can take spec v3 code and drop-in without refactoring in place of v2 code for the major platforms (Java, Python2 & 3, and PHP) then I would be ok with deferring `anyOf`.

tweirtx:
After learning that the spec changes do not automatically replace server code, my testing for breaking changes is going to be completed later today hopefully. I'm going to write a test suite that works on the code generated by the current spec, then run the code generated by the new spec against that test suite and see if it passes.

tweirtx:
After further review, there are not any breaking changes made in the current state of the PR

bdaroz:
Ok, last change is I'd like to see the version number and changelog in the spec file updated to reflect the change, given that it seems you need different packages now to build code from the API (at least the `requirements.txt` has been changed).

Also, do the Swagger generated doc URL still work with the new spec files, or does that need a version bump as well?

tweirtx:
I'll bump the version. As for the docs, I've tested on my own App Engine instance and it doesn't seem to have any problems. You can check it out for yourself on [my instance](https://tba-firebase-test-tweirtx.appspot.com/apidocs/v3) if you would like.

fletch3555:
Do we honestly NEED to bump the version? None of this had been merged it yet and the existing version was already bumped to 3.0 for this change as it is.  Unless I'm misunderstanding

bdaroz:
@fletch3555 I was referring the the version number in the spec file and it's changelog, namely the bump from `3.4.1` to `3.5`

fletch3555:
Ah fair point, I looked quick and missed that one.  Good by me!

PR #2538, opened 2019-06-01
Add button to request API write tokens

jaredhasenklein:
Adds a button on `/account` to request API tokens.

## Description
Adds a "Request Tokens" button next to the API Write Key area of the account page. 
</img>
The button matches the one found [here](https://www.thebluealliance.com/request/apiwrite) and is in a similar style to the one above on read keys.

## Motivation and Context
I went to my Account page, remebering that there were API keys there, trying to request write keys for an offseason. I realized on that page, I can request read keys but not write keys. I think there should be a button or link that directs to `/request/apiwrite` from there because I had to follow three links to get to the page I intended. 

#### New button
image
#### Current view
image

#### Read button - for reference 
image

## Types of changes

### DIFF
--- a/templates_jinja2/account_overview.html
+++ b/templates_jinja2/account_overview.html
@@ -198,7 +198,7 @@
   <div class="row">
     <div class="col-md-10 col-md-offset-1">
       <h2>Write API Keys</h2>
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
We should maybe also add some info copy about why one would want to do this, so we don't get a bunch of requests from the wrong people

jaredhasenklein:
What about changing the existing line `These keys allow you to import data for your offseason event to TBA.` to `These keys grant volunteers planning events and working at the scoring table the ability to import offseason event data to TBA.`

PR #2539, opened 2019-06-09
Update valid match breakdowns for 2019

lethosor:
## Description

This allows 2019 match breakdowns to be uploaded via the write API. 
## Motivation and Context
Currently, any attempts to do this result in this error:
```
Valid score breakdowns for 2019 are: set([])
```
I would like for this to be fixed before one of my upcoming events, and I'm pretty sure this is the only place that needs to change. (I'm not sure if the rankings endpoint needs to be updated or not, though.)

## How Has This Been Tested?
I haven't gotten Docker to work locally so it hasn't been. But I did run this file through the Python interpreter to make sure there aren't any syntax errors (there aren't). The field list is taken from https://www.thebluealliance.com/api/v3/match/2019micmp1_qm1, which I believe is complete.

## Types of changes


lethosor:
I hate to bump PRs, but is it possible to get this deployed by Friday? I'd like to use this for Big Bang (although I can live without it and upload any data I have later). If there's anything else I need to do here, by all means let me know.

lethosor:
Thanks!

PR #2542, opened 2019-06-16
Fix task enqueue URLs for insights

tervay:
The buttons on the admin console for launching tasks related to generating insights were broken.

## Description
I found URLs that work in `cron.yaml`, so I just replaced the 404ing URLs with the corresponding ones from there.

## Motivation and Context
Makes dev lives easier

## How Has This Been Tested?
Manually tested, you should probably test it on your own as well

## Types of changes

### DIFF
--- a/templates/admin/tasks.html
+++ b/templates/admin/tasks.html
@@ -235,7 +235,7 @@
         </form>
         <script>
         $("#enqueue_award_insights_year_submit").click(function() {
-            window.location = "/tasks/math/enqueue/insights/awards/" + $("#enqueue_award_insights_year").val();
+            window.location = "/backend-tasks-b2/math/enqueue/insights/awards/" + $("#enqueue_award_insights_year").val();
         });
         </script>
     </div>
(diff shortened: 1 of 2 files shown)


templates/admin/tasks.html
```diff
@@ -251,18 +251,18 @@ <h3>Calculate Year Insights</h3>
...
 </div>
 <h3>Calculate Overall Insights</h3>
 <div class="row">
     <div class="col-sm-6">
-        <a class="btn btn-warning" href="/tasks/math/enqueue/overallinsights/awards">Enqueue Overall Award Insights</a>
+        <a class="btn btn-warning" href="/backend-tasks-b2/math/enqueue/overallinsights/awards">Enqueue Overall Award Insights</a>
     </div>
     <div class="col-sm-6">
-        <a class="btn btn-warning" href="/tasks/math/enqueue/overallinsights/matches">Enqueue Overall Match Insights</a>
+        <a class="btn btn-warning" href="/backend-tasks-b2.yaml/math/enqueue/overallinsights/matches">Enqueue Overall Match Insights</a>
```
fletch3555: Was this `.yaml` intentional?  Seems out of place

PR #2543, opened 2019-07-01
Added "direct_link" to webcasts type and Removed "rookie_year" as requirement for teams

xNovax:
Fixes issue 
1) TBA API where get_event("2019micmp") returns with a webcast of type "direct_link" which is not an allowed type webcasts

2) TBA API where get_team("frc7") creates error because their rookie_year is None

## Description
Adds "direct_link" to the allowed types for webcasts
Removes "rookie_year" requirement from teams since many teams do not have a rookie year EG : frc326, frc7

### Webcasts
event_response_webcast
error_using_api

### Rookie Year
rookie_year_error

## Types of changes

### DIFF
--- a/static/swagger/api_v3.json
+++ b/static/swagger/api_v3.json
@@ -6626,7 +6626,8 @@
               "iframe",
               "html5",
               "rtmp",
-              "livestream"
+              "livestream",
+              "direct_link"
             ]
           },
           "channel": {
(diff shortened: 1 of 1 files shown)


ZachOrr:
fwiw this isn't a breaking change, yeah? Just an add change to the schema

xNovax:
Updated the type of change.

phil-lopreiato:
Thanks for the fix!

Issue #2469, opened 2019-03-19
Request for OnShape to be a supported format for CAD


smitty1707:
## Expected Behavior

Many teams have been migrating to online only CAD packages for their robot designs. It would be great if TBA supported these new types of online CAD tools as valid entries for the robot CAD field. (OnShape being the most widely used -- to my knowledge)

## Current Behavior

Currently, attempting to add an onshape line to the media field results in an error for an unsupported format.

## Steps to Reproduce (for bugs)

N/A

## Your Environment

N/A

jaredhasenklein:
Also would like to note that there are several FIRST Alumni who work at OnShape and have been working with FRC teams to tailor the platform towards their needs. If contacting these people would be helpful, anyone can feel free to ask me on Slack.

bjmboy:
please

phil-lopreiato:
This is a good place for a new contributor to make some impact - I'd be happy to show anyone how to implement this, if they're interesting in taking a stab at it :)

jaredhasenklein:
Don't currently have the time to try and build this but I've been doing some detective work for anyone who wants to build this.

To get a screenshot/thumbnail:

If your public document links is 

1. `https://cad.onshape.com/documents/58ae7f2f1173be0f8386d6ab/w/1e7ab48f638b05adf8bcdc25/e/4d75d5bfbfafd244bc464f02`, you need to extract the part after `/documents/` and before `/e/.
2. So, you'll get something like **`58ae7f2f1173be0f8386d6ab/w/1e7ab48f638b05adf8bcdc25`**.
3. Then, you need to plug it in to the URL `https://cad.onshape.com/api/thumbnails/d/[document-key]/s/300x300` so you'll get something like **`https://cad.onshape.com/api/thumbnails/d/58ae7f2f1173be0f8386d6ab/w/1e7ab48f638b05adf8bcdc25/s/300x300`** for the example document.
4. The part after `/s/` is the size. Options include `300x300`, `70x40`, `300x170`, and `600x340`.

Bo887:
@phil-lopreiato Hi, I'm interested in working on this (if it hasn't been started/completed already). Could you give me a few starting pointers?

phil-lopreiato:
@Bo887 sorry about the delay here; I'd be happy to give some pointers. This should go something like this:

 - add onshape as a new media type and allow it to be linked to the robot model in `consts/media_type.py`
[...]

PR #2547, opened 2019-07-09
Add onshape CAD support

Bo887:
Adds support for users to upload onshape CAD models alongside Grabcad ones.

## Description
Adds Onshape as an extra media type, and adds support for Onshape document urls into the CAD submission field. To fetch information about the model (i.e. name, creation date, etc), a request is sent to the Onshape api. Overall, this is based on the implementation of the Grabcad CAD models.

One limitation (as of now) is that the public api is only able to fetch data about Onshape models that are accessible via link (i.e. models that do not require login to an onshape account). From my testing, to fetch information using the Onshape api about models that are public but require a user to login, a set of api keys will need to be generated for the Onshape api. As of now, to upload an Onshape document, the document will have to be made publicly accessible by link.

## Motivation and Context
Addresses #2469

## How Has This Been Tested?
I have only tested this by running TBA locally and uploading/deleting various Grabcad/Onshape cad models to team pages through both the team submission and admin panel (I created some simple Onshape models for this, the specific ones are [here](https://cad.onshape.com/documents/12970ee341c0009679e6c1aa/w/7f5911670fbe5c55d35c2159/e/64741577e7314d669e10599f) and [here](https://cad.onshape.com/documents/cfb3e7b1e271d7a73b4fb21e/w/d7248be0d0967489a13c9fd1/e/c5d077c56af2055f77276b63)). 
Also, I checked to make sure the images/titles rendered correctly on the team page, admin page, and team submission page as well as if the image/cad links worked correctly.

## Types of changes


Bo887:
@phil-lopreiato Here is my current progress. Can you let me know how it looks so far and if there are any things to fix/improve?

templates_jinja2/suggestions/suggest_team_media.html
```diff
@@ -129,15 +129,20 @@ <h4>Existing Videos</h4>
           </div>
           {% endif %}
 
-          {% if medias_by_slugname.grabcad %}
+          {% if medias_by_slugname.grabcad or medias_by_slugname.onshape %}
```
phil-lopreiato: Only think is we should add a note that onshape is now a supported format in the copy above in this file

Bo887: Updated. Does this look good?
submission-description

PR #2553, opened 2019-07-31
Fix issues with importing 2019 FMS schedule reports

### LINKED ISSUE #2548: 2019 FMS schedule reports dropped "Match" column, breaks event wizard schedule import
Specifically, this check fails because the "Match" column is no longer present:
https://github.com/the-blue-alliance/the-blue-alliance/blob/22daf5ee5bd45567862350951918568bcf5f6adc/static/javascript/tba_js/eventwizard_apiwrite.js#L110
A couple FIM offseasons have run into this, but I forgot to report it - we added the column to the spreadsheet manually at first, then I put together a script to import a CSV schedule report.

Fortunately, the "Description" column does contain the number that used to be in the "Match" column in a relatively easy-to-parse format: either `Qualification X` for quals or `(#X)` for playoffs.

This year's FMS can also export schedules as CSV - is that worth supporting in the event wizard as well? It could allow dropping the XLSX parser dependency eventually (although that might not be a good idea this year, in case FMS offseason doesn't support CSV exports or people are still running an older version for some reason).

2019mibb schedules attached for reference:
[MIBB2019_ScheduleReport_all.zip](https://github.com/the-blue-alliance/the-blue-alliance/files/3379656/MIBB2019_ScheduleReport_all.zip)

lethosor:
This adds support for schedule reports exported from 2019 FMS. These reports dropped the "Match" column from last year and added that information to the "Description" column for playoff matches, so attempting to import them would result in a "No matches found in the file" message.

Fixes #2548

Match result reports seem to work as expected without changes.

## How Has This Been Tested?
Tested in a local instance with both qualification and playoff schedule reports from 2019MIBB and 2018MIBB, which were recognized and uploaded successfully. (I'm unsure if some people are still using 2018-style reports, so I have retained support for those as well.)

## Types of changes

### DIFF
--- a/static/javascript/tba_js/eventwizard_apiwrite.js
+++ b/static/javascript/tba_js/eventwizard_apiwrite.js
@@ -107,20 +107,27 @@ $('#schedule_file').change(function(){
             var match = matches[i];
 
             // check for invalid match
-            if(!match['Match'] || !match['Red 1']){
+            if(!match['Description'] || !match['Red 1']){
                 continue;
             }
 
-            var compLevel, setNumber, matchNumber, matchKey;
+            var compLevel, setNumber, matchNumber, rawMatchNumber, matchKey;
             var has_octo = $('input[name="alliance-count-schedule"]:checked').val() == "16";
-            matchNumber = parseInt(match['Match']);
+            if (match['Match']) {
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
Thanks for the fix!

PR #2554, opened 2019-07-31
Fix parsing match times in event wizard schedule import

lethosor:
This fixes an issue where match times submitted in schedule reports via the event wizard would fail to parse, resulting in "?" placeholder scores being displayed instead of the scheduled match time. (`re.match` only looks for matches at the start of a string, and schedule reports contain strings such as "Sat 11:40 AM", so the weekday was causing issues here.) 

(Note that I based this on #2553 to get schedule imports to work at all, so that should probably be merged first. 0615b3c is the only new commit here.)

## How Has This Been Tested?
Tested uploading schedule reports on a local instance

## Screenshots:
Before:
image

After:
image

## Types of changes

### DIFF
--- a/helpers/match_helper.py
+++ b/helpers/match_helper.py
@@ -33,7 +33,7 @@ class MatchHelper(object):
         last_match_time = None
         cur_date = event.end_date + datetime.timedelta(hours=23, minutes=59, seconds=59)  # end_date is specified at midnight of the last day
         for match in matches_reversed:
-            r = re.match(r'(\d+):(\d+) (am|pm)', match.time_string.lower())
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
Mind rebasing this now that #2553 is merged?

lethosor:
Oh, did you rebase/squash #2553? A regular merge should've avoided the need to rebase this. I can take care of it later today.

lethosor:
@phil-lopreiato rebased. (Although I suspect I can work around the issue by just removing weekdays from the schedule, if necessary)

phil-lopreiato:
Thanks!

PR #2556, opened 2019-08-01
Remove Team motto from APIv2 and APIv3 responses

ZachOrr:
## Description
As we've been discussing for a while, this removes `motto` from Team responses in APIv2 and APIv3. This doesn't touch the backing data models/stores or parsers for team mottos.

## Motivation and Context
Second part of https://github.com/the-blue-alliance/the-blue-alliance/pull/2410

## How Has This Been Tested?
#YOLO

## Types of changes

### DIFF
--- a/database/dict_converters/team_converter.py
+++ b/database/dict_converters/team_converter.py
@@ -3,7 +3,7 @@ from database.dict_converters.converter_base import ConverterBase
 
 class TeamConverter(ConverterBase):
     SUBVERSIONS = {  # Increment every time a change to the dict is made
-        3: 4,
+        3: 5,
     }
 
     @classmethod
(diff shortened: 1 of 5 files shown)


fangeugene:
This is technically a breaking change. Setting it to `None` was the non-breaking solution.

ZachOrr:
Yep - I think we've bumped the API version accordingly. We've also communicated this change outwardly in advance.

bdaroz:
Concur with breaking change. Recommend holding until after the 2019 off seasons complete.

ZachOrr:
We can hold off on merging for a while - if we can review in the meantime and make sure things look good that'd be killer.

fangeugene:
I think the point is we don't ever want to merge it. Until APIv4.

bdaroz:
If we keep "breaking" changes to a minimum (removal of a deprecated field being pretty minimal here) and commit after the season (and off season) is over, I don't feel so strongly about waiting for v4. Now if we re-architect all the calls into, say, GraphQL, or something, as a breaking change, then yeah, v4.

PR #2557, opened 2019-08-01
Send myTBA update favorite/subscription messages via FCM HTTPv1 multicast

ZachOrr:
## Description
Adds endpoints in TBANS for sending update favorites/subscriptions notifications to iOS/web clients when devices update their myTBA models.

The weird piece of this PR is moving shared classes between the TBA monolith and TBANS in to a `tba` directory, for namespacing reasons. Python won't let us do `import consts` if there's both a `consts/` and `tbans/consts` directory since it doesn't know how to resolve the namespacing. We saw this same problem when adding directories inside of the `tests` directory. This moves minimal shared classes (constants and some models) in to a `tba` directory for TBANS. The idea for this folder is to keep the dependencies VERY low, and inside of the `tba` folder, and not importing things upward. This gets us slightly closer to my ideal goal of being able to deploy pieces of the monolith when deploying TBANS, as opposed to deploying the entire codebase.

All the consts being moved either depend on nothing or other consts. The only model being moved is the `MobileClient` model.

Note for reviewers - most of the changed files are just moves/import changes

## Motivation and Context
Part of the work to build out TBANS to include support for sending notifications to iOS/web

## How Has This Been Tested?
Tested via tests and end-to-end in a dev instance


tbans/controllers/notification_controller.py
```diff
@@ -0,0 +1,91 @@
...
+            pass
+
+        vertical_type = int(self.request.get('vertical_type'))
+        if vertical_type == VerticalType.FCM:
+            success = NotificationHandler._send_fcm(notification, tokens)
+            # If our FCM send failed, requeue to try again
+            if not success:
+                self.response.set_status(500)
+        elif vertical_type == VerticalType.WEBHOOK:
+            # TODO: Dispatch to webhooks
```
ZachOrr: - [ ] Dispatch notifications to webhooks (this code shouldn't execute for myTBA notifications)

tbans/controllers/notification_controller.py
```diff
@@ -0,0 +1,91 @@
...
+            if response.exception.code in ['mismatched-credential', 'registration-token-not-registered']:
+                logging.warning(log_string)
+                logging.warning('Removing bad token - {}'.format(token))
+                from tba.models.mobile_client import MobileClient
+                MobileClient.delete_for_token(token)
+            else:
+                # Log and have this retry later
+                logging.error(log_string)
+                success = False
+            # TODO: 'invalid-argument' should raise a noisy error, without retrying
```
ZachOrr: - [ ] Figure out if we should or shouldn't handle `invalid-argument` quietly

Steaks are fairly low on iOS since registration happens frequently. Throwing errors would only alert us that there's some problem - they're not necessarily actionable. But having failing push tokens hanging out in the database is probably okay.

tbans/requests/fcm_request.py
```diff
@@ -83,17 +67,13 @@ def _fcm_message(self):
 
         # Add `notification_type` to data payload
         data_payload = self.notification.data_payload if self.notification.data_payload else {}
-        from consts.notification_type import NotificationType
+        from tba.consts.notification_type import NotificationType
         data_payload['notification_type'] = NotificationType.type_names[self.notification.__class__._type()]
 
+        # TODO: Add back in platform payloads
```
ZachOrr: - [ ] Add back in platform payloads

Still going back and forth with Firebase support on this one - absolutely wild

tests/tbans_tests/controllers/test_notification_controller.py
```diff
@@ -0,0 +1,146 @@
...
+    def tearDown(self):
+        self.testbed.deactivate()
+        self._mock_batch_response = None
+
+    def _stub_send_all(self, message, dry_run):
+        self.assertFalse(dry_run)
+        self.assertIsNotNone(message)
+        return self._mock_batch_response
+
+    # def test_notify_user_favorites(self):
```
ZachOrr: - [ ] Enable?

tests/tbans_tests/controllers/test_notification_controller.py
```diff
@@ -0,0 +1,146 @@
+from firebase_admin import messaging
```
ZachOrr: - [ ] Double check we have full coverage

tests/tbans_tests/requests/test_fcm_request.py
```diff
@@ -25,87 +25,70 @@ def setUp(self):
             self._app = firebase_admin.get_app('tbans')
```
ZachOrr: - [ ] Double check we have full coverage

tbans/requests/fcm_request.py
ZachOrr: Finally got this - sending an `apns` object in a request with only a `headers` object will cause the request to fail, for some reason. If we're going to pass an `apns` object, we need to send both a `headers` and a `payload` object, and the `payload` object needs to have an `aps` object setup. This seems like a bug in FCM - but we should look to work around it for now.

tbans/tbans_service.py
```diff
@@ -39,6 +43,10 @@ def _application_error(self, message):
         logging.error(message)
         return TBANSResponse(code=400, message=message)
 
+    @remote.method(response_type=TBANSResponse)
+    def test(self, request):
```
ZachOrr: - [ ] Remove this

tbans/controllers/notification_controller.py
ZachOrr: Going to table this for now, since we don't need this for sending myTBA notifications

ZachOrr:
Closing this - working on converting to using cloud functions, which will pull in this functionality.

PR #2566, opened 2019-08-08
Add APIv3 endpoints for all teams and all events

fangeugene:
## Description
Add `/api/v3/teams/all/(simple|keys)` and `/api/v3/events/all/(simple|keys)` endpoints that return ALL teams and events respectively.

## Motivation and Context
Enables apps to download an index of all teams and events without hitting 40+ endpoints.

## How Has This Been Tested?
Local dev.

## Types of changes

### DIFF
--- a/api/apiv3/api_event_controller.py
+++ b/api/apiv3/api_event_controller.py
@@ -30,7 +30,7 @@ class ApiEventListAllController(ApiBaseController):
 
     def _render(self, model_type=None):
         futures = []
-        for year in xrange(1992, tba_config.MAX_YEAR + 1):
+        for year in xrange(tba_config.MIN_YEAR, tba_config.MAX_YEAR + 1):
             futures.append(EventListQuery(year).fetch_async(dict_version=3, return_updated=True))
 
         events = []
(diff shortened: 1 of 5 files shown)


api/apiv3/api_event_controller.py
```diff
@@ -17,6 +18,33 @@
...
+
+    def _track_call(self, model_type=None):
+        action = 'event/list'
+        if model_type:
+            action += '/{}'.format(model_type)
+        self._track_call_defer(action, 'all')
+
+    def _render(self, model_type=None):
+        futures = []
+        for year in xrange(1992, tba_config.MAX_YEAR + 1):
```
ZachOrr: Do we have a lot of references to 1992 in the app? I know it doesn't change - but it might be worth setting as a `MIN_YEAR` in our config.

fangeugene: Yeah we definitely do. We should start moving to `MIN_YEAR` in config, starting from this PR.

ZachOrr: :+1: I can file a cleanup ticket - seems fairly straightforward

PR #2568, opened 2019-08-08
Convert all 1992 references to using tba_config.MIN_YEAR

### LINKED ISSUE #2567: Replace `1992` references with `tba_config.MIN_YEAR`
Depends on https://github.com/the-blue-alliance/the-blue-alliance/pull/2566/files (unless we pull it out in to a different PR) - we should replace hardcoded references to 1992 with the tba config `MIN_YEAR` constant

ZachOrr:
Closes #2567

### DIFF
--- a/controllers/api/api_event_controller.py
+++ b/controllers/api/api_event_controller.py
@@ -15,6 +15,7 @@ from helpers.model_to_dict import ModelToDict
 
 from models.event import Event
 
+import tba_config
 
 class ApiEventController(ApiBaseController):
     CACHE_KEY_FORMAT = "apiv2_event_controller_{}"  # (event_key)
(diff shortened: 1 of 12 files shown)


controllers/api/api_event_controller.py
```diff
@@ -196,7 +197,7 @@ def _track_call(self, *args, **kw):
         self._track_call_defer('event/list', self.year)
 
     def _render(self, year=None):
-        if self.year < 1992 or self.year > datetime.now().year + 1:
+        if self.year < tba_config.MIN_YEAR or self.year > datetime.now().year + 1:
```
ZachOrr: I'm not going to modify this bit to use `VALID_YEARS` - since it doesn't look at our constant year range, it looks at the current date (is this correct behavior?)

PR #2570, opened 2019-08-17
Insert official/unofficial offseason events properly. Don't insert offseason events during EventListGet

ZachOrr:
This is a two-parter -

First, `FMSAPIEventListParser` has been modified to respect non-official `offseason` events from the API. This will fix issues we've been seeing where events without will have their data overwritten because the event is marked as official TBA-side.

Second, the `OffseasonEventListGet` class and corresponding cron job has been removed and combined in to the `EventListGet` method. In `EventListGet` we'll fetch all events from the FRC Events API and insert non-offseason events automatically. For offseason events, we compare them to existing offseason events in TBA. If we find a match event (either directly by code, or indirectly by dates/location), we'll make sure the codes match and update the TBA event accordingly. For new offseason events, we'll automatically put them in the mod suggestion queue.

@phil-lopreiato I've modified your offseason helper to only return two groups - matched and unmatched events. There didn't seem to be much of a difference between "matched" and "likely" events, so I simplified it for clarity.


controllers/datafeed_controller.py
```diff
@@ -453,14 +453,29 @@ def get(self, year):
...
+        fmsapi_events_offseason = [e for e in fmsapi_events if e.is_offseason]
+        events_to_put = set(fmsapi_events) - set(fmsapi_events_offseason)
+
+        matched_offseason_events, new_offseason_events = \
+            OffseasonEventHelper.categorize_offseasons(int(year), fmsapi_events_offseason)
+
+        # For all matched offseason events, make sure the FIRST code matches the TBA FIRST code
+        for tba_event, first_event in matched_offseason_events:
+            tba_event.first_code = first_event.event_short
+            events_to_put.append(tba_event) # Update TBA events - discard the FIRST event
```
ZachOrr: This seems half-done - do we want to update the underlying TBA event with the upstream FIRST event data? Does this get taken care of when we kickoff an `event_details` task for these events?

backend_main.py
```diff
@@ -21,7 +21,6 @@
                                ('/backend-tasks/get/team_avatar/(.*)', TeamAvatarGet),
                                ('/backend-tasks/do/team_blacklist_website/(.*)', TeamBlacklistWebsiteDo),
                                ('/backend-tasks/get/district_rankings/(.*)', DistrictRankingsGet),
-                               ('/backend-tasks/get/fms-sync-offseasons/(.*)', OffseasonEventListGet),
```
phil-lopreiato: yeah, it's time for this job to go

controllers/datafeed_controller.py
phil-lopreiato: I'm pretty sure this will be sufficient - the `event_details` task will use the `first_code` when fetching data from upstream.

tests/test_event.py
```diff
@@ -141,7 +156,6 @@ def __init__(self, week, year):
 
     @property
     def week(self):
-        print(self._week)
```
phil-lopreiato: lol

tests/test_offseason_event_helper.py
```diff
@@ -0,0 +1,223 @@
...
+from google.appengine.ext import testbed
+
+from consts.event_type import EventType
+from consts.district_type import DistrictType
+
+from helpers.event.offseason_event_helper import OffseasonEventHelper
+from models.event import Event
+
+
+class TestOffseasonEventHelper(unittest2.TestCase):
```
phil-lopreiato: :heart:

PR #2578, opened 2019-09-27
Remove unused imports from SuggestApiWriteReviewController

ZachOrr:
I was doing some work in a different branch and saw this file used `app_identity` - but it doesn't actually use `app_identity` or `tba_config`. Both dependencies got removed in https://github.com/the-blue-alliance/the-blue-alliance/commit/48e2c21bf82fd3a18db07c8d912bb55a148833b8#diff-7f331489138411648b1b87c7bccfc913

And it looks like we've never used `logging` in this controller


fangeugene:
We need better linters.

PR #2581, opened 2019-09-29
Update Docker Ubuntu version to 19.04

ZachOrr:
This fixes the locale error issue I have when installing system-wide dependencies inside the container via pip, as well as gives us a version of Python 3.7 inside our dev container (I know we're still a little ways off from any upgrading)

### DIFF
--- a/ops/dev/Dockerfile
+++ b/ops/dev/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16.04
+FROM ubuntu:19.04
 MAINTAINER The Blue Alliance
 
 # Set debconf to run non-interactively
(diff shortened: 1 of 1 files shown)


ops/dev/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN apt-get update && apt-get install -y \
 # Configure ssh server
 RUN mkdir /var/run/sshd
 RUN echo 'root:tba' | chpasswd
-RUN sed -i 's/PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config
+RUN sed -i 's/#PermitRootLogin prohibit-password/PermitRootLogin yes/' /etc/ssh/sshd_config
```
ZachOrr: This is commented out by default in the new config - we could also replace this line with

```
RUN echo "PermitRootLogin yes" >> /etc/ssh/sshd_config
```

Either way works - I kept the `sed` mostly to show that removing the `sed` in favor of an `echo` wouldn't be a regression

PR #2582, opened 2019-09-29
Add Google Analytics ID Sitevar

ZachOrr:
More cleanup - adds a `GoogleAnalyticsID` sitevar

Tested both the APIv2 and APIv3 controllers - having no `GOOGLE_ANALYTICS_ID` value set properly no-ops/logs having it set does hit the GA API and returns a 200.

I can't confirm that the analytics are showing up properly, since I don't see them in my staging GA, but I don't think there's a change that would break it.

### DIFF
--- a/api/apiv3/api_base_controller.py
+++ b/api/apiv3/api_base_controller.py
@@ -3,17 +3,13 @@ import logging
 import random
 import tba_config
 import time
-import urllib
-import uuid
 import webapp2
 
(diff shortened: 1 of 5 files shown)


ZachOrr:
It also feels like we should condense our Google Analytics logging in to a helper, since this is repeated code throughout the codebase. Let's land this first though, and then we can do the next cleanup step.

PR #2583, opened 2019-09-29
Remove APIv1 routes

ZachOrr:
Saw that the old APIv1 routes were still in `api_main.py` - they've been deprecated for going on 5 years and the endpoints don't go anywhere. Feels like a safe removal.

The last Google Analytics tracking call for the `api` event happened on May 28th, 2014 - back when we were using the old Google Analytics ID that pointed at web. I don't see any calls for `api` event with the API Google Analytics ID.


fangeugene:
Fine by me.

PR #2584, opened 2019-09-29
Fix 2020+ event sorting

ZachOrr:
Confirmed on my staging instance that 2020 events match up properly with what's on frc-events, and that re-scraped 2019 events are still sorted properly

Screen Shot 2019-09-29 at 5 51 54 PM

**Edit: Ignore this part - see https://github.com/the-blue-alliance/the-blue-alliance/pull/2584#issuecomment-536346649 - our week sorting is right**
The one exception is this Israeli DCMP which sorts as being Week 10 in our system, but shows up as Week 9 in frc-events. 99% success rate is pretty good I guess.

Screen Shot 2019-09-29 at 5 58 18 PM

No tests because holy cow it takes so much setup to tests these - I can file a ticket to make sure we get tests

### DIFF
--- a/database/event_query.py
+++ b/database/event_query.py
@@ -10,7 +10,7 @@ from models.team import Team
 
 
 class EventQuery(DatabaseQuery):
-    CACHE_VERSION = 0
+    CACHE_VERSION = 1
     CACHE_KEY_FORMAT = 'event_{}'  # (event_key)
     DICT_CONVERTER = EventConverter
 
@@ -22,7 +22,7 @@ class EventQuery(DatabaseQuery):
 
 
 class EventListQuery(DatabaseQuery):
-    CACHE_VERSION = 1
+    CACHE_VERSION = 2
     CACHE_KEY_FORMAT = 'event_list_{}'  # (year)
     DICT_CONVERTER = EventConverter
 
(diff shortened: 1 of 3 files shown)


ZachOrr:
Oh wild - actually, that Week 10 event is *actually* on Week 10. I don't know why it's in frc-events as Week 9. Week 7 starts on the 6th, Week 8 is the 13th, Week 9 is the 20th through the 26th, and Week 10 starts on the 27th.

So we're right.

fangeugene:
What's the call on numbering weeks with no events? FIRST doesn't count those weeks, but it makes the logic harder...

ZachOrr:
I can update my PR to take in to account weeks without events. Feels like the right thing to do, even if it's harder

dracco1993:
Two thoughts:
1. It's easier to do date math in your head (and visually) if `Week X` was always one calendar week off from `Week X+1`
2. In general, I'd say that matching how FIRST labels things is a good principle to follow

Yes, those are contradictory arguments.  👀

fangeugene:
Related to skipping weeks without events: How does cache clearing for event week even work right now? I'm pretty sure the current code won't properly serve an event's week in the API if the first event changes it's starting date, since all other week computations are based off of that.

This is going to be even more problematic if determine the week requires fetching all of a year's events instead of just the first.

ZachOrr:
Re: Slack conversation - I'm going to close this

PR #2588, opened 2019-10-01
Add 8000-level page to team list

lethosor:
## Description
This adds a "8000's" tab to the team list, and hopefully makes it a bit easier to change in the future.

## Motivation and Context
https://www.thebluealliance.com/teams/9 (8000s) currently 404s. This makes it work.

## How Has This Been Tested?
Locally, by importing 2020milak (which currently has 8126 registered)

image

## Types of changes

### DIFF
--- a/controllers/team_controller.py
+++ b/controllers/team_controller.py
@@ -13,7 +13,7 @@ from template_engine import jinja2_engine
 
 
 class TeamList(CacheableHandler):
-    VALID_PAGES = [1, 2, 3, 4, 5, 6, 7, 8]
+    VALID_PAGES = range(1, (8000 // 1000) + 2)
     CACHE_VERSION = 2
     CACHE_KEY_FORMAT = "team_list_{}"  # (page)
 
--- a/controllers/team_controller.py
+++ b/controllers/team_controller.py
@@ -1,4 +1,5 @@
 import datetime
+import math
 import os
 
 from google.appengine.ext import ndb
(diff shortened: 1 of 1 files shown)


controllers/team_controller.py
```diff
@@ -13,7 +13,7 @@
 
 
 class TeamList(CacheableHandler):
-    VALID_PAGES = [1, 2, 3, 4, 5, 6, 7, 8]
+    VALID_PAGES = range(1, (8000 // 1000) + 2)
```
ZachOrr: This approach feels a little confusing from several magic numbers. If we add a comment that gives a little information (`support up to 8999 teams - 1000 per page`) that would probably be helpful. Better yet - we could make these numbers constants and give the constants some comments and descriptive names.

Also changing to `(9000 // 1000) + 1` and having a comment about making the range inclusive in Python feels slightly better than the `+ 2` (they both kinda blow but - one feels better IMO)

lethosor: Yeah, not sure what I like here, but I added in some constants and got rid of the +2 bit

controllers/team_controller.py
```diff
@@ -13,7 +14,9 @@
 
 
 class TeamList(CacheableHandler):
-    VALID_PAGES = [1, 2, 3, 4, 5, 6, 7, 8]
+    MAX_TEAM_NUMBER = 8999
```
ZachOrr: We can probably just set this to 9000 and avoid bringing in the math lib

Issue #2227, opened 2018-04-02
Deprecate all Jinja1 templates by migrating to Jinja2


gregmarra:
We have some template files that are duplicated btw Jinja1 and Jinja2 versions. It'd be a small cleanliness win to migrate everything to Jinja2.

## Expected Behavior
Only one copy of each template.

## Current Behavior
Some templates are duplicated.

## Possible Solution
Move everything to Jinja2.

## Context
I made a change where I only edited the Jinja1 version of a template, so didn't change all instances of where that concept was used.

bdaroz:
Completed in #2589

PR #2589, opened 2019-10-02
Kill Main jinja1 templates

ZachOrr:
Related to #2227 - this kills all the jinja1 templates used in `controllers/main_controller.py`

I've verified all pages/partials work properly (outside of the `index_insights` template) - see screenshots here https://docs.google.com/document/d/1aA9NmlIJ5CBNYdKDRW9J0K2PsTuQNylnLt5kQluKyN4/edit?usp=sharing

For date format conversions, we're going from using the [Django date format](https://github.com/django/django/blob/master/django/utils/dateformat.py) to using the [strftime format](https://docs.python.org/2/library/datetime.html#strftime-strptime-behavior) - I've verified the conversions are correct, but feel free to double check my work

Comments on this PR should clarify why certain decisions were made, but I'm happy to modify as needed.


controllers/main_controller.py
```diff
@@ -285,20 +262,16 @@ def __init__(self, *args, **kw):
         self._cache_expiration = 60 * 60 * 24
 
     def _render(self, *args, **kw):
-        kickoff_datetime_utc = datetime.datetime.strptime(
```
ZachOrr: Note that - we pull this `kickoff_datetime` variable for both the kickoff and offseason template. But in the kickoff template, we're (properly) treating it as an EST datetime, and in the offseason template we're treating it as a UTC datetime. We don't store these values with timezones either, which makes datetime logic weird.

Anyways - this feels like too much scope for a single controller to be doing, which is why I've pulled all of this datetime logic out in to the `SeasonHelper`

template_engine/jinja2_engine.py
```diff
@@ -12,18 +12,19 @@ def get_jinja_env(force_filesystemloader=False):
         env = jinja2.Environment(
             auto_reload=False,
             loader=jinja2.ModuleLoader(os.path.join(os.path.dirname(__file__), '../templates_jinja2_compiled.zip')),
-            extensions=['jinja2.ext.autoescape'],
+            extensions=['jinja2.ext.autoescape', 'jinja2.ext.with_'],
```
ZachOrr: If we update to `jinja2==2.9` we don't need to pull in these extensions anymore - but upgrading our jinja to a new version felt out of scope for this work. I'll probably file a ticket to bump that dependency.

templates_jinja2/event_partials/event_table.html
```diff
@@ -42,7 +42,7 @@
             {% endif %}
           {% endif %}
         </td>
-        <td>{% if event.start_date %}<time itemprop="startDate" datetime="{{event.start_date|strftime("%c")}}">{{ event.start_date|strftime("%b %d") }}{% if event.start_date.date() != event.end_date.date() %}</time> to <time itemprop="endDate" datetime="{{event.end_date|strftime("%c")}}">{{ event.end_date|strftime("%b %d") }}{% endif %}{{ event.end_date|strftime(", %Y") }}</time>{% endif %}</td>
```
ZachOrr: `strftime("%c")` actually isn't what we're looking for here - my guess is this got botched from a `date:c` conversion. The `c` format is isoformat https://github.com/django/django/blob/master/django/utils/dateformat.py#L217

templates_jinja2/index/index_buildseason.html
```diff
@@ -45,9 +45,7 @@ <h2>
 
       {% if events %}
         <h2>This Week's Events</h2>
-        {% with events as events %}
-          {% include "event_partials/event_table.html" %}
-        {% endwith %}
+        {% include "event_partials/event_table.html" %}
```
ZachOrr: We don't need to do `with` here since `events` is already namespaced properly

controllers/main_controller.py
```diff
@@ -533,8 +498,7 @@ def __init__(self, *args, **kw):
         self._cache_expiration = 60 * 60 * 24 * 7
 
     def _render(self, *args, **kw):
-        path = os.path.join(os.path.dirname(__file__), "../templates/apiwrite.html")
-        return template.render(path, self.template_values)
+        return jinja2_engine.render('apiwrite.html', self.template_values)
 
 
 class MatchInputHandler(CacheableHandler):
```
ZachOrr: Can we kill this altogether? This mentions being for the 2015 Championship - feels like we might not need this anymore?

controllers/main_controller.py
phil-lopreiato: uh, this is clowny and probably my fault :(

templates_jinja2/event_partials/event_table.html
phil-lopreiato: :+1:

controllers/main_controller.py
phil-lopreiato: yeah, this can go

PR #2602, opened 2019-10-12
Add "OTHER" award type to award helper

FletcherS7:
## Description
Added OTHER award type to the award helper.

## Motivation and Context
Allow submitting other award type from Cheesy Arena

## How Has This Been Tested?
Simple change. 
not tested. 

## Types of changes


fangeugene:
This actually doesn't work. Awards of the same type will get merged into one award. If we want to do this, we need `OTHER1`, `OTHER2`, `OTHER3`... for as many "Other" awards we want to support.

FletcherS7:
Would it not be fine if only one award was submitted as other per event?

fangeugene:
Yes that would be fine.

FletcherS7:
sounds good.
I might add a note to the cheesy award display.

PR #2606, opened 2019-10-22
Fix 2020 event's weeks after Week 8 being off-by-one

ZachOrr:
## Description
Adds a case when figuring out `week` for an event to account for the empty week between Week 8 and Week 9 in 2020

## Motivation and Context
Right now, 2020 events after Week 8 are shown with the wrong week.

Screen Shot 2019-10-22 at 12 00 57 PM

## How Has This Been Tested?
Tested in a staging instance - made sure that `2020cmptx` (although we don't return a week for CMP events) was still being calculated as Week 8, whereas `2020iscmp`/`2020cmpmi` are being calculated as Week 9

Screen Shot 2019-10-22 at 11 51 47 AM

## Types of changes

### DIFF
--- a/models/event.py
+++ b/models/event.py
@@ -254,7 +254,12 @@ class Event(ndb.Model):
 
         if self._week is None and week_start is not None:
             days = (self.start_date - week_start).days
-            self._week = days / 7
+            week = days / 7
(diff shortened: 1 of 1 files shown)


fangeugene:
Looks good, will merge after checks pass.

PR #2609, opened 2019-10-25
Fix #2552

### LINKED ISSUE #2552: Odd button alignment
The **Watch Now** button seems to be floating over an `<hr>`. This might be caused by the long description in it's way?

tbacss

jaredhasenklein:
Fixes #2552 - the live stream button aliment issue.
## Description
Instead of using `pull-right` use `float-right`. See examples below.
#### Current System
Currently, the button has the `pull-right` attribute which aligns it to the right of the section. This runs into an issue with any text that approaches the area of the button since thy can get very close to overlap, making it look awkward, as pointed out in #2552.
#### Short title example
image
#### Long title exampe
image

### Proposed change
Instead of using `pull-right` for the button, use `float-right`. This will not align it to the right of the section but will simply place the button to the right of the text.
#### Short title example
image
#### Long title example
image
## Motivation and Context
Issue per #2552

## How Has This Been Tested?
Edited HTML directly in Chrome (sorry, not sorry).

See above under description.

## Types of changes

### DIFF
--- a/templates_jinja2/media_partials/live_special_webcast_partial.html
+++ b/templates_jinja2/media_partials/live_special_webcast_partial.html
@@ -4,7 +4,7 @@
 {% for webcast in special_webcasts %}
   {% if webcast.status == 'online' %}
     <h3>
-      <a class="btn btn-md btn-success float-right"
+      <a class="btn btn-md btn-success pull-right"
          href="/gameday#layout=0&view_0={{ webcast.key_name }}-0{% if webcast.type == 'twitch' %}&chat={{ webcast.channel }}{% endif %}">
         <span class="glyphicon glyphicon-facetime-video"></span> Watch Now
       </a> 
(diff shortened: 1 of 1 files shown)


lethosor:
I think buttons for multiple streams were intended to be aligned on the right, which this doesn't do anymore. Maybe moving the link before the stream title in the header and using `pull-right` would work, i.e.
```html
    <h3>
      <a class="btn btn-md btn-success pull-right"
         href="/gameday#layout=0&view_0={{ webcast.key_name }}-0{% if webcast.type == 'twitch' %}&chat={{ webcast.channel }}{% endif %}">
        <span class="glyphicon glyphicon-facetime-video"></span> Watch Now
      </a>
      {{ webcast.name }}
      <small>{{ webcast.stream_title }}</small>
```

jaredhasenklein:
That's a solid idea. Fixed it above.

PR #2616, opened 2019-11-16
Let simple matches return blank strings

Spazzlo:
Fixed the match and match simple schemas

## Description
Simple matches that had no winner (tie) returned empty strings (as they should), but client apis crashed when they received it. Also, added enums to the winning_alliance value for normal matches, since they should be the same.

## Motivation and Context
This fixes official client apis crashing when requesting a simple match that ended in a tie

## How Has This Been Tested?
I compiled a client api with this updated schema and it fixed the issue

## Types of changes


Spazzlo:
Committed!

PR #2624, opened 2019-12-03
Fix webhook verifies with X-TBA-HMAC.

brennonbrimhall:
Re-implement #2608, fix issues that led to #2620.

## Description

* Add in `str()` casts to prevent breakage of webhook verifies.
* Always generate webhook secrets server-side.
* Disallow client supplying webhook secrets.

## Motivation and Context

See #2608, #2620.

## How Has This Been Tested?

Given that #2608 passed the test suite, I'm going to suggest that we
have a very thorough code reveiew.

## Types of changes

- [X] Breaking change (fix or feature that would change API specifications or require data migrations)


controllers/webhook_controller.py
```diff
@@ -30,10 +30,7 @@ def post(self):
             return self.redirect('/webhooks/add?error=1')
 
         # Secret may be none - but we'll generate a secret for the user
```
ZachOrr: nit: Update this comment

brennonbrimhall:
Important note: as far as I can tell, `uuid.uuid4()` is not guaranteed to be cryptographically secure.  Ideally, we would upgrade to Python 3 and use the `secrets` API...but that's going to be a while.

templates/webhookdocs.html
```diff
@@ -47,7 +47,7 @@ <h2 id="overview">Overview</h2>
...
-        <p>Each request also contains a <code>X-TBA-Checksum</code> header that you can use to validate requests on your client. The header will contain a hash of a secret key that you set when you create the webhook concatenated with the payload string. This is caculated using SHA1(secret+payload). You can see <a href="https://github.com/the-blue-alliance/the-blue-alliance/blob/master/helpers/notification_sender.py#L32" target="_blank">in the code</a> where this is calculated.
+        <p>Each request also contains a <code>X-TBA-HMAC</code> header that you can use to validate requests on your client. The header will contain a hash of a secret key that is set by the server when you create the webhook concatenated with the payload string. This is caculated using <a href="https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.198-1.pdf">NIST's HMAC algorithm</a>, using SHA256. You can see <a href="https://github.com/the-blue-alliance/the-blue-alliance/blob/master/helpers/notification_sender.py#L32" target="_blank">in the code</a> where this is calculated.
```
ZachOrr: 👍

tests/tbans_tests/requests/test_webhook_request.py
```diff
@@ -21,7 +21,7 @@ def tearDown(self):
     def _stub_webhook_request(self, url, payload, method, headers, request, response, follow_redirects=False, deadline=None, validate_certificate=None, http_proxy=None):
         # While we're here - check that our request looks right
         self.assertEqual(url, 'https://www.thebluealliance.com/')
-        self.assertEqual({header.key(): header.value() for header in headers}, {'X-TBA-Checksum': 'dbecd85ae53d221c387647085912d2107fa04cd5', 'Content-Type': 'application/json', 'X-TBA-Version': '1'})
+        self.assertEqual({header.key(): header.value() for header in headers}, {'X-TBA-Checksum': 'dbecd85ae53d221c387647085912d2107fa04cd5', 'X-TBA-HMAC': 'fb2b61d55884a648b35801688754924778b3e71ea2bf8f5effb0f3ffecb5c940', 'Content-Type': 'application/json', 'X-TBA-Version': '1'})
```
ZachOrr: Let's add another test in this class that uses a unicode string for a secret and makes sure no exceptions are thrown when generating the message

tbans/requests/webhook_request.py
```diff
@@ -72,9 +74,16 @@ def _json_string(self):
...
         ch = hashlib.sha1()
         ch.update(self.secret)
         ch.update(payload)
         return ch.hexdigest()
+
+    def _generate_webhook_hmac(self, payload):
+        import hashlib
+        import hmac
+
+        return hmac.new(str(self.secret), str(payload), hashlib.sha256).hexdigest()
```
ZachOrr: We shouldn't *need* to `str` payload, because we use `ensure_ascii=True` which ensures we'll get a `str` back with only ASCII characters. But this feels like an implementation detail, so it's probably fine

brennonbrimhall:
We should be good to go now, barring any further feedback.  I've also gone ahead and squashed the intermediate edits out for a nicer history.

controllers/webhook_controller.py
```diff
@@ -29,11 +29,8 @@ def post(self):
         if not url or not name:
             return self.redirect('/webhooks/add?error=1')
 
-        # Secret may be none - but we'll generate a secret for the user
-        secret = self.request.get('secret', None)
-        if not secret:
-            import uuid
```
ZachOrr: We need to make sure to add this UUID import back in

PR #2634, opened 2019-12-30
Fix per-year bugs in redeeming team admin access

### LINKED ISSUE #2633: Unable to Redeem Team Mod Code Two Years In A Row
This fails if you've already redeemed a code in a prior year.

It shouldn't...

phil-lopreiato:
No longer fail if the user has redeemed a team admin code for a prior year

## Description
We should be checking the years in here

## Motivation and Context
Fixes #2633 

## How Has This Been Tested?
added unit test

### DIFF
--- a/controllers/team_admin_controller.py
+++ b/controllers/team_admin_controller.py
@@ -212,12 +212,6 @@ class TeamAdminRedeem(LoggedInHandler):
         self._require_registration()
         user = self.user_bundle.account.key
 
-        existing_access = TeamAdminAccess.query(
-            TeamAdminAccess.account == user).count(1)
-        if existing_access > 0:
-            self.redirect('/mod/redeem?status=already_linked')
-            return
-
         team_number = self.request.get("team_number")
         if not team_number or not team_number.isdigit():
             self.redirect('/mod/redeem?status=invalid_code')
--- a/templates_jinja2/team_admin_redeem.html
+++ b/templates_jinja2/team_admin_redeem.html
@@ -12,10 +12,6 @@
   <div class="alert alert-danger">
     Invalid auth code!
   </div>
-{% elif status == "alread_linked" %}
-  <div class="alert alert-danger">
-    This account is already linked to a team!
-  </div>
 {% elif status == "code_used" %}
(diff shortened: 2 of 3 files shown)


controllers/team_admin_controller.py
```diff
@@ -83,7 +83,8 @@ def get(self):
             for team in teams_future
         }
         team_num_to_robot_name = {
-            int(robot.get_result().team.id()[3:]): robot.get_result().robot_name
+            int(robot.get_result().team.id()[3:]):
+            robot.get_result().robot_name
```
gregmarra: Should this be indented for style?

controllers/team_admin_controller.py
```diff
@@ -112,14 +113,22 @@ def get(self):
...
         self.template_values.update({
-            "existing_access": existing_access,
-            "teams": team_num_to_team,
-            "robot_names_by_team": team_num_to_robot_name,
-            "team_medias": team_medias,
-            "team_social_medias": team_social_medias,
-            "suggestions_by_team": suggestions_by_team,
-            "suggestion_names": self.SUGGESTION_NAMES,
-            "suggestion_review_urls": self.SUGGESTION_REVIEW_URL,
+            "existing_access":
```
gregmarra: Why split these?

gregmarra:
FIRST released 2020 codes in 2019. You're generating `TeamAdminAccess.year` from datetime.now(). Will that create any issues?

controllers/team_admin_controller.py
phil-lopreiato: argh, stupid autoformat

controllers/team_admin_controller.py
phil-lopreiato: more autoformat :(

gregmarra:
I'd probably allow them to work on any year, and expire each year. You're
basically an authorized rep of that team's full history, but you shouldn't
be a rep 2 years from now.

phil-lopreiato:
Updated to accept codes from any year, removes requirement that one account be linked to a team at any point in time

PR #2637, opened 2020-01-02
Change build season to count down to first event

fangeugene:
## Description
Display countdown to the first official event of the season.

## Motivation and Context
Stop build day no longer exists

## How Has This Been Tested?
Local dev

image

## Types of changes

### DIFF
--- a/controllers/main_controller.py
+++ b/controllers/main_controller.py
@@ -150,8 +150,7 @@ class MainBuildseasonHandler(CacheableHandler):
         effective_season_year = SeasonHelper.effective_season_year()
 
         self.template_values.update({
-            'endbuild_datetime_est': SeasonHelper.stop_build_datetime_est(effective_season_year),
-            'endbuild_datetime_utc': SeasonHelper.stop_build_datetime_utc(effective_season_year),
+            'seasonstart_datetime_utc': SeasonHelper.first_event_datetime_utc(effective_season_year),
             'events': EventHelper.getWeekEvents(),
         })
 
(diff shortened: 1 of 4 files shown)


controllers/main_controller.py
```diff
@@ -150,8 +150,7 @@ def _render(self, *args, **kw):
         effective_season_year = SeasonHelper.effective_season_year()
 
         self.template_values.update({
-            'endbuild_datetime_est': SeasonHelper.stop_build_datetime_est(effective_season_year),
-            'endbuild_datetime_utc': SeasonHelper.stop_build_datetime_utc(effective_season_year),
```
ZachOrr: We should be able to kill `stop_build_datetime_est` and `stop_build_datetime_utc` then, right? It's all dead code now.

helpers/season_helper.py
```diff
@@ -74,3 +75,16 @@ def stop_build_datetime_est(year=datetime.now().year):
     def stop_build_datetime_utc(year=datetime.now().year):
         """ Converts stop_build_date to a UTC datetime """
         return SeasonHelper.stop_build_datetime_est(year).astimezone(UTC)
+
+    @staticmethod
+    def first_event_datetime_utc(year=datetime.now().year):
+        """ Computes day the first in-season event begins """
+        events = event_query.EventListQuery(year).fetch()
```
ZachOrr: I think we might already have an index for querying events by start date and filtering for only official events, if we wanna do that. But I don't care much either way 🤷‍♂

PR #2640, opened 2020-01-03
Zebra MotionWorks Trusted API

fangeugene:
Enable adding Zebra MotionWorks data through the trusted API

## Description
JSON format is like:
```
[{
    key: '2020casj_qm1',
    times: [<elapsed time (float seconds)>, 0.0, 0.5, 1.0, 1.5, ...],
    red: [
        {
            team_key: "frc254",
            xs: [<float feet or null>, null, 1.2, 1.3, 1.4, ...],
            ys: [<float feet or null>, null, 0.1, 0.1, 0.1, ...],
        },
        {
            team_key: "frc971",
            xs: [1.1, 1.2, 1.3, 1.4, ...],
            ys: [0.1, 0.1, 0.1, 0.1, ...],
        },
        ...
    ],
    blue: [...],
},
{
  key: '2020casj_qm2',
  ...
}]
```

## Motivation and Context
https://www.chiefdelphi.com/t/zebra-motionworks-for-frc-formerly-zebra-dart-for-frc-2020-teaser/368918

## How Has This Been Tested?
Local dev tests

## Types of changes


models/zebra_motion.py
```diff
@@ -0,0 +1,9 @@
+from google.appengine.ext import ndb
+
+
+class ZebraMotion(ndb.Model):
```
phil-lopreiato: any particular reason why this is its own db model rather than being inline with match (it probably doesn't matter much though)

ZachOrr: Keeping our data models light-er by splitting out different types of data seems fine to me. The idea being we won't always need Zebra data if we're querying for a match (we probably just want time and whatnot)

fangeugene: Also, I'm worried about the QueryCache for EventMatches being > 1MB if we put it all in the same model.

phil-lopreiato: works for me (I just wanted to have the reason documented here)

tests/test_api_trusted.py
```diff
@@ -1096,3 +1103,65 @@ def test_remapping_after(self):
             self.assertTrue(str(team) in {'5'})
         for team in Award.get_by_id('2014casj_2').recipient_dict.keys():
             self.assertTrue(str(team) in {'6'})
+
+    def test_zebra_motionworks_add(self):
```
ZachOrr: Should we add test cases for non-success states as well?

datafeeds/parsers/json/json_zebra_motion_parser.py
```diff
@@ -0,0 +1,100 @@
...
+
+class JSONZebraMotionWorksParser(ParserBase):
+    @classmethod
+    def parse(cls, zebra_motion_json):
+        """
+        Parse JSON that contains Zebra MotionWorks data
+        Format is as follows:
+        [{
+            key: '2020casj_qm1',
+            times: [<elapsed time (float seconds)>, 0.0, 0.5, 1.0, 1.5, ...],
```
ZachOrr: Just confirming - the times I saw in the sample Zebra data on Chief looked to be absolute times, not relative times. Are these going to be absolute or relative?

tests/test_api_trusted.py
fangeugene: Yep, I'll add some.

datafeeds/parsers/json/json_zebra_motion_parser.py
fangeugene: We're talking to them, and they will be relative.

models/zebra_motionworks.py
```diff
@@ -0,0 +1,9 @@
+from google.appengine.ext import ndb
+
+
+class ZebraMotionWorks(ndb.Model):
+    """
+    The ZebraMotionWorks model represents robot tracking data from the
+    Zebra MotionWorks system
+    """
```
phil-lopreiato: many want to include the schema in this comment as well?

fangeugene: Schema is described in the parser. Maybe I'll make a note of it.

PR #2648, opened 2020-01-10
Support avatars for 2020

ZachOrr:
Based on https://github.com/the-blue-alliance/the-blue-alliance/commit/906f35c075ad07f777f3f0170e3e3ce69c98b409


ofekashery:
We should also update the link here:
https://github.com/the-blue-alliance/the-blue-alliance/blob/cd72707230728ac7518c26a931f58aed6b035937/templates_jinja2/team_list.html#L23

ZachOrr:
Ah good catch - I was wondering where the links went. Thanks!

PR #2649, opened 2020-01-10
Add Broadcast notifications to TBANS

ZachOrr:
Adds support for the the Broadcast notification for FCM clients, moves the Broadcast webhooks to be handled by TBANS, and consolidates Android notification calling code to exist in TBANS.

This also wires up the batch sending/retrying of FCM messages, which has through tests. Under proper failure conditions, we'll attempt to resend to failed FCM clients up to 5 additional times using exponential backoff between 1 and 32 seconds. Logging is in place for deletions, errors, and critical paths. Errors and critical logs will get alerting metrics.

### DIFF
--- a/controllers/admin/admin_mobile_controller.py
+++ b/controllers/admin/admin_mobile_controller.py
@@ -8,7 +8,7 @@ from google.appengine.ext.webapp import template
 
 from consts.client_type import ClientType
 from controllers.base_controller import LoggedInHandler
-from helpers.notification_helper import NotificationHelper
+from helpers.tbans_helper import TBANSHelper
 from models.mobile_client import MobileClient
 from sitevars.notifications_enable import NotificationsEnable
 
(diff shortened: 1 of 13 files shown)


controllers/admin/admin_mobile_controller.py
```diff
@@ -121,8 +122,8 @@ def post(self):
             return
 
         try:
-            clients = [int(c) for c in clients]
-            deferred.defer(NotificationHelper.send_broadcast, clients, title, message, url, app_version, _queue="admin")
+            client_types = [int(c) for c in client_types]
+            deferred.defer(TBANSHelper.broadcast, client_types, title, message, url, app_version, _queue="admin")
```
ZachOrr: This is going to be a problem - we can't defer static calls like this https://cloud.google.com/appengine/docs/standard/python/refdocs/google.appengine.ext.deferred.deferred

PR #2650, opened 2020-01-11
Enable event insights for 2020

ZachOrr:
Going to let this hang out for a moment - feel free to merge when we're ready to enable these cron jobs for 2020


ZachOrr:
Mmm just kidding - there needs to be a lot more backend work before enabling the cron jobs... closing for now

PR #2661, opened 2020-01-19
Conditional the FMS API callout to Events during or after 2006

ZachOrr:
Can someone (cc @phil-lopreiato) confirm we started pulling Event info from the FMS API starting in 2015? https://github.com/the-blue-alliance/the-blue-alliance/commit/3ba1a6eb94a7b1a8d31ba114fe428c89a48809f8


phil-lopreiato:
Yes, 2015 was the year the frc events api was released, although they have now backfilled older events so this pull might not be fully necessary (if old event links do work, then we probably don't need this)

ZachOrr:
Alright - setting it to the min limit of the FMS API then (2006) and gonna call it good for now 🤷‍♀

PR #2674, opened 2020-01-24
Send myTBA update favorites and update subscriptions via TBANS

ZachOrr:
Going to do some integration tests later to make sure things are working properly

### DIFF
--- a/helpers/mytba_helper.py
+++ b/helpers/mytba_helper.py
@@ -1,6 +1,6 @@
 from google.appengine.ext import ndb
 
-from helpers.notification_helper import NotificationHelper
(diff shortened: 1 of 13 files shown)


ZachOrr:
This is actually not super well thought out - the idea of keeping favorites and subscriptions up-to-date client-side is helpful on Android because we generate local notifications based on data push notifications, and we do rich text formatting. On iOS, keeping the data up-to-date is important for features like filtering match lists and whatnot, but that's all UI-related. I probably won't support these notifications for TBANS, and we'll keep the existing Android codepaths in place until we migrate Android over.

PR #2687, opened 2020-02-02
Better Awards notifications, update webhooks docs

ZachOrr:
A bit of "two PRs in one" because things got a little out-of-hand

## Update Awards notification
* Update Awards notifications to send more-specific Team and Event notifications - Team notifications will have a more detailed description of the Awards a team won, Event notifications are unchanged
(Will get screenshots from mobile apps hopefully)

## Update Webhooks Docs
* Update docs (starting with Awards, working my way through them all as I update the endpoints) to provide clearer descriptions of the fields and conditions under which we'll send notifications
* Only show notification testing boxes if the user is logged in
* Add sending to TBANS for Awards notification debugging
* Update Awards notification documentation for new behavior

Screen Shot 2020-02-02 at 12 53 41 PM
Screen Shot 2020-02-02 at 12 53 41 PM

### DIFF
--- a/controllers/base_controller.py
+++ b/controllers/base_controller.py
@@ -47,6 +47,10 @@ class CacheableHandler(webapp2.RequestHandler):
                     self.response.headers['Vary'] = 'Accept-Encoding'
 
     @property
+    def logged_in(self):
+        return self._user_bundle.user is not None
+
+    @property
     def cache_key(self):
         return self._render_cache_key(self._partial_cache_key)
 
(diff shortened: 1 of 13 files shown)


controllers/base_controller.py
```diff
@@ -46,6 +46,10 @@ def __init__(self, *args, **kw):
                 if self.response:
                     self.response.headers['Vary'] = 'Accept-Encoding'
 
+    @property
+    def logged_in(self):
+        return self._user_bundle.user is not None
+
```
ZachOrr: @fangeugene PTAL - does this seem okay? I want to know if a user is signed in in the `CacheableHandler` so we can conditionally hide/show the notification testing boxes on the webhooks page

fangeugene: This doesn't work, since the cached version could think that a user is signed in.

The correct thing to do is to not use `CacheableHandler` and just use `LoggedInHandler`

ZachOrr: Thanks! Made this change and that seems to work.

PR #2689, opened 2020-02-02
Revert "Revert "Better Awards notifications, update webhooks docs (#2687)" (#2688)

ZachOrr:
Re-introduces reverted changes - going to do some more testing in dev to make sure these changes are safe before merging

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,6 +85,7 @@ ops/tbatv-prod-hrd-deploy.json
 ops/dev/conf.json
 ops/dev/*-auth.json
 test_keys.json
+service-account-key.json
 
 # Vagrant #
 ###########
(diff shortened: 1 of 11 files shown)


ZachOrr:
Alright - I feel MUCH better about this. The list of changes include -

* Not double-pinging webhooks via TBANS and old notification codepaths when testing notifications
* Only allowing sending to webhooks via the webhook docs
* Not using production codepaths for the TBANS notifications - instead, calling in to a private method to only send to webhooks (gross but far less risky, if there's a regression)
* Setup some local-push-notification-testing stuff for TBANS (will write docs soon)

(Webhook testing element, when there's no verified webhooks)
Screen Shot 2020-02-02 at 8 38 34 PM

(Webhook testing element, when there are webhooks)
Screen Shot 2020-02-02 at 8 38 01 PM

Still left todo
- [ ] Update all the old webhook-testing sends so they take an explicit webhook

ZachOrr:
I'm actually going to back out the webhook docs changes for now - with only 18 days left it's more important that notification support gets shipped as opposed to webhook docs upgrades

PR #2694, opened 2020-02-11
Run paver make during PR on CI

ZachOrr:
Mostly attempting to fix `paver make`


ZachOrr:
Confirmed this is just a problem with `less@3.11.0` https://github.com/less/less.js/issues/3469 - we should just be able to pin our less version for now (we should be pinning everything but that's a whole thing)

PR #2713, opened 2020-02-17
Team attribution tooltips

jaredhasenklein:
Adds tooltips to the 2020 match breakdown.

## Description
On the 2020 match breakdown page, this PR will add tooltips over initiation line and end game actions, attributing it to the teams.

## Motivation and Context
On [Chief](https://www.chiefdelphi.com/t/the-blue-alliance-endgame-data/377891), someone was asking if endgame data was in the order of the schedule (e.g. is "Red Robot 1" the same as "Red 1" on the schedule). I thought it was fairly obvious but also it definitely makes sense to add these tooltips because if someone is scrolled down to the match breakdown, they shouldn't have to scroll back up to the schedule (so much work!).

## How Has This Been Tested?
Inspect element 🤣 

image
image

## Types of changes

### DIFF
--- a/templates_jinja2/match_partials/match_breakdown/match_breakdown_2020.html
+++ b/templates_jinja2/match_partials/match_breakdown/match_breakdown_2020.html
@@ -23,38 +23,38 @@
     <tr>
       <td class="red" colspan="2">
         {% if match.score_breakdown.red.initLineRobot1 == 'Exited' %}
-        <span class="glyphicon glyphicon-ok"></span>
+        <span class="glyphicon glyphicon-ok" rel="tooltip" title="{{match.alliances.red.teams.0}}"></span>
         {% else %}
-        <span class="glyphicon glyphicon-remove"></span>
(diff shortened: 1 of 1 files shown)


gregmarra:
These look wide for the amount of text in them – does Bootstrap set a
min-width?

jaredhasenklein:
We can define the width but for reference, here is an example of a tooltip already on the page. It also seems large for the amount of content:
image

We have it defined for surrogates and DQ at 130px. Here's how that looks:
image
image

jaredhasenklein:
https://github.com/the-blue-alliance/the-blue-alliance/blob/1324e9e5b7c4ab21315bd00a768112991bada108/static/css/less_css/less/tba/tba_match_table.less#L153

gregmarra:
I feel it'd be cleaner to change them so they just are the width of their
contents, but don't let that block this :)

jaredhasenklein:
Since I believe that is an independent change of this one, I created #2714.

PR #2714, opened 2020-02-17
Changes tooltip widths

jaredhasenklein:
Autofit tooltips to size of contents

## Description
Max-width isn't what we're concerned with here. It is the (currently undefined) `width` attribute that, when set to 100%, will fit the size of content.

## Motivation and Context
@gregmarra brought up a good point on #2713.

#### Current behavior:
image
image

#### New behavior:
image
image

## Types of changes


gregmarra:
Looks snazzy!

ofekashery:
Amazing
I'd add some more horizontal padding

PR #2754, opened 2020-03-01
Upgrade from ga.js to analytics.js

ZachOrr:
TIL our Google Analytics tracking script is legacy. This upgrades all the old `ga.js` to `analytics.js`

Migration information here https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs

I removed all the Google Analytics tracking from the datafeed and admin pages - it doesn't feel particularly useful to include those in our Google Analytics data.

### DIFF
--- a/templates/gameday2.html
+++ b/templates/gameday2.html
@@ -50,7 +50,7 @@
       return ((parseInt(t[1]) || 0) + 1) + ':00';
     })(tos.split(':').reverse());
     // Collect and send the time to Google Analytics
-    window.pageTracker ? pageTracker._trackEvent('gameday_time', 'Log', tos) : _gaq.push(['_trackEvent', 'gameday_time', 'Log', tos]);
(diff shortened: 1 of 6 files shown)


templates/gameday2.html
```diff
@@ -50,7 +50,7 @@
       return ((parseInt(t[1]) || 0) + 1) + ':00';
     })(tos.split(':').reverse());
     // Collect and send the time to Google Analytics
-    window.pageTracker ? pageTracker._trackEvent('gameday_time', 'Log', tos) : _gaq.push(['_trackEvent', 'gameday_time', 'Log', tos]);
+    ga('send', 'event', 'gameday_time', 'gameday_time', 'Log', tos);
```
ZachOrr: cc @fangeugene Can you double check this looks right? I don't want to break GameDay tracking

PR #2756, opened 2020-03-05
Add component OPRs to event pages

tervay:
## Description
Add component OPRs to API and to event details page.

## Motivation and Context
Lots of people have been looking for an easy way to view component OPRs, so its probably good to finally integrate it into TBA. Saves a lot of people a lot of time writing their own component OPR calculations and helps a lot of scouts around the world.

## How Has This Been Tested?
Only manual testing for now, will update as the PR evolves.

https://i.imgur.com/ZGL8JsJ.png

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)
- [ x? ] Breaking change (fix or feature that would change API specifications or require data migrations)


tervay:
https://github.com/the-blue-alliance/the-blue-alliance/blob/a578e5b9e0c85c087e22743102eeb37a9c1b2dc0/database/dict_converters/event_details_converter.py#L7

```
class EventDetailsConverter(ConverterBase):
        3: 3,
    }
```

I'm not exactly sure what qualifies "the dict" or "a change"; should this be modified to `4: 4`?

fangeugene:
`{3: 4}`

ZachOrr:
Let's make sure we get some tests too, if possible 🙆‍♂

tervay:
that didn't work the way I wanted it to

tervay:
I chose to re-base this into `py3` rather than `master` since it seems that's the branch for the future anyway

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -0,0 +1,72 @@
...
+            - match.alliances[get_opp_color(color)]["score"]
+        ),
+    }
+
+    COMPONENTS = {
+        2020: {
+            "Foul Points": lambda match, color: match.score_breakdown[color].get(
+                "foulPoints", 0
+            ),
+            "Auton div Total": lambda match, color: (
```
tervay, line 27: this is just a placeholder stat that is just to show off custom component OPR ability

src/backend/common/helpers/opr_helper.py
```diff
@@ -0,0 +1,126 @@
+# taken from https://github.com/the-blue-alliance/the-blue-alliance/blob/bd16a7daf19774a2dc84509f71fa786c45d0a99e/old_py2/helpers/matchstats_helper.py
```
tervay, line 1: I forgot to make sure git knows where this is moved from, oops. I can fix that later

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -0,0 +1,72 @@
+from typing import Dict, Callable, List
+
+from backend.common.consts.alliance_color import AllianceColor
+from backend.common.models.match import Match
+from backend.common.helpers.opr_helper import OPRHelper
+
+
+def get_opp_color(color: AllianceColor) -> AllianceColor:
```
phil-lopreiato, line 8: this can be done with `from ...alliance_color import OPPONENT`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -0,0 +1,72 @@
...
+from backend.common.models.match import Match
+from backend.common.helpers.opr_helper import OPRHelper
+
+
+def get_opp_color(color: AllianceColor) -> AllianceColor:
+    return [AllianceColor.RED, AllianceColor.BLUE][color == AllianceColor.RED]
+
+
+class MatchstatsHelper:
+    EVERGREEN: Dict[str, Callable[[Match, AllianceColor], float]] = {
```
phil-lopreiato, line 13: use `models.stats.StatType` instead of `str` here

src/backend/common/helpers/matchstats_helper.py
phil-lopreiato: Let's maybe use something we could publish for 2020. Like maybe cell or climb OPR?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -0,0 +1,72 @@
...
+            ),
+        }
+    }
+
+    component_default = lambda comp: (
+        lambda match, color: match.score_breakdown[color].get(comp, 0)
+    )
+
+    @classmethod
+    def calculate_matchstats(cls, matches: List[Match]):
```
phil-lopreiato, line 39: annotate a return type, you can do `models.stats.EventMatchStats`, I think (assuming we add the right stuff to the stat type enum)

tervay:
I still need to add tests for copr calculations. There is also a local bootstrap test failing and I'm not entirely sure how the test suite works in order for me to fix it. It just has an extra matchstats property or something along that lines that a certain  dictionary equality assertion isn't expecting.

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,27 +8,139 @@
 
 # x is OPR and should be n x 1
 
-from collections import defaultdict
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Callable, Dict, List
+from typing import Tuple
 
 import numpy as np
```
phil-lopreiato, line 14: add `from pyre_extensions import none_throws`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,27 +8,139 @@
...
+        if not (
+            matches is not None
+            and len(matches) > 0
+            and matches[0].score_breakdown is not None
+        ):
+            return coprs
+
+        # for each string-like key in the score_breakdown object
+        # (just take the red score_breakdown from match 0, it's an arbitrary selection)
+        for component, value in matches[0].score_breakdown[AllianceColor.RED].items():
```
phil-lopreiato, line 103: since you're already checking whether `score_breakdown` is `None` above, this is fine. `pyre` just can't track the nullability of properties (because other side effects can change them).

Do
```
none_throws(matches[0].score_breakdown)[AllianceColor.RED].items()
```

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -37,217 +149,79 @@ def build_team_mapping(
...
-    def build_Minv_matrix(
-        cls, matches: List[Match], team_id_map: TTeamIdMap, played_only: bool = False
-    ) -> "np.ndarray[int]":
-
+    def __build_m_inv_matrix(
+        cls, matches: List[Match], team_id_map: TTeamIdMap, played_only: bool = False,
+    ) -> "np.ndarray[float]":
         n = len(team_id_map.keys())
-        M = np.zeros((n, n))
+        m = np.zeros([n, n])
```
phil-lopreiato, line 172: this should either stay tuple `(n, n)`

src/backend/web/local/bootstrap.py
```diff
@@ -132,6 +133,24 @@ def update_event(cls, key: EventKey, auth_token: str) -> None:
         event_awards = cls.fetch_event_detail(key, "awards", auth_token)
         list(map(lambda t: cls.store_award(t, event), event_awards))
 
+        # Get insights from prod API
+        # event_oprs = cls.fetch_event_detail(key, "oprs", auth_token)
+        # cls.store_eventdetail(event, "matchstats", event_oprs)
+
+        # Calculate insights locally
+        matchstats_dict = MatchstatsHelper.calculate_matchstats(
```
phil-lopreiato, line 141: this bit (and the associated test fixes) should probably be split out into a separate PR we can land first to make your (and my!) life simpler

src/backend/web/handlers/event.py
```diff
@@ -190,7 +207,7 @@ def event_detail(event_key: EventKey) -> Response:
     )
 
     # status_sitevar = status_sitevar_future.get_result()
-    print(f"{event.details}")
+    # print(f"{event.details}")
```
phil-lopreiato, line 210: this can just be deleted

src/backend/common/models/event_details.py
```diff
@@ -123,7 +123,7 @@ def rankings_table(self) -> Optional[List[List[str]]]:
             return None
 
         rankings = self.renderable_rankings
-        print(f"RANKINGS {rankings}")
+        # print(f"RANKINGS {rankings}")
```
phil-lopreiato, line 126: can rm

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -37,217 +149,79 @@ def build_team_mapping(
...
         for match in matches:
-            if match.comp_level != CompLevel.QM:  # only consider quals matches
+            if match.comp_level != "qm":  # only consider quals matches
```
phil-lopreiato, line 174: Use `CompLevel.QM`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -37,217 +149,79 @@ def build_team_mapping(
...
                 continue
             if played_only and not match.has_been_played:
                 continue
-            for alliance_color in ALLIANCE_COLORS:
+            for alliance_color in [AllianceColor.RED, AllianceColor.BLUE]:
```
phil-lopreiato, line 178: Use `ALLIANCE_COLORS`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -37,217 +149,79 @@ def build_team_mapping(
...
-        init_stats_default: int = 0,
-        limit_matches: Optional[int] = None,
-    ) -> "np.ndarray[int]":
+        point_accessor: Callable[[Match, AllianceColor], float],
+    ) -> "np.ndarray[float]":
         n = len(team_id_map.keys())
         s = np.zeros((n, 1))
         for match in matches:
-            if match.comp_level != CompLevel.QM:  # only consider quals matches
+            if match.comp_level != "qm":  # only consider quals matches
```
phil-lopreiato, line 197: `CompLevel.QM`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -37,217 +149,79 @@ def build_team_mapping(
...
                 continue
 
-            treat_as_unplayed = (
-                limit_matches is not None and match.match_number > limit_matches
-            )
-
-            for alliance_color in ALLIANCE_COLORS:
+            for alliance_color in [AllianceColor.RED, AllianceColor.BLUE]:
```
phil-lopreiato, line 200: `ALLIANCE_COLORS`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -37,28 +154,27 @@ def build_team_mapping(
...
-    def build_Minv_matrix(
-        cls, matches: List[Match], team_id_map: TTeamIdMap, played_only: bool = False
-    ) -> "np.ndarray[int]":
-
+    def __build_m_inv_matrix(
+        cls, matches: List[Match], team_id_map: TTeamIdMap, played_only: bool = False,
+    ) -> "np.ndarray[float]":
         n = len(team_id_map.keys())
-        M = np.zeros((n, n))
+        m = np.zeros((n, n))
```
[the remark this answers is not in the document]
tervay: Just pip guidelines but I can revert it or update the equation. No big deal

tervay:
Reviving this - things will be broken for the time being while I catch up on code changes.

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+                + match.score_breakdown[color].get("teleopCellPoints")
+            ),
+        },
+        2019: {
+            "Cargo + Panel Points": lambda match, color: (
+                match.score_breakdown[color].get("cargoPoints")
+                + match.score_breakdown[color].get("hatchPanelPoints")
+            )
+        },
+    }
```
ZachOrr, line 67: We should add a `Pre-Competition Season` section to the [New Season](https://github.com/the-blue-alliance/the-blue-alliance/wiki/New-Season) page of the runbook and make a note to update the component OPR, along with some details around how we should pick which thing(s) to show for component OPRs. Those pages are markdown files located in the `docs/` folder https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/docs/Maintainers/New-Season.md

This section needs to be added anyways for things like match breakdowns, rankings, etc. We just haven't gotten that stuff moved to py3 yet, so that section doesn't exist yet.

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+                    matches, cls.DEFAULT_COMPONENT_ACCESSOR(component), keyed=keyed
+                )
+
+        return coprs
+
+    @classmethod
+    def calculate_stat(
+        cls,
+        matches: List[Match],
+        point_accessor: Callable[[Match, AllianceColor], float],
```
ZachOrr, line 130: Can't we use the `StatAccessor` type here?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -39,26 +156,28 @@ def build_team_mapping(
...
             team_id_map[team] = i
 
         return team_list, team_id_map
 
     @classmethod
-    def build_Minv_matrix(
-        cls, matches: List[Match], team_id_map: TTeamIdMap, played_only: bool = False
-    ) -> "np.ndarray[int]":
-
+    def __build_M_inv_matrix(
```
ZachOrr, line 175: These double underscore methods should just be single underscores

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -39,26 +156,28 @@ def build_team_mapping(
         team_id_map: A dict of key: team_str, value: row index in x_matrix that corresponds to the team
         """
         # Build team list
-        team_set = set()
+        team_list = set()
```
ZachOrr, line 159: nit: Why the rename here? This is a set - the name `team_set` feels fine, right?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -72,184 +191,51 @@ def build_Minv_matrix(
                     team1_id = team_id_map[team1[3:]]
                     for team2 in alliance_teams:
                         M[team1_id, team_id_map[team2[3:]]] += 1
+
```
ZachOrr, line 194: nit: Can we back out this newline for a cleaner diff?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -72,184 +191,51 @@ def build_Minv_matrix(
...
+    def __build_s_matrix(
         cls,
         matches: List[Match],
         team_id_map: TTeamIdMap,
-        stat_type: StatType,
-        init_stats: Optional[Dict[StatType, Dict[TeamId, int]]] = None,
-        init_stats_default: int = 0,
-        limit_matches: Optional[int] = None,
-    ) -> "np.ndarray[int]":
+        point_accessor: Callable[[Match, AllianceColor], float],
```
ZachOrr, line 202: Same question here as before - can we not use the `StatAccessor` type for this?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -72,184 +191,51 @@ def build_Minv_matrix(
...
         n = len(team_id_map.keys())
         s = np.zeros((n, 1))
         for match in matches:
             if match.comp_level != CompLevel.QM:  # only consider quals matches
                 continue
 
-            treat_as_unplayed = (
-                limit_matches is not None and match.match_number > limit_matches
-            )
-
```
ZachOrr, line 96: I'm assuming this was removed because we never passed `limit_matches`, so all of the downstream code can be removed?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -72,184 +191,51 @@ def build_Minv_matrix(
...
-                    alliance_teams,
-                    init_stats,
-                    init_stats_default,
-                    treat_as_unplayed,
-                )
+
+                stat = point_accessor(match, alliance_color)
                 for team in alliance_teams:
-                    s[team_id_map[team]] += stat
+                    s[team_id_map[team]] += float(stat)
```
ZachOrr, line 217: nit: We shouldn't need to cast here, if the return type for this specifies it's a float. Why are we casting to a float?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -72,184 +191,51 @@ def build_Minv_matrix(
...
         matches: List[Match],
         team_list: List[TeamId],
         team_id_map: TTeamIdMap,
-        Minv: Any,
-        stat_type: StatType,
-        init_stats: Optional[Dict[StatType, Dict[TeamId, int]]] = None,
-        init_stats_default: int = 0,
-        limit_matches: Optional[int] = None,
+        M_inv: np.ndarray,
+        point_accessor: Callable[[Match, AllianceColor], float],
```
ZachOrr, line 228: Can we use the `StatAccessor` type for this?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -72,184 +191,51 @@ def build_Minv_matrix(
...
-        stats: EventMatchStats = {
-            StatType.OPR: oprs_dict,
-            StatType.DPR: dprs_dict,
-            StatType.CCWM: ccwms_dict,
-        }
-
-        return stats
-
-    @classmethod
-    def get_last_event_stats(
```
ZachOrr, line 208: For removing this method - do we just not use this anywhere? I see it was being used in py2 in `calculate_matchstats` - did it get dropped somewhere during the py3 move? https://github.com/the-blue-alliance/the-blue-alliance/blob/1d05f3bb147b26e228efe40801f13a0692a97153/helpers/matchstats_helper.py#L152

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+    def calculate_coprs(
+        cls, matches: List[Match], keyed: bool = True
+    ) -> Dict[Component, TeamStatMap]:
+        coprs = OrderedDict()
+
+        # If there is not valid data for COPRs, skip
+        if not (
+            matches is not None
+            and len(matches) > 0
+            and matches[0].score_breakdown is not None
```
ZachOrr, line 102: Can we DRY references to this `matches[0]`? It probably means we have to split this check in to two checks - the first one to make sure that `matches` is non-zero, do the assignment, and then to check that the `score_breakdown is not None`

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+
+    @classmethod
+    def calculate_coprs(
+        cls, matches: List[Match], keyed: bool = True
+    ) -> Dict[Component, TeamStatMap]:
+        coprs = OrderedDict()
+
+        # If there is not valid data for COPRs, skip
+        if not (
+            matches is not None
```
ZachOrr, line 100: nit: We shouldn't need to have this `is not None` check with the typing, but I don't care if it sticks around. Maybe it's better to be safer when possible.

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+                coprs[title] = cls.calculate_stat(matches, accessor, keyed=keyed)
+
+        # for each string-like key in the score_breakdown object
+        # (just take the red score_breakdown from match 0, it's an arbitrary selection)
+        for component, value in none_throws(matches[0].score_breakdown)[
+            AllianceColor.RED
+        ].items():
+            # if the component is an int or a float, we can compute it easily
+            # (as opposed to, say, a string)
+            if isinstance(value, int) or isinstance(value, float):
```
ZachOrr, line 119: One thought here is we could attempt to cast to a float, and if our cast succeeds, we're good to go. Otherwise we can't count it. That way we can support more types (like strings) without needing to add explicit support.

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+            return {}
+
+        M_inv = cls.__build_M_inv_matrix(matches, team_id_map, played_only=True)
+
+        return {
+            f"frc{k}" if keyed else k: v
+            for k, v in cls.__calc_stat(
+                matches, team_list, team_id_map, M_inv, point_accessor
+            ).items()
+        }
```
ZachOrr, line 147: Can we get a comment around this code? I'm super unclear what this is doing.

src/backend/web/local/bootstrap.py
```diff
@@ -132,6 +133,22 @@ def update_event(cls, key: EventKey, auth_token: str) -> None:
...
+
+        # if any([v != {} for v in matchstats_dict.values()]):
+        #     pass
+        # else:
+        #     matchstats_dict = None
+        #     print("Convert the following to a warning!")
+        #     print(f"Matchstat calculation for {event.key} failed!")
+
+        cls.store_eventdetail(event, "matchstats", matchstats_dict)
+
```
ZachOrr, line 151: Should this be uncommented?

src/backend/web/local/tests/bootstrap_test.py
```diff
@@ -249,12 +275,19 @@ def test_bootstrap_event(ndb_context, requests_mock: RequestsMocker) -> None:
         year=2020,
         team_list=[],
     )
+    matchstats = EventMatchstats(
+        oprs=make_team_stat_map(),
+        dprs=make_team_stat_map(),
+        ccwms=make_team_stat_map(),
+        coprs={},
```
ZachOrr, line 282: Do we want a test that includes `coprs` as well to make sure that round-trips properly?

src/backend/web/handlers/event.py
```diff
@@ -129,13 +131,30 @@ def event_detail(event_key: EventKey) -> Response:
...
+            event.matchstats["oprs"].items(), key=lambda t: -t[1]
+        )[:num_matchstats]
+        matchstats["DPRs"] = sorted(
+            event.matchstats["dprs"].items(), key=lambda t: -t[1]
+        )[:num_matchstats]
+        matchstats["CCWMs"] = sorted(
+            event.matchstats["ccwms"].items(), key=lambda t: -t[1]
+        )[:num_matchstats]
+
+        if "coprs" in event.matchstats:
```
ZachOrr, line 147: Can we replace this `"coprs"` string with `StatType.COPR.value`?

src/backend/web/handlers/event.py
```diff
@@ -129,13 +131,30 @@ def event_detail(event_key: EventKey) -> Response:
...
-        if (event.matchstats is not None and "oprs" in event.matchstats)
-        else []
-    )
-    oprs = sorted(oprs, key=lambda t: t[1], reverse=True)  # sort by OPR
-    oprs = oprs[:15]  # get the top 15 OPRs
+    num_matchstats = 15
+    matchstats = collections.OrderedDict()
+    if event.matchstats is not None:
+        matchstats["OPRs"] = sorted(
+            event.matchstats["oprs"].items(), key=lambda t: -t[1]
```
ZachOrr, line 138: Can we replace these strings with `StatType.XXX.value`?

src/backend/web/handlers/event.py
```diff
@@ -129,13 +131,30 @@ def event_detail(event_key: EventKey) -> Response:
...
+        )[:num_matchstats]
+        matchstats["DPRs"] = sorted(
+            event.matchstats["dprs"].items(), key=lambda t: -t[1]
```
ZachOrr, line 141: Is there a way for us to DRY this lambda sort?

src/backend/web/templates/event_partials/event_matchstat_table.html
```diff
@@ -0,0 +1,67 @@
...
+
+            const coprTd = document.createElement('td');
+            coprTd.innerText = e[1].toFixed(2); // round to 2 decimal places
+
+            const tr = document.createElement('tr');
+            tr.appendChild(teamTd);
+            tr.appendChild(coprTd);
+            document.getElementById('coprTableBody').appendChild(tr);
+        });
+    }
```
ZachOrr, line 51: Why do we need to do this in Javascript? Can we really not do this with Jinja?

src/backend/web/templates/event_partials/event_matchstat_table.html
```diff
@@ -0,0 +1,67 @@
...
+
+        {% for component, component_id in matchstat_dropdown_id_map.items() %}
+            $("#dropdown_{{ component_id }}").click(function(e) {
+               e.preventDefault();
+               buildCoprTable("{{ component }}");
+               console.log("{{ component }}")
+            });
+        {% endfor %}
+
+        buildCoprTable("OPRs");
```
ZachOrr, line 63: Should this bit be removed?

src/backend/common/helpers/matchstats_helper.py
```diff
@@ -8,29 +8,146 @@
...
+            "Cargo + Panel Points": lambda match, color: (
+                match.score_breakdown[color].get("cargoPoints")
+                + match.score_breakdown[color].get("hatchPanelPoints")
+            )
+        },
+    }
+
+    @classmethod
+    def calculate_matchstats(
+        cls, matches: List[Match], skip_coprs: bool = False, keyed: bool = True
```
ZachOrr, line 71: The only place I'm seeing us pass this `keyed` variable is in tests. I feel like we should just always key these? Removing the complexity of this method feels helpful.

src/backend/common/queries/dict_converters/event_details_converter.py
```diff
@@ -27,15 +25,6 @@ def eventsDetailsConverter_v3(cls, event_details: List[EventDetails]):
 
     @classmethod
     def eventDetailsConverter_v3(cls, event_details: EventDetails) -> EventDetailsDict:
-        normalized_oprs = defaultdict(dict)
-        if event_details and event_details.matchstats:
-            for stat_type, stats in event_details.matchstats.items():
-                if stat_type in {"oprs", "dprs", "ccwms"}:
-                    for team, value in cast(Dict[TeamKey, float], stats).items():
-                        if "frc" not in team:  # Normalize output
-                            team = "frc{}".format(team)
```
ZachOrr, line 36: Are we worried about removing this normalization? I'm assuming there's a reason this code exists. Especially with the `keyed` optional method parameter.

src/backend/common/queries/dict_converters/event_details_converter.py
```diff
@@ -52,7 +41,7 @@ def eventDetailsConverter_v3(cls, event_details: EventDetails) -> EventDetailsDi
             "insights": event_details.insights
             if event_details
             else {"qual": {}, "playoff": {}},
-            "oprs": normalized_oprs if normalized_oprs else {},  # OPRs, DPRs, CCWMs
+            "oprs": event_details.matchstats,
```
ZachOrr, line 44: Will this always be non-null, or do we need to keep this check?

ZachOrr (changes requested):
This is exciting!

src/backend/common/helpers/matchstats_helper.py
tervay: No idea how that happened, honestly.

src/backend/common/helpers/matchstats_helper.py
tervay: The reason was because some stats are returns as `int`s, but I'll change that behavior elsewhere to obey the typing better. (By making the default StatAccessor guarantee a float return)

src/backend/common/helpers/matchstats_helper.py
tervay: It's actually an unused variable even in the code you linked, so I'm guessing someone removed that unused variable in the transition, but then the method became unused across the repo.

src/backend/common/helpers/matchstats_helper.py
tervay: Correct, yeah. I can put it back in if you'd want it for the future, but it's never used anywhere.

src/backend/web/handlers/event.py
tervay: For `event.matchstats["oprs"]` and the like, yes. For `matchstats["OPRs"]`, it will affect the template:

3o9A4pV

src/backend/web/templates/event_partials/event_matchstat_table.html
tervay: This is the default COPR table rendered when the user initially clicks on the Insights tab.

src/backend/web/templates/event_partials/event_matchstat_table.html
tervay: Truthfully, this is just how I've done it on my own projects in the past. I'm not sure how I would do it with templating. I think this is the easiest approach, but I am open to other methods of doing it if you have ideas towards it.

src/backend/common/queries/dict_converters/event_details_converter.py
tervay: I've changed match stats to always return keyed values. I believe that should be sufficient, but I can add logic back into here if you would like.

src/backend/common/queries/dict_converters/event_details_converter.py
tervay: Truthfully I don't know enough to say it will be always non-null - however, the previous check didn't make a ton of sense to me, as `normalized_oprs` was always non-null anyway: https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/src/backend/common/queries/dict_converters/event_details_converter.py#L30

src/backend/common/helpers/matchstats_helper.py
tervay: Could you clarify more?

src/backend/web/local/bootstrap.py
tervay: Leftover debugging code. We need to calculate locally since prod does not calculate component OPRs (yet).

docs/Maintainers/New-Season.md
```diff
@@ -1,3 +1,7 @@
+## Pre-Competition Season
+
```
ZachOrr, line 2: nit: Let's have a `Pre-Competition Season` h2 and then underneath an `Update Component OPRs` h3

docs/Maintainers/New-Season.md
```diff
@@ -1,3 +1,7 @@
...
+Within `src/backend/common/helpers/matchstats_helper.py`, add a new entry `2022: {}` (replace `2022` with the given upcoming year) to `COMPONENTS` for future custom component OPRs.
```
ZachOrr, line 3: Can we add a note about adding custom components as well? Just like - custom components can have human-readable names, can be calculated based on score breakdowns, etc. In 2020 it made sense to have this component OPR because XYZ

src/backend/common/helpers/matchstats_helper.py
ZachOrr: Yeah! So we reference `matches[0]` 3 different times in this method. Why not write this as...

```
if not (
    matches is not None
    and len(matches) > 0
):
    return coprs

match = matches[0]
if match.score_breakdown is None:
    return coprs
```

and then later on in the code we can do things like -

```
year = match.year
```
```
for component, value in none_throws(match.score_breakdown):
```

It's not a huge thing - just a DRY thing.

ZachOrr: Also realizing this conditional is written pretty weird. We can probably write it like -

```
if matches is None or len(matches) == 0:
```

The length check can also be replaced with `not matches`, but I'm unsure if that's clearer or just more confusing

src/backend/common/queries/dict_converters/event_details_converter.py
ZachOrr: Confirmed in our database that this CAN be null. My quick spot-check was `2000cmp`, which is null. So yes - let's make sure to add a null check here.

src/backend/web/local/bootstrap.py
```diff
@@ -132,6 +133,14 @@ def update_event(cls, key: EventKey, auth_token: str) -> None:
         event_awards = cls.fetch_event_detail(key, "awards", auth_token)
         list(map(lambda t: cls.store_award(t, event), event_awards))
 
+        # Get insights from prod API
+        # event_oprs = cls.fetch_event_detail(key, "oprs", auth_token)
+        # cls.store_eventdetail(event, "matchstats", event_oprs)
```
ZachOrr, line 138: Let's go ahead and remove this if it's not being used

src/backend/web/templates/event_partials/event_matchstat_table.html
ZachOrr: Lemme take a pass at this later. We can always merge this as-is (I'm not opposed to this approach - I'd just rather do it in templating)

tervay:
At this point the branch is so extremely outdated that I don't feel its worth going through rebase conflicts. Opening a new PR here: https://github.com/the-blue-alliance/the-blue-alliance/pull/4847

PR #2805, opened 2020-05-30
[py3] Add jinja2-filter tests

ZachOrr:
Just adding some missing tests. Note - the `register_template_filters` function has changed a little bit.

### DIFF
--- a/src/web/tests/conftest.py
+++ b/src/web/tests/conftest.py
@@ -3,5 +3,5 @@ import pytest
 
 
 @pytest.fixture
-def app() -> Flask:
+def empty_app() -> Flask:
     return Flask(__name__)
--- a/stubs/_pytest/mark/__init__.pyi
+++ b/stubs/_pytest/mark/__init__.pyi
@@ -1,1 +1,2 @@
 from .structures import Mark as Mark, MarkDecorator as MarkDecorator, MarkGenerator as MarkGenerator, get_empty_parameterset_mark as get_empty_parameterset_mark
+from .structures import MARK_GEN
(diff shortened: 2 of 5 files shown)


stubs/_pytest/mark/__init__.pyi
```diff
@@ -4,3 +4,4 @@ from .structures import (
     MarkGenerator as MarkGenerator,
     get_empty_parameterset_mark as get_empty_parameterset_mark,
 )
+from .structures import MARK_GEN
```
ZachOrr, line 7: cc @phil-lopreiato getting this error when running pyre
Screen Shot 2020-05-30 at 1 06 45 PM

Weirdly when I do a stubgen I get wildly different generated files then what's in the repo. Are we using different versions of things?

```
$ pytest --version
This is pytest version 5.4.2, imported from /Users/zach/Desktop/the-blue-alliance/venv/lib/python3.7/site-packages/pytest/__init__.py
setuptools registered plugins:
  pytest-cov-2.9.0 at /Users/zach/Desktop/the-blue-alliance/venv/lib/python3.7/site-packages/pytest_cov/plugin.py
$ mypy --version
mypy 0.770
```

src/web/tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+from flask import Flask
+import pytest
+
+
+@pytest.fixture
+def app() -> Flask:
```
phil-lopreiato, line 6: nit: name this something other than just `app`? Maybe `empty_app`?

stubs/_pytest/mark/__init__.pyi
phil-lopreiato: seems like this is resolved in the latest push?

ZachOrr: It's not :\

PR #2809, opened 2020-05-30
[py3] Cleanup ops scripts in to more logical folders/names

ZachOrr:
Alright two things I need buy-in on -

First, cc @fangeugene I renamed all the `run_xxx.sh` scripts to be what they do, like `lint.sh`, `format.sh`, etc. I figured naming them after the action they're taking makes more sense then naming them after the technology we're using. That way we can make the scripts extendable and still refer to the 3/4 different things we do

Second, cc @phil-lopreiato I moved around a lot of the `ops/` scripts in to logical folders. Things we only run when setting up Vagrant are in the `ops/dev/vagrant` folder, things we only run when setting up Docker are in the `ops/dev/docker` folder, and things we run for the container or in multiple places are in the `ops/dev` catchall. Any build-related things (building for deployment/hosting) are in `ops/build` folder.

I nuked `vars.sh` since I don't think we need any of it yet - feels okay to add back in once we need it and make sure we're specific about where we're using it


.github/workflows/pull_request.yml
```diff
@@ -27,7 +27,7 @@ jobs:
           pip install -r requirements.txt
           pip install -r src/requirements.txt
       - name: Run Pyre
-        run: ./ops/run_pyre.sh
+        run: ./ops/type_check.sh
```
phil-lopreiato, line 30: update the name to "Type Check"?

.github/workflows/push.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           pip install -r src/requirements.txt
       - name: Run Pyre
```
phil-lopreiato, line 29: ditto

.github/workflows/pull_request.yml
ZachOrr: Oops good catch!

PR #2815, opened 2020-06-03
[py3] Add Setup and Dev Runbook docs

ZachOrr:
Updated the setup instructions for the new py3 environment, as well as a few more QOL things. I've also added a "Dev Runbook" page, which I'd like to use as a bit of a catch-all for one-off things we end up doing from time to time that should be documented for other maintainers.


ZachOrr:
Gonna close and wait for the wiki stuff to get merged, then merge these in

PR #2823, opened 2020-06-05
[py3] Move existing sitevars from py2 to py3

ZachOrr:
This looks like it's going to require a few changes in IMDB (are we calling it that?)

We'll need to implement `BeginTransaction`. I'm also hitting [this assertion](https://github.com/phil-lopreiato/google-cloud-datastore-stub/blob/9d7e2c6110b89d954c23ad731cb0207032d044cd/InMemoryCloudDatastoreStub/datastore_stub.py#L203), I believe from the `get_or_insert` which isn't doing a get if it already exists, just an insert


src/backend/common/models/sitevar.py
```diff
@@ -1,6 +1,6 @@
 import json
 
-from google.appengine.ext import ndb
+from google.cloud import ndb
 
 
 class Sitevar(ndb.Model):
```
phil-lopreiato, line 6: I wonder if we can also subclass from ABC here to force new SVs to define a schema 

https://docs.python.org/3/library/abc.html

ZachOrr: Sitevar models (like `WebsiteBlacklist`) don't actually subclass `Sitevar`, they just wrap `Sitevar`.

PR #2824, opened 2020-06-07
[py3] [nodeploy] Update wiki auto-updating

dracco1993:
## Description
Update the self-updating wiki to not use a `git` submodule.

## Motivation and Context
After I thought about my current implementation a bit more, I realized that we 100% don't actually need the wiki submodule at all.  Instead of having it in the repo, which has to then be cloned down, and changed for forks (if testing), we could just clone the wiki repo at the time the GitHub Action is run.  This cleans up the repo, and doesn't add any more time/complexity to the Action.

## How Has This Been Tested?
Passing build with this code:
https://github.com/dracco1993/the-blue-alliance/runs/738962512?check_suite_focus=true

Currently running at:
https://github.com/dracco1993/the-blue-alliance/tree/py3
https://github.com/dracco1993/the-blue-alliance/wiki

## Types of changes

### DIFF
--- a/.github/workflows/update_wiki.yml
+++ b/.github/workflows/update_wiki.yml
@@ -4,8 +4,8 @@ on:
   push:
     branches:
       - py3
-    paths:
-      - "docs/**"
+    # paths:
+    #   - "docs/**"
 
 jobs:
   update-wiki:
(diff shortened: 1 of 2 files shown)


phil-lopreiato:
yup, makes sense to me. Nice!

PR #2834, opened 2020-06-10
[py3] Start implementing APIv3

fangeugene:
- Adds decorator for API auth (hook into DB in a future PR)
- Adds decorator for caching (hook into Redis/etc. in a future PR)
- Adds APIv3 team routes
- Query model & convert to dict
- Key name validation


src/backend/common/queries/dict_converters/converter_base.py
```diff
@@ -0,0 +1,58 @@
+import abc
+from backend.common.queries.types import QueryReturn
+from typing import Callable
+
+
+class ConverterBase(object):
```
phil-lopreiato, line 6: try making this 
```
from typing import Generic
class ConverterBase(Generic[QueryReturn])
```
```

File: src/backend/common/queries/dict_converters/team_converter.py [hunk 2]
```diff
@@ -0,0 +1,35 @@
+from backend.common.queries.dict_converters.converter_base import ConverterBase
+
+
+class TeamConverter(ConverterBase):
```
phil-lopreiato, line 4: And this would be 
```
class TeamConverter(ConverterBase[Team])
```

src/backend/common/queries/team_query.py
```diff
@@ -1,19 +1,22 @@
...
+
     @ndb.tasklet
     def _query_async(self, team_key: str) -> TypedFuture[Team]:
         team = yield Team.get_by_id_async(team_key)
         raise ndb.Return(team)
 
 
 class TeamListQuery(DatabaseQuery[List[Team]]):
-
+    DICT_CONVERTER = TeamConverter
```
phil-lopreiato, line 19: you'd probably also need to make a `TeamListConverter` that delegates to `TeamConverter` to make the type checker happy

src/backend/common/queries/database_query.py
```diff
@@ -1,16 +1,16 @@
...
-
-
-QueryReturn = TypeVar("QueryReturn")
+from typing import Generic
 
 
 class DatabaseQuery(abc.ABC, Generic[QueryReturn]):
 
     _query_args: int
+    DICT_CONVERTER: ConverterBase
```
phil-lopreiato, line 13: this would be
```
    DICT_CONVERTER: ConverterBase[QueryReturn]
```

src/backend/common/queries/dict_converters/converter_base.py
```diff
@@ -0,0 +1,58 @@
...
+        else:
+            return cls._delistify(converted_thing)
+
+    @classmethod
+    @abc.abstractmethod
+    def _convert(cls, thing: QueryReturn, version: int) -> Callable:
+        ...
+
+    @classmethod
+    def _listify(cls, thing):
```
phil-lopreiato, line 21: We may need to have separate type hinted classes for lists?

Or, we'd need to maybe make the generic type here

```
Union[QueryReturn, List[QueryReturn]]
```

phil-lopreiato: I think I'd prefer separate classes for list types (since they have their own `QueryReturn` types as well), we'd just have to make the base class aware enough?

src/backend/common/queries/database_query.py
```diff
@@ -29,3 +30,15 @@ def fetch_async(self) -> TypedFuture[QueryReturn]:
             # Type-hinting the tasklet decorator is hard, but it consumes
             # the generator and returns the overall future
             return query_result  # pyre-ignore
+
+    def fetch_dict(self, version: int) -> dict:
```
phil-lopreiato, line 34: nit: `typing.Dict` here

phil-lopreiato: Also nit - should we have an 
```
class ApiMajorVersion(enum.IntEnum):
  API_V3 = 3
```

so keep this to a know set of major versions? 

and you can then typehint this to `version: ApiMajorVersion`

src/backend/common/queries/database_query.py
```diff
@@ -29,3 +30,15 @@ def fetch_async(self) -> TypedFuture[QueryReturn]:
...
+        return self.fetch_dict_async(version).get_result()
+
+    @ndb.tasklet
+    def fetch_dict_async(self, version: int) -> TypedFuture[dict]:
```
phil-lopreiato, line 38: same 2 nits

src/backend/common/queries/database_query.py
```diff
@@ -29,3 +30,15 @@ def fetch_async(self) -> TypedFuture[QueryReturn]:
...
+        query_result = yield self.fetch_async()
+        if query_result is None:
+            raise DoesNotExistException()
+        # Type-hinting the tasklet decorator is hard, but it consumes
+        # the generator and returns the overall future
+        return self.DICT_CONVERTER(query_result).convert(version)  # pyre-ignore
```
phil-lopreiato, line 44: is this ignore needed? We can look into better typing here later if it is...

I think we need to make some improvements to better handle the generators here, but I'm not quite sure what yet

src/backend/common/queries/dict_converters/converter_base.py
```diff
@@ -0,0 +1,63 @@
...
+from backend.common.profiler import Span
+
+
+class ConverterBase(Generic[QueryReturn]):
+    _query_return: QueryReturn
+
+    def __init__(self, query_return: QueryReturn):
+        self._query_return = query_return
+
+    def convert(self, version: int) -> Union[None, dict, List[dict]]:
```
phil-lopreiato, line 14: `typing.Dict`

src/backend/api/handlers/decorators.py
```diff
@@ -0,0 +1,39 @@
+from flask import request
+from functools import wraps
+
+from backend.common.models.team import Team
+from backend.common.queries.exceptions import DoesNotExistException
+
+
+def api_authenticated(func):
```
ZachOrr, line 8: We should be able to type these decorators by using `Callable` https://docs.python.org/3/library/typing.html#typing.Callable - I'm not super sure about how the args/kwargs works for typing though

ZachOrr: At minimum we should be able to add none return types or something right?

src/backend/api/main.py
```diff
@@ -1,9 +1,14 @@
...
+from backend.api.handlers.team import team, team_list
 from backend.common.middleware import install_middleware
 
 
 app = Flask(__name__)
 install_middleware(app)
+app.config["JSONIFY_PRETTYPRINT_REGULAR"] = True
 
-app.add_url_rule("/api/<path:path>", view_func=root)
+app.add_url_rule("/api/v3/team/<team_key>", view_func=team)
```
ZachOrr, line 11: We should be able to use `str:team_key` here (not a big deal though)

src/backend/common/decorators.py
```diff
@@ -0,0 +1,28 @@
+from flask import request, make_response, Response
+from functools import partial, wraps
+
+
+def cached_public(func=None, timeout: int = 61):
```
ZachOrr, line 5: Same comment here for typing with `Callable`

src/backend/common/queries/database_query.py
ZachOrr, line 34: I'm not sure I understand the `dict` return type bit - I thought `Dict` was what we used with types right?

ZachOrr: Like I'm sure this is a type somehow I just don't see where 😅

src/backend/common/queries/dict_converters/converter_base.py
```diff
@@ -0,0 +1,63 @@
...
+from typing import Generic, List, Union
+
+from backend.common.queries.types import QueryReturn
+from backend.common.profiler import Span
+
+
+class ConverterBase(Generic[QueryReturn]):
+    _query_return: QueryReturn
+
+    def __init__(self, query_return: QueryReturn):
```
ZachOrr, line 11: `__init__` methods can have `None` returns

src/backend/common/queries/dict_converters/team_converter.py
```diff
@@ -0,0 +1,35 @@
+from typing import List
+
+from backend.common.models.team import Team
+from backend.common.queries.dict_converters.converter_base import ConverterBase
+
+
+class TeamConverter(ConverterBase):
+    SUBVERSIONS = {  # Increment every time a change to the dict is made
```
ZachOrr, line 8: We can actually type these as well to be `Dict[int, int]`

src/backend/common/queries/dict_converters/converter_base.py
```diff
@@ -0,0 +1,63 @@
+import abc
+from typing import Generic, List, Union
+
+from backend.common.queries.types import QueryReturn
+from backend.common.profiler import Span
+
+
+class ConverterBase(Generic[QueryReturn]):
+    _query_return: QueryReturn
+
```
phil-lopreiato, line 10: Maybe we can promote the versioning to first class?

```
SubversionMap = TypedDict("SubversionMap", {f"{v}": int for v in list(ApiMajorVersion)})

@abc.abstractproperty
SUBVERSIONS: SubversionMap
```

And this way we can make the typechecker tell us everywhere we'd need to add a new set of versions if/when we add apiv4

PR #2839, opened 2020-06-13
Move Account and MobileClient from py2 to py3

ZachOrr:
Moving Account and MobileClient in preparation for moving TBANS code to test tasks/queues


src/backend/common/queries/mobile_client_query.py
```diff
@@ -0,0 +1,34 @@
...
+from typing import List
+
+
+class MobileClientListQuery(DatabaseQuery[List[MobileClient]]):
+    @ndb.tasklet
+    def _query_async(
+        self, users: List[str], client_types: List[ClientType] = list(ClientType)
+    ) -> TypedFuture[List[MobileClient]]:
+        if not users or not client_types:
+            return []
```
phil-lopreiato, line 15: I have a fix for `pyre` complaining about this, will be up shortly

src/backend/common/queries/mobile_client_query.py
```diff
@@ -0,0 +1,34 @@
...
+
+        client_types = [c.value for c in client_types]
```
phil-lopreiato, line 17: Is this needed? It may need to be a separate variable?

ZachOrr: TIL apparently not 😲

phil-lopreiato: yeah, since `IntEnum` extends `int`, it should be implicitly convertable

PR #2853, opened 2020-06-21
[py3] Cache Clearing Base

fangeugene:
Adds a set of `_<model>_affected_queries` methods that returns the set of `DatabaseQueries` affected by the change.

Later, model manipulators will call all of the `_*_affected_queries` methods of all the `DatabaseQueries` whenever a model is updated to clear appropriate caches.

### DIFF
--- a/src/backend/common/queries/database_query.py
+++ b/src/backend/common/queries/database_query.py
@@ -1,5 +1,5 @@
 import abc
-from typing import Any, Dict, Generic, Type
+from typing import Any, Dict, Generic, Optional, Set, Type
 
 from google.cloud import ndb
 from pyre_extensions import safe_cast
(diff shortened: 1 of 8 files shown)


src/backend/common/queries/database_query.py
```diff
@@ -13,12 +13,30 @@
...
+    CACHE_KEY_FORMAT: str = ""
+    CACHE_VERSION: int = 0
     DICT_CONVERTER: Type[ConverterBase[QueryReturn]] = ConverterBase
 
     def __init__(self, *args, **kwargs):
         self._query_args = kwargs
 
+    @property
+    def cache_key(self) -> str:
+        if self._cache_key == "":
```
ZachOrr, line 31: Can we do `if not self._cache_key` here? Or is actually comparing to an empty string important? The other bit would be to just leave `_cache_key` as an uninitialized variable

src/backend/common/queries/database_query.py
```diff
@@ -46,3 +64,19 @@ def fetch_dict_async(self, version: ApiMajorVersion) -> TypedFuture[Dict]:
             safe_cast(QueryReturn, query_result)
         ).convert(version)
         return safe_cast(TypedFuture[Dict], res)
+
+    @classmethod
+    def _event_affected_queries(
```
ZachOrr, line 69: I don't love this approach but I also don't have any better ideas so I'm not going to push back

PR #3066, opened 2020-10-03
[py3] Allow independent setup of GOOGLE_APPLICATION_CREDENTIALS env variable in tba_dev_config

ZachOrr:
This PR comes out of some discussions we've had in Slack around setting up the `GOOGLE_APPLICATION_CREDENTIALS` independent of the remote datastore. This also sets up the local/remote configuration for Google Cloud Tasks.

This should also make [setting up Auth locally](https://github.com/the-blue-alliance/the-blue-alliance/wiki/Firebase-Setup#configure-authentication) easier by allowing contributors to set their `GOOGLE_APPLICATION_CREDENTIALS` without having to ssh in to the Vagrant image

### DIFF
--- a/ops/dev/vagrant/dev_appserver.sh
+++ b/ops/dev/vagrant/dev_appserver.sh
@@ -8,7 +8,7 @@ function get_config_prop {
         local dev_config_file="tba_dev_config.local.json"
     fi
 
-    jq -r -s ".[0] * (.[1] // {}) | .$prop_name" tba_dev_config.json $dev_config_file
+    jq -r -s ".[0] * (.[1] // {}) | .$prop_name | select (.!=null)" tba_dev_config.json $dev_config_file
 }
 
 function get_project_from_key {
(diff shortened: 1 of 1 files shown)


ops/dev/vagrant/dev_appserver.sh
```diff
@@ -24,18 +24,34 @@ datastore_args=""
...
+    cred_file=$(realpath $google_application_credentials)
+    project=$(get_project_from_key "$cred_file")
+    application="--application $project"
+    env="$env --env_var GOOGLE_APPLICATION_CREDENTIALS=$cred_file"
+else
+    application="--application test"
+fi
+
+function assert_google_application_credentials {
+    if [[ "$google_application_credentials" = "null" ]]; then
```
ZachOrr, line 39: cc @phil-lopreiato Is there a better way to do this? I was trying to see if we could have `get_config_prop` return a proper empty string instead of a `null` but I have no idea what I'm doing

phil-lopreiato: This seems fine, although if you'd rather make the `get_config_prop` return an empty string if the key is not found, you could add a `| select (.!=null)` in the jq query, making it 

```
jq -r -s ".[0] * (.[1] // {}) | .$prop_name | select (.!=null)" tba_dev_config.json $dev_config_file
```

PR #3162, opened 2020-11-21
Support Google Cloud Tasks

ZachOrr:
## Description
This PR adds the base for deferred tasks in the py3 codebase for both remote and local tasks.

## Motivation and Context
As part of the Python 2 -> Python 3 migration, the [`google.appengine.ext.deferred`](https://cloud.google.com/appengine/docs/standard/python/refdocs/google.appengine.ext.deferred.deferred) library has been deprecated. We planned to move from deferred to the new [Cloud Tasks](https://cloud.google.com/appengine/docs/standard/python/taskqueue/push/migrating-push-queues) in https://github.com/the-blue-alliance/the-blue-alliance/issues/2663.

Redis/rq is used in the development environment in an attempt mirror the Cloud Taks upstream environment. This will not be a perfect 1:1, but discrepancies will be called out.

## How Has This Been Tested?
Mostly manual testing - making sure tasks show up in queues and get executed properly, etc.

## TODO
- [x] Tests
- [x] Add RedisTaskQueue url reformatting
- [x] Add docs for setting up services to use new deferred handling (similar to existing deferred handling)
- [x] Add docs with information around setting up queues via the command line, etc.
- [x] Add docs around monitoring local + remote tasks
- [x] Support calling tasks "on services" locally
- [x] File a tracking ticket to support `countdown`, `eta`, `name`, `retry_options` when enqueueing a task
- [x] Investigate auth mechanism for `handle_defer`

PR #3180, opened 2020-11-27
Add Interface to Replace Legacy Memcache

phil-lopreiato:
This PR copies the legacy GAE memcache interface (ish, limited to the feature we currently use elsewhere in the code).

We add an interface so we can have multiple implementations. Namely, a noop one that will miss 100% of the time. This will be our default client for "raw" memcache, for cases where we can't/don't run redis locally.

~Still todo: write some tests, and refactor redis client creation so we can create a single one and share it with the ndb cache~

### DIFF
--- a/src/backend/common/redis.py
+++ b/src/backend/common/redis.py
@@ -1,0 +1,24 @@
+import os
+from typing import Optional
+
+import redis
+from pyre_extensions import none_throws
+
+
+class RedisClient:
+    """
+    This will let us share redis clients across use cases in the code
+    """
+
+    client: Optional[redis.Redis] = None
+
+    @classmethod
+    def get(cls) -> Optional[redis.Redis]:
+        redis_url = os.environ.get("REDIS_CACHE_URL")
+        if not redis_url:
+            return None
+
+        if cls.client is None:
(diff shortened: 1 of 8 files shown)


src/backend/common/redis.py
```diff
@@ -0,0 +1,24 @@
...
+class RedisClient:
+    """
+    This will let us share redis clients across use cases in the code
+    """
+
+    client: Optional[redis.Redis] = None
+
+    @classmethod
+    def get(cls) -> Optional[redis.Redis]:
+        redis_url = os.environ.get("REDIS_CACHE_URL")
```
ZachOrr, line 17: Thoughts on putting this in the `Environment` and then pulling it from there? I've been trying to sub out any os.environ` stuff to all live in `Environment`, and then pass the `Environment` variables in via inits so things are less magic https://github.com/the-blue-alliance/the-blue-alliance/pull/3162/files#diff-23ddd72bfb97c613cc2ea8396a5daf356f1381d836b168cd8525909866cbc30bR35

phil-lopreiato: :+1:

PR #3227, opened 2020-12-29
Configure Redis in Prod

phil-lopreiato:
After defeating my own stupidity and fixing the incorrect port (h/t @fletch3555), let's get redis working in prod.

https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-to-cloud-ndb#caching is the guide of what we configured on the server side (I'll also add some wiki docs to this pull before merging)

In short, we need to configure two things:
 - the `REDIS_CACHE_URL` environment variable for our application code to pick up the fact we want to use redis (this is the same way we get the url for the dev app)
  - a VPC network connector to tell Google Cloud that our GAE instance is allowed to talk to the network containing the redis resources

Both of these things are to be configured in the service YAML files. The downside is, we don't really want anything tied to the prod deployment committed in the repo, so we need to dynamically generate those two items.

Enter the service postprocessing script that is our new deploy entry point (I wrapped that and the `gcloud` invocation in a simple bash script). This python script will read those two values from secrets and add them into the YAML file at deploy time

### DIFF
--- a/docs/Deployment/Prod-Setup.md
+++ b/docs/Deployment/Prod-Setup.md
@@ -1,0 +1,13 @@
+This document lists the items that need to be configured on production Google Cloud in order for The Blue Alliance to run.
+
+## Task Queues
+
+Task queues need to be created in advance. See [this section](https://github.com/the-blue-alliance/the-blue-alliance/wiki/Queues-and-defer#creating-queues) for the queues we need to set up.
+
+### Memorystore (redis)
+
+Roughly, the setup is outlined in [this page](https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-to-cloud-ndb#caching). We need the following steps:
+ 1. Create a [VPC Network](https://cloud.google.com/vpc/docs/vpc)
+ 2. Create a [VPC Access Connector](https://cloud.google.com/vpc/docs/configure-serverless-vpc-access#creating_a_connector)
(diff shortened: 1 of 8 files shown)


docs/Deployment/Prod-Setup.md
```diff
@@ -0,0 +1,13 @@
...
+
+Task queues need to be created in advance. See [this section](https://github.com/the-blue-alliance/the-blue-alliance/wiki/Queues-and-defer#creating-queues) for the queues we need to set up.
+
+### Memorystore (redis)
+
+Roughly, the setup is outlined in [this page](https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3/migrate-to-cloud-ndb#caching). We need the following steps:
+ 1. Create a [VPC Network](https://cloud.google.com/vpc/docs/vpc)
+ 2. Create a [VPC Access Connector](https://cloud.google.com/vpc/docs/configure-serverless-vpc-access#creating_a_connector)
+ 3. Create a [Redis Instance](https://cloud.google.com/memorystore/docs/redis/creating-managing-instances#creating_redis_instances) using the network
+ 4. Make note of the `REDIS_CACHE_URL` and VPC Connector name to set as deploy secrets
```
fangeugene, line 13: Can you also update https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/docs/Deployment/CI-CD-Setup.md

PR #3228, opened 2020-12-29
[py3] Track API calls after response

fangeugene:
Function to enqueue a callback to be run after the request response. Used for API tracking.

Usage examples:
`run_after_response(lambda: ...)`
`run_after_response(function_to_run)`
```
@run_after_response
def function_to_run():
    ...
```


src/backend/api/handlers/helpers/track_call.py
```diff
@@ -0,0 +1,16 @@
+from flask import request
+
+from backend.common.google_analytics import GoogleAnalytics
+from backend.common.models.api_auth_access import ApiAuthAccess
+from backend.common.run_after_response import run_after_response
+
+
+def track_call_after_response(api_action: str, api_label: str) -> None:
+    auth_key = request.headers.get("X-TBA-Auth-Key", request.args.get("X-TBA-Auth-Key"))
+    auth_owner_id = ApiAuthAccess.get_by_id(auth_key).owner.id()
```
fangeugene, line 10: It seems silly that we have to go through this logic again since we're already doing this in the `@api_authenticated` decorator. Is there a way to avoid this? The old class implementation let us store stuff in a request-instance variable.

src/backend/common/middleware.py
```diff
@@ -48,23 +49,21 @@ def __call__(self, environ: Any, start_response: Any):
         return self.app(environ, start_response)
 
 
-class AfterResponseMiddleware:
+class AfterResponseMiddleware(NdbMiddleware):
```
fangeugene, line 52: Is there a better way of doing this so we don't have to repeat the ndb client / redis object creation?

src/backend/api/handlers/helpers/track_call.py
```diff
@@ -0,0 +1,16 @@
+from flask import request
+
+from backend.common.google_analytics import GoogleAnalytics
+from backend.common.models.api_auth_access import ApiAuthAccess
+from backend.common.run_after_response import run_after_response
+
+
+def track_call_after_response(api_action: str, api_label: str) -> None:
+    auth_key = request.headers.get("X-TBA-Auth-Key", request.args.get("X-TBA-Auth-Key"))
```
fangeugene, line 9: It seems silly that we have to go through this logic again since we're already doing this in the @api_authenticated decorator. Is there a way to avoid this? The old class implementation let us store stuff in a request-instance variable.

src/backend/common/run_after_response.py
```diff
@@ -0,0 +1,32 @@
+import logging
+from typing import Callable
+
+from werkzeug.local import Local
+
+
+local = Local()
+local.callbacks = []
```
fangeugene, line 8: @phil-lopreiato how do i make pyre not be unhappy about this?

fangeugene: ill just pyre ignore it I guess

PR #3231, opened 2020-12-30
[py3] Implement more of APIv3

fangeugene:
Implement some more of APIv3. Not doing all of APIv3 to keep this PR small.


src/backend/common/queries/dict_converters/award_converter.py
```diff
@@ -9,6 +9,8 @@
 from backend.common.models.team import Team
 from backend.common.queries.dict_converters.converter_base import ConverterBase
 
+AwardDict = NewType("AwardDict", Dict)
```
phil-lopreiato, line 12: we can probably TODO these, but we should turn them into `TypedDict`s so we can properly schema-fy them

phil-lopreiato (approved):
seemslegit though

PR #3252, opened 2021-01-01
Remove Elasticsearch

ZachOrr:
Now that we're unable to access Elasticsearch, we should look to remove it from our codebase


controllers/datafeed_controller.py
```diff
@@ -484,11 +469,7 @@ def get(self, year):
         # For all new offseason events we can't automatically match, create suggestions
         SuggestionCreator.createDummyOffseasonSuggestions(new_offseason_events)
 
-        merged_events = EventManipulator.mergeModels(
-            list(events_to_put),
-            elasticsearch_events) if elasticsearch_events else list(
-                events_to_put)
-        events = EventManipulator.createOrUpdate(merged_events) or []
+        events = EventManipulator.createOrUpdate(events_to_put) or []
```
ZachOrr, line 472: This bit of code needs a review 👀

controllers/datafeed_controller.py
```diff
@@ -557,22 +538,15 @@ def get(self, event_key):
...
 
         # Update event
         fmsapi_events, fmsapi_districts = df.getEventDetails(event_key)
-        elasticsearch_events = df2.getEventDetails(event)
-        updated_event = EventManipulator.mergeModels(
-            fmsapi_events,
-            elasticsearch_events)
-        if updated_event:
-            event = EventManipulator.createOrUpdate(updated_event)
+        event = EventManipulator.createOrUpdate(fmsapi_events)
```
ZachOrr, line 549: This bit of code needs a review 👀

ZachOrr: [rant] The `delistify` method is such an anti-pattern. It changes our return types in unexpected ways. We have methods named `createOrUpdate` and `mergeModels` that take *lists* but return *single objects*.

If we're merging to lists with a single object, we should return a list with a single object. If we're merging two single objects, we should return a single object. The mental overhead of keeping track that a method called `mergeModels` that takes two lists will return a single object isn't worth not indexing in to an array to get an object.

PR #3256, opened 2021-01-03
Port DB Query Cache Clearing

phil-lopreiato:
This ports `get_affected_queries` from the old app and hooks it into the manipulators.

The other half of `CacheClearer` was about clearing apiv2 caches, which we're not porting, so I skipped it all

### DIFF
--- a/src/backend/common/cache_clearing/get_affected_queries.py
+++ b/src/backend/common/cache_clearing/get_affected_queries.py
@@ -6,7 +6,6 @@ from backend.common.models.cached_model import TAffectedReferences
 from backend.common.models.district_team import DistrictTeam
 from backend.common.models.event import Event
 from backend.common.models.event_team import EventTeam
-from backend.common.models.keys import TeamKey
 from backend.common.queries import (
     award_query,
     district_query,
@@ -29,10 +28,6 @@ def _queries_to_cache_keys_and_queries(
     for query in queries:
         out.append((query.cache_key, type(query)))
     return out
-
-
-def _get_team_page_num(team_key: TeamKey) -> int:
-    return int(team_key[3:]) // team_query.TeamListQuery.PAGE_SIZE
 
 
 def _filter(refs: Set[Any]) -> Set[Any]:
@@ -216,13 +211,13 @@ def team_updated(affected_refs: TAffectedReferences) -> List[TCacheKeyAndQuery]:
     queries: List[CachedDatabaseQuery] = []
     for team_key in team_keys:
         queries.append(team_query.TeamQuery(team_key.id()))
-        page_num = _get_team_page_num(team_key.id())
+        page_num = team_query.get_team_page_num(team_key.id())
         queries.append(team_query.TeamListQuery(page_num))
 
     for et_key in event_team_keys_future.get_result():
         year = int(et_key.id()[:4])
         event_key = et_key.id().split("_")[0]
-        page_num = _get_team_page_num(et_key.id().split("_")[1])
+        page_num = team_query.get_team_page_num(et_key.id().split("_")[1])
         queries.append(team_query.TeamListYearQuery(year, page_num))
         queries.append(team_query.EventTeamsQuery(event_key))
         queries.append(team_query.EventEventTeamsQuery(event_key))
@@ -243,7 +238,7 @@ def eventteam_updated(affected_refs: TAffectedReferences) -> List[TCacheKeyAndQu
     for team_key in team_keys:
         queries.append(event_query.TeamEventsQuery(team_key.id()))
         queries.append(team_query.TeamParticipationQuery(team_key.id()))
-        page_num = _get_team_page_num(team_key.id())
+        page_num = team_query.get_team_page_num(team_key.id())
         for year in years:
             queries.append(event_query.TeamYearEventsQuery(team_key.id(), year))
             queries.append(event_query.TeamYearEventTeamsQuery(team_key.id(), year))
--- a/src/backend/common/cache_clearing/get_affected_queries.py
+++ b/src/backend/common/cache_clearing/get_affected_queries.py
@@ -1,0 +1,313 @@
+from typing import Any, List, Set, Tuple, Type
+
+from google.cloud import ndb
+
+from backend.common.models.cached_model import TAffectedReferences
+from backend.common.models.district_team import DistrictTeam
+from backend.common.models.event import Event
+from backend.common.models.event_team import EventTeam
+from backend.common.models.keys import TeamKey
+from backend.common.queries import (
+    award_query,
+    district_query,
+    event_details_query,
+    event_query,
+    match_query,
+    media_query,
+    robot_query,
+    team_query,
+)
+from backend.common.queries.database_query import CachedDatabaseQuery
+
+TCacheKeyAndQuery = Tuple[str, Type[CachedDatabaseQuery]]
+
+
+def _queries_to_cache_keys_and_queries(
+    queries: List[CachedDatabaseQuery],
+) -> List[TCacheKeyAndQuery]:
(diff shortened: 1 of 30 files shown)


src/backend/common/cache_clearing/get_affected_queries.py
```diff
@@ -0,0 +1,313 @@
...
+    queries: List[CachedDatabaseQuery],
+) -> List[TCacheKeyAndQuery]:
+    out = []
+    for query in queries:
+        out.append((query.cache_key, type(query)))
+    return out
+
+
+def _get_team_page_num(team_key: TeamKey) -> int:
+    return int(team_key[3:]) // team_query.TeamListQuery.PAGE_SIZE
```
ZachOrr, line 35: This looks like it's just moving over code but - can we DRY this? It looks like this same bit of code is used in `team_query.py`

src/backend/common/cache_clearing/get_affected_queries.py
```diff
@@ -0,0 +1,313 @@
...
+    # Default filter() filters zeros, so we can't use it.
+    return {r for r in refs if r is not None}
+
+
+def award_updated(affected_refs: TAffectedReferences) -> List[TCacheKeyAndQuery]:
+    event_keys = _filter(affected_refs["event"])
+    team_keys = _filter(affected_refs["team_list"])
+    years = _filter(affected_refs["year"])
+    event_types = _filter(affected_refs["event_type_enum"])
+    award_types = _filter(affected_refs["award_type_enum"])
```
ZachOrr, line 48: Can we make `TAffectedReferences` a bit of a safer class here, where instead of referencing a dictionary we could have these be typed? So `affected_refs.event`, `affected_refs.year`, etc?

src/backend/common/deferred/__init__.py
```diff
@@ -42,7 +42,6 @@ def defer(
 def _client_for_env() -> TaskClient:
     # If we're running in tests, always return a FakeTestClient
     if os.environ.get("TBA_UNIT_TEST", None) == "true":
-        print("Returning here")
```
ZachOrr, line 45: Oops sorry

src/backend/conftest.py
```diff
@@ -53,5 +56,20 @@ def test_data_importer(ndb_client) -> JsonDataImporter:
     return JsonDataImporter(ndb_client)
 
 
+@pytest.fixture()
+def run_tasks_inline(monkeypatch: MonkeyPatch) -> fake_client.InlineTaskClient:
```
phil-lopreiato, line 60: @ZachOrr - a handy fixture to add that'll let us run queued work inline

src/backend/common/deferred/clients/fake_client.py
```diff
@@ -1,4 +1,46 @@
 from backend.common.deferred.clients.rq_client import RQTaskClient
+from backend.common.deferred.queues.rq_queue import RQTaskQueue
+from backend.common.deferred.tasks.task import Task
+from backend.common.redis import RedisClient
+
+
+class InlineRQTaskQueue(RQTaskQueue):
```
ZachOrr, line 7: nit: Can we move this to live with the rest of the queues?

src/backend/common/deferred/clients/fake_client.py
```diff
@@ -1,4 +1,46 @@
...
+
+
+class InlineRQTaskClient(RQTaskClient):
+    def queue(self, name) -> InlineRQTaskQueue:
+        return InlineRQTaskQueue(
+            name, default_service=self.default_service, redis_client=self._client
+        )
+
+
+class InlineTaskClient(InlineRQTaskClient):
```
ZachOrr, line 24: Can we just remove all of this `Inline` stuff and make this how the `FakeTaskClient` works?

I'm also not sure I understand why the double classes here (`InlineRQTaskClient` and `InlineTaskClient`)

src/backend/common/manipulators/manipulator_base.py
```diff
@@ -187,3 +200,40 @@ def _clearCache(cls, models: List[TModel]) -> None:
         Needs to save _affected_references and the dirty flag
         TODO implement this
```
ZachOrr, line 201: Can we remove this TODO comment?

src/backend/common/manipulators/manipulator_base.py
```diff
@@ -187,3 +200,40 @@ def _clearCache(cls, models: List[TModel]) -> None:
...
+                all_affected_references.append(model._affected_references)
+
+        if all_affected_references:
+            defer(
+                cls._clearCacheDeferred,
+                all_affected_references,
+                _queue="cache-clearing",
+                # this does not exist in Cloud Tasks
+                # _transactional=ndb.in_transaction(),
+                _target="default",
```
ZachOrr, line 215: Can we not have this work happen on `default`? Part of the thing we want to clean up is having `default` handle all of the catchall stuff.

src/backend/common/deferred/clients/fake_client.py
```diff
@@ -1,4 +1,46 @@
...
+        from fakeredis import FakeRedis
+
+        super().__init__(
+            default_service="test", redis_client=RedisClient.get() or FakeRedis()
+        )
+
+    def pending_job_count(self, queue_name: str) -> int:
+        return len(self.queue(queue_name).jobs())
+
+    def drain_pending_jobs(self, queue_name: str) -> None:
```
ZachOrr, line 41: Are these methods things we want to have on the `TaskClient` protocol? Or only here because they're only ever going to be used for tests? (The latter feels right to me - just making sure)

src/backend/common/cache_clearing/get_affected_queries.py
phil-lopreiato: Good catch :+1:

src/backend/common/cache_clearing/get_affected_queries.py
phil-lopreiato: This would be tricky, because the schema is dependent on the schema of the model. There may be some fancy way to do it, but I'd rather tackle that one separately.

src/backend/common/deferred/clients/fake_client.py
phil-lopreiato: sure, I can move them

src/backend/common/deferred/clients/fake_client.py
phil-lopreiato: I did that originally, but it broke all the `deferred` tests that constructed http requests. We can potentially change the default with some more work, but again, that would be better for a separate diff IMO

src/backend/common/manipulators/manipulator_base.py
phil-lopreiato: Yeah; what's a better value to use?

src/backend/common/deferred/clients/fake_client.py
phil-lopreiato: The latter - these should only be used in tests to run stuff inline

src/backend/common/deferred/clients/fake_client.py
ZachOrr: Maybe I can take a pass at it! I'm willing to update tests in order to make this better

src/backend/common/manipulators/manipulator_base.py
ZachOrr: Uhhh - `tasks-io` I believe

src/backend/common/deferred/clients/fake_client.py
phil-lopreiato: I think it makes sense as the default behavior, but it is probably work to be done separately from this diff

PR #3300, opened 2021-01-17
Install RegexConverter on all handlers

phil-lopreiato:
I'll need this in order to port the district detail url handlers, and it
just seems like a good idea for us to have available everywhere

## How Has This Been Tested?
defer tests cover this + write some explicit ones


ZachOrr:
Can we pull these tests and reorganize them so if we have to find the regex routing tests or where we install this via the middleware we know where to look?

phil-lopreiato:
I'm actually gonna split it out into a different function to add url converters, it'll be easier to understand that way. This is just me working through and splitting a bunch of little stuff out from my giant wip to support the district detail page

PR #3326, opened 2021-01-23
Rename EventHelper methods to be snake_case instead of camelCase, and to be less verbose. Add EventHelper.*_or_distant_future tests

ZachOrr:
The `EventHelper` methods have peev'd me for a while now, since they're named pretty poorly. I also added some tests, since it looked like these methods didn't have complete test coverage (but pretty good!)

### DIFF
--- a/src/backend/common/helpers/tests/event_helper_test.py
+++ b/src/backend/common/helpers/tests/event_helper_test.py
@@ -16,7 +16,7 @@ from backend.common.models.match import Match
 
 
 def test_calculate_avg_score_no_matches() -> None:
-    avg_score = EventHelper.calculateTeamAvgScoreFromMatches("frc254", [])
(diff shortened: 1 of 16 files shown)


src/backend/common/helpers/tests/event_helper_test.py
```diff
@@ -240,10 +278,14 @@ def test_sort_events(ndb_context) -> None:
     EventHelper.sort_events(events)
     assert events == [e2, e1]
 
+    events = [e2, e1]
```
ZachOrr, line 281: Note to future people who are like "why would we ever do this" - mostly just to make sure that this method is sorting events.  Passing the array in different states to ensure it's always the same does this. For a larger array you might want to pass the array in a random order during every test run but - this is just two elements, so this is easier

Issue #3368, opened 2021-02-05
Fatals From Bad Protobuf Encoding With Multi-Byte Characters


phil-lopreiato:
This is causing sporadic fatals, mostly around events that have participating teams with accents in their names.

Trace looks like
```

google.cloud.ndb._legacy_protocol_buffer.ProtocolBufferDecodeError: truncated

        at getPrefixedString (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/_legacy_protocol_buffer.py:204)
        at TryMerge (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/_legacy_entity_pb.py:280)
        at TryMerge (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/_legacy_entity_pb.py:456)
        at TryMerge (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/_legacy_entity_pb.py:699)
        at MergePartialFromString (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/_legacy_protocol_buffer.py:39)
        at __setstate__ (/workspace/backend/common/models/cached_model.py:86)
        at _from_base_type (/workspace/backend/common/models/cached_query_result.py:39)
        at call (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1824)
        at _call_from_base_type (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1636)
        at _opt_call_from_base_type (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1582)
        at _apply_to_values (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1860)
        at _get_user_value (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1525)
        at _get_value (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1890)
        at __get__ (/layers/google.python.pip/pip/lib/python3.8/site-packages/google/cloud/ndb/model.py:1932)
        at _do_query (/workspace/backend/common/queries/database_query.py:130)
[...]

PR #3369, opened 2021-02-05
Fix proto encoding bug with multibyte characters

phil-lopreiato:
This is due to the harder distinction between str/bytes in py3, which
meant the py2 copy pasta'd code had this bug.

To easily see how the bug comes up (but only when there are multi-byte
charactres in the string, most commonly with accented characters)

```python
# An example using the name from https://www.thebluealliance.com/team/3472
>>> s = b'PrepaTec CEM&Instituto Tecnol\xc3\xb3gico y de Estudios Superiores de Monterrey'
>>> print(len(s))
73
>>> print(len(s.decode()))
72
```

When we write the actual string into the output buffer, we will ensure
that we encode it to bytes, but we weren't doing that when we were
writing the control message for the string's length. That meant we had
an off-by-one, which is what caused later-on decodings to throw

```python
# Where we actually write the string, which didn't have the bug
# https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/src/backend/common/legacy_protobuf/legacy_gae_protobuf.py#L168-L175
    def putPrefixedString(self, v):
        assert isinstance(v, (str, bytes))
        if isinstance(v, str):
            v = v.encode()

        self.putVarInt32(len(v))
        self.buf.frombytes(v)
```

Fixes https://github.com/the-blue-alliance/the-blue-alliance/issues/3368


src/backend/common/legacy_protobuf/legacy_gae_entity_pb_encoder.py
```diff
@@ -369,7 +369,13 @@ def _byte_size_property_value(self: PropertyValue) -> int:
         if self.has_booleanvalue_:
             n += 2
         if self.has_stringvalue_:
-            n += 1 + EntityProtoByteSize.lengthString(len(self.stringvalue_))
+            assert isinstance(self.stringvalue_, (str, bytes))
+            if isinstance(self.stringvalue_, str):
+                str_len = len(cast(str, self.stringvalue_).encode())
```
phil-lopreiato, line 374: this bit is the key bug fix, since this is what we also do when we write the string

PR #3374, opened 2021-02-08
[py3] Support API auth from currently signed in web user

ZachOrr:
Porting code from https://github.com/the-blue-alliance/the-blue-alliance/blob/0f79f5ff1202bc1d9d45f3161334cd0ed211510e/api/apiv3/api_base_controller.py#L158

This does *not* add the `Origin`/`The Blue Alliance` owner authentication stuff. I assume that code is so we can hit the API from `web` but - I'm not actually sure we do that. So I'm not porting it.

- [ ] Add docs for `login_user` so no one else makes the same testing mistake I did 🤦


phil-lopreiato:
The origin thing is so the examples on the apidocs page works without a token, I believe

ZachOrr:
Good call! Thanks. Maybe I will add back in that auth mechanism in this PR so we can just write tests for it/be done

.gitignore
```diff
@@ -6,7 +6,6 @@ __pycache__/
 .pytest_cache/
 .pyre/
 .vagrant/
-.watchmanconfig
```
ZachOrr, line 9: I'll remove this - but I'm making a note to myself we need to add a note to create one of these files in the `Repo Setup` or something

src/backend/api/handlers/decorators.py
```diff
@@ -15,26 +16,43 @@ def decorated_function(*args, **kwargs):
         auth_key = request.headers.get(
```
ZachOrr, line 16: cc @fangeugene This probably the only interesting bit if you wanna take a look

src/backend/api/handlers/helpers/track_call.py
```diff
@@ -13,7 +13,7 @@ def track_call_after_response(
     Schedules a callback to Google Analytics to track an API call.
     """
     # Save |auth_owner_id| whiel we stil have access to the flask request context.
-    auth_owner_id = g.api_auth_access.owner.id() if g.api_auth_access.owner else None
+    auth_owner_id = g.auth_owner_id if g.auth_owner_id else None
```
ZachOrr, line 16: @fangeugene This bit as well - passing just the ID we need here as opposed to the full object in `g` so we can support other auth mechanisms but still get the logging

src/backend/api/handlers/decorators.py
fangeugene: `current_user()` doesn't make any DB calls if the session cookie is empty, right? Just want to make sure that we're not hitting the db excessively for normal API calls.

src/backend/api/handlers/helpers/track_call.py
fangeugene: Makes sense. Unrelated to this particular PR, but I'm wondering if we should wrap the `g` object with some sort of API so it's easier to know what we're setting. Right now it's not clear what we expect to be available in `g` without digging through a lot of code.

src/backend/api/handlers/decorators.py
ZachOrr: Right. If we don't have a cookie *or* if the cookie is invalid, we don't hit the database. We only hit the database if we have a valid cookie.

We could modify this code path as well to only attempt to look for a user if we don't have an auth token. I kinda started in one place and ended in another on this PR - so it might be worth a second pass that does that optimization.

ZachOrr: I filed a tracking issue for this, because this is a good thought, and I don't want to lose sight of it. I've got an idea of how we might go about designing this as well. For now - I wanna try to work on some more pressing issues we have in the codebase. https://github.com/the-blue-alliance/the-blue-alliance/issues/3386

src/backend/api/handlers/helpers/track_call.py
ZachOrr: I filed a tracking issue for this, because this is a good thought, and I don't want to lose sight of it. I've got an idea of how we might go about designing this as well. For now - I wanna try to work on some more pressing issues we have in the codebase. #3386

src/backend/api/handlers/decorators.py
ZachOrr: I updated this code to fallback to fetching a user object only if an auth key is not used. That way it's clearer that we won't be doing DB calls if they're not absolutely necessary without needing to understand how `current_user` works under the hood. Good call on this 👍

ZachOrr:
Updating - I'm punting on adding the `Origin` support until I can think about a more secure way to go about implementing this. I'm sure there's a better solution for this. I just don't want to get hung up on that bit right now.

I'm also going to do the doc updates in a different PR, since the docs will need to be updated with this change

PR #3432, opened 2021-03-17
Add FMS API event alliances + district list datafeed parsers

ZachOrr:
Adding the FMS API datafeed parsers to prep for adding the FMS API datafeed


src/backend/common/datafeeds/parsers/fms_api/fms_api_event_alliances_parser.py
```diff
@@ -0,0 +1,42 @@
+from typing import Any, Dict, List
+from backend.common.datafeeds.parsers.parser_base import ParserBase
+
+
+class FMSAPIEventAlliancesParser(ParserBase[List[Dict[str, Any]]]):
```
ZachOrr, line 5: We should look to see if we have a type alias for this sort of thing - or maybe expose one here? I'm sure we do this all over for alliances.

ZachOrr: Ah yeah sure enough - see [`EventAlliance`](https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/src/backend/common/models/alliance.py)

src/backend/common/datafeeds/parsers/parser_base.py
```diff
@@ -3,22 +3,22 @@
...
+class ParserBase(abc.ABC, Generic[TParsedResponse]):
     """
     Provides a basic structure for parsing pages.
     Parsers are not allowed to return Model objects, only dictionaries.
     """
 
     @classmethod
     @abc.abstractmethod
-    def parse(cls, html: bytes) -> TModel:
+    def parse(cls, html: bytes) -> TParsedResponse:
```
ZachOrr, line 21: Note to Zach in the AM - there are two different types of parsers - JSON parsers and HTML parsers. HTML parsers can take bytes, but JSON parsers will need to take a Dict. Right now the FMS API parsers don't have a common subclass, so we'll need to namespace and differentiate these parsers. Or maybe only add a JSON parser for now.

src/backend/common/datafeeds/parsers/fms_api/fms_api_event_alliances_parser.py
```diff
@@ -0,0 +1,59 @@
+from typing import Any, cast, Dict, List, Optional
+
+from backend.common.datafeeds.parsers.parser_json import ParserJSON
+from backend.common.models.alliance import EventAlliance, EventAllianceBackup
+
+
+class FMSAPIEventAlliancesParser(ParserJSON[Optional[List[EventAlliance]]]):
+    def parse(self, response: Dict[str, Any]) -> Optional[List[EventAlliance]]:
```
ZachOrr, line 8: Note to Zach: Either this should return no event alliances, or `FMSAPIDistrictListParser` should return an `Optional`. I'm leaning towards the latter, but you'll need to be careful on the integration component/what it means to have no data.

src/backend/common/helpers/event_team_status_helper.py
```diff
@@ -432,7 +432,11 @@ def _build_alliance_info(
             return None
 
         # Calculate the role played by the team on the alliance
-        backup_info = alliance.get("backup", {}) if alliance.get("backup") else {}
+        backup_info = (
+            cast(EventAllianceBackup, alliance["backup"])
+            if alliance.get("backup")
+            else {}
+        )
```
ZachOrr, line 439: cc @phil-lopreiato This file changed in two places due to the changes in https://github.com/the-blue-alliance/the-blue-alliance/pull/3432/files#diff-5c40aafca78f0698e6d073af075cb8120c7a4cdf3cea4615406def4fe52fab67 - could you confirm this looks right?

Also - why can't we do `alliance.backup`? Do the typed dicts not give us shorthand semantics? I tried adding a property to the `TypedDict` class without any luck.

src/backend/common/models/alliance.py
```diff
@@ -11,7 +11,7 @@
 class _EventAllianceOptional(TypedDict, total=False):
     declines: List[TeamKey]
     name: str
-    backup: EventAllianceBackup
+    backup: Optional[EventAllianceBackup]
```
ZachOrr, line 14: cc @phil-lopreiato This ended up being the big change here - we store `backup` as `null` in the database if it doesn't exist, which seems right? I'm unsure about this one.

phil-lopreiato: We were omitting it from the structure before if there was no backup (hence the `total=False` part), although maybe the parser was explicitly setting `None`? I suppose as long as we have all the cases null checks, which the type checker should enforce, this is an okay change

src/backend/common/helpers/event_team_status_helper.py
phil-lopreiato: I'm not sure why we need the cast here? Is this because of the default access thing?

Also, I don't think dot notation works for python dicts?

src/backend/common/models/alliance.py
ZachOrr: Oh wild. Okay I get this. Lemme take another pass and see if I can do this without making it optional.

ZachOrr: Yeah okay - I forget my context but I probably modeled it the same way the `event_team_status` https://github.com/the-blue-alliance/the-blue-alliance/blob/d9131246e3bddb96979b18610068b43e9e7d0aaa/src/backend/common/models/event_team_status.py#L65

phil-lopreiato: We could also store nulls - it would be fine (assuming the rest of the code passes the type check too)

ZachOrr: I'm good - we'll keep this as a non-Optional for now. I was trying to remember how I wound up where I did.

https://github.com/the-blue-alliance/the-blue-alliance/pull/3432/files#diff-b65753f8e7bb8f759942deb3731da998d5541339cd9d45675c07baac8cbe74f9R48

PR #3445, opened 2021-03-25
Add FMS API parsers for event awards and district rankings

ZachOrr:
This PR introduces a slight change where there are now two different types of JSON parsers - paginated and non-paginated. Also, the JSON parsers (paginated and non-paginated) take a non-optional generic type, but the `parse` method will return an optional. It's a bit of a safer way to define functionality (we want all of these parsers to return Optionals, for now)


src/backend/common/datafeeds/parsers/fms_api/fms_api_awards_parser.py
```diff
@@ -0,0 +1,76 @@
...
+        self.valid_team_nums = valid_team_nums
+
+    def parse(self, response: Dict[str, Any]) -> Optional[List[Award]]:
+        awards_by_type = {}
+        for award in response["Awards"]:
+            team_number = award["teamNumber"]
+
+            valid_team_nums = self.valid_team_nums
+            if valid_team_nums is not None and team_number not in valid_team_nums:
+                continue
```
ZachOrr, line 25: This is not currently tested and needs a test

src/backend/common/datafeeds/parsers/fms_api/fms_api_awards_parser.py
```diff
@@ -0,0 +1,76 @@
...
+
+            award_type_enum = AwardHelper.parse_award_type(award["name"])
+            if award_type_enum is None:
+                continue
```
ZachOrr, line 29: This is not currently tested and needs a test

src/backend/common/datafeeds/parsers/fms_api/fms_api_awards_parser.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class FMSAPIAwardsParser(ParserJSON[List[Award]]):
+    def __init__(self, event: Event, valid_team_nums: Optional[List[int]] = None):
+        self.event = event
+        self.valid_team_nums = valid_team_nums
+
+    def parse(self, response: Dict[str, Any]) -> Optional[List[Award]]:
+        awards_by_type = {}
+        for award in response["Awards"]:
```
ZachOrr, line 20: Thinking about this - these methods have 3 returns. Some values, an empty array (valid from the server), or incorrectly formatted data. The parsers right now are responsible for handling the zero data case (empty arrays) in an attempt to not overwrite data, but this feels like something we should be doing in the datafeeds or elsewhere.

We should update these models to remove the `return ... if ... else None` checks and only return `None` if the JSON was incorrect.

ZachOrr: Nah ya know what - I don't care about this right now. This is a complicated situation that will be more obvious what is the right answer once the datafeed gets brought over. I'm going to leave this as-is for now.

PR #3446, opened 2021-03-25
Move FMS API event rankings parser to py3

ZachOrr:


src/backend/common/helpers/rankings_helper.py
```diff
@@ -121,7 +122,6 @@ class RankingsHelper:
 
     QUAL_AVERAGE_YEARS = {2015}
 
-    """
     @classmethod
     def build_ranking(
```
ZachOrr, line 126: - [ ] `build_ranking` needs tests

PR #3488, opened 2021-04-17
Configure Dynamic NDB Cache TTLs

phil-lopreiato:
Until we can cutover writes to the new site fully, this should let us have live-updating content in py3.

It's dumb, but if the underlying events are live, then write to the NDB global cache with a short TTL, so if the data gets updated out from underneath us, it'll eventually expire here.

We can remove this once we cut all writes to be going through py3

### DIFF
--- a/src/backend/common/models/event.py
+++ b/src/backend/common/models/event.py
@@ -152,7 +152,6 @@ class Event(CachedModel):
 
     @classmethod
     def _global_cache_timeout(cls, key: datastore_key.Key) -> Optional[int]:
-        print(f"TTL: {key.__module__} {key} {key.id_or_name}")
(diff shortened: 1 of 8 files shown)


src/backend/common/models/event.py
```diff
@@ -149,6 +150,17 @@ def __init__(self, *args, **kw):
         self._week = None
         super(Event, self).__init__(*args, **kw)
 
+    @classmethod
+    def _global_cache_timeout(cls, key: datastore_key.Key) -> Optional[int]:
+        print(f"TTL: {key.__module__} {key} {key.id_or_name}")
```
ZachOrr, line 155: nit: Can we get rid of this print?

PR #3491, opened 2021-04-18
Add 2021/2020 Offseason support for rankings

ZachOrr:
- [x] Add `EventDetails.game_year` tests
- [x] Add `EventDetails.renderable_rankings` tests for 2021


old_py2/controllers/admin/admin_cron_controller.py
```diff
@@ -237,16 +237,13 @@ class AdminBackfillPlayoffTypeDo(LoggedInHandler):
...
-    EXCEPTIONS_2015 = ['2015cc', '2015cacc', '2015mttd']
-
     def get(self, year):
         self._require_admin()
         year = int(year)
         events = EventListQuery(year).fetch()
         for event in events:
             if not event.playoff_type:
-                if event.year == 2015 and event.key_name not in self.EXCEPTIONS_2015:
+                if event.game_year == 2015:
```
ZachOrr, line 246: This is no longer right - but also, I'm PRETTY SURE we don't need the `AdminBackfillPlayoffTypeDo` command anymore

phil-lopreiato:
We will at some point also need to update `RankingIndexes` (I'm not sure if this file has been ported over yet)

ZachOrr:
@phil-lopreiato The `RankingIndexes` look like they're only used for the `convert_rankings` call, which we mostly use to convert from `rankings` -> `rankings2`, which we're really no longer doing.

The other place we use `convert_rankings` is the `ApiTrustedEventRankingsUpdate` call, which is marked as a temp workaround https://github.com/the-blue-alliance/the-blue-alliance/blob/9700179c83c4a82cfca68b63e75f7c941b7a7ad3/old_py2/controllers/api/api_trusted_controller.py#L201

I was hoping to drop all of the `convert_rankings`-related code during the py3 migration just to clean things up.

phil-lopreiato:
Makes sense, and I do think the trusted API is the last place we actually rely on it

PR #3506, opened 2021-04-18
Bootstrap event year data in queue

ZachOrr:
Doing a bootstrap for an entire year can be VERY expensive, and usually the request times out before finishing. This PR moves the year task to a `bootstrap` queue to happen in the background.


ZachOrr:
Ex: A 2021 year bootstrap enqueues 189 events, which each hit several URLs. It's hardly possible to finish all of this work before the request times out.

Screen Shot 2021-04-18 at 2 31 36 PM 1

ZachOrr:
I actually might refactor this to use `run_after_response`, since it's probably easier than using a task queue

ZachOrr:
This needs to be updated for the new defer guidance 👀

PR #3513, opened 2021-04-20
Add support for FIRST Innovation Challenge Semi-Finalist, Rookie Game Changer, Skills Competition Winner/Finalist, Rookie Design, Engineering Design, Designer's, and Concept Award

ZachOrr:
We're failing to pull the FIRST Innovation Challenge Semi-Finalist Award right now because it's unsupported. This adds support for this new award.


helpers/award_helper.py
```diff
@@ -121,6 +121,7 @@
     (AwardType.MOST_IMPROVED_TEAM, (["most improved team"], [])),
     (AwardType.WILDCARD, (["wildcard"], [])),
     (AwardType.AUTONOMOUS, (["autonomous"], [])),
+    (AwardType.INNOVATION_CHALLENGE_SEMI_FINALIST, (["innovation", "challenge", "semi-finalist"], [])),
```
ZachOrr, line 124: Asking for a review because I've never done this before. Does this seem right? What's the best way to do this?

ofekashery (approved):

helpers/award_helper.py
fangeugene: Yep, this looks right. You can also just make it `["innovation challenge semi-finalist"]` to be extra safe, since we know the exact name of the award.

PR #3520, opened 2021-04-21
Move awards datafeed fetch to py3

ZachOrr:
Non-complete rolling todo list

- [x] Fix URL routing to match py2
- [ ] Point routing to py3 URLs
- [ ] Update `DatafeedFMSAPI._parse` to match new `_parse` method signature

## File tickets for...
- [ ] Add back in `DatafeedFMSAPI` `sim_time` and `save_response`
- [ ] Move `get_awards` `EventTeam.query` to database query (uncached)


src/backend/tasks_io/datafeeds/datafeed_fms_api.py
```diff
@@ -0,0 +1,320 @@
...
+    def _get_event_short(self, event_short: str, event: Optional[Event] = None) -> str:
+        # First, check if we've manually set the FRC API key
+        if event and event.first_code:
+            return event.first_code
+
+        # Otherwise, check hard-coded exceptions
+        return DatafeedFMSAPI.EVENT_SHORT_EXCEPTIONS.get(event_short, event_short)
+
+    def _parse(
+        self, response: requests.Response, parser: ParserBase
```
ZachOrr, line 114: - [ ] Don't merge this until we get the storage stuff put in. Doesn't need `sim_time` for now, does need writes.

PR #3529, opened 2021-04-24
Bypass global cache for insert/get sitevars

ZachOrr:
Part of https://github.com/the-blue-alliance/the-blue-alliance/pull/3528 - we don't have to clear the cache for the `api/status` endpoint if we don't cache sitevars.


ZachOrr:
Closing this - it looks like calling `put` will clear the global ndb cache properly, so we don't need to worry about evicting sitevars on updates https://github.com/googleapis/python-ndb/blob/master/google/cloud/ndb/_datastore_api.py#L385

PR #3530, opened 2021-04-25
Add EventHelper.remapteams_awards

ZachOrr:
Also adding FOC test to `EventHelper` since it didn't have any tests

### DIFF
--- a/src/backend/common/helpers/event_helper.py
+++ b/src/backend/common/helpers/event_helper.py
@@ -1,3 +1,4 @@
+import json
 from collections import OrderedDict
 from datetime import datetime, timedelta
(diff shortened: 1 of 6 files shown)


src/backend/common/helpers/event_helper.py
```diff
@@ -233,3 +236,29 @@ def week_events(cls) -> List[Event]:
                     events.append(event)
 
         return cls.sorted_events(events)
+
+    @classmethod
+    def remapteams_awards(
+        cls, awards: List[Award], remap_teams: Dict[str, str]
```
ZachOrr, line 242: Question asked in Slack - can we type this `str, str` situation so it's like, `old team, new team` or something?

PR #3541, opened 2021-04-26
[py3] Update from master

fangeugene:
Periodic merge with `master` to keep things in sync


fangeugene:
Wait I think I did this wrong :disappointed:

fangeugene:
Per Slack, we've decided to let the two branches diverge and manually open up PRs in both `master` and `py3` if the changes are worth pulling in.

PR #3574, opened 2021-05-01
Adds Firebase Auth emulator

ZachOrr:
Adds the Firebase Auth emulator for easier testing of accounts locally.

- [x] Add docs

### DIFF
--- a/ops/dev/vagrant/dev_appserver.sh
+++ b/ops/dev/vagrant/dev_appserver.sh
@@ -1,21 +1,9 @@
 #! /bin/bash
 set -e
 
-function get_config_prop {
-    local prop_name="$1"
-    local dev_config_file=""
-    if [ -f "tba_dev_config.local.json" ]; then
-        local dev_config_file="tba_dev_config.local.json"
-    fi
+source ops/dev/vagrant/config.sh
 
-    jq -c -r -s ".[0] * (.[1] // {}) | .$prop_name | select (.!=null)" tba_dev_config.json $dev_config_file
-}
-
-function get_project_from_key {
-    local key_path="$1"
-    jq -r '.project_id' "$key_path"
-}
-
+auth_use_prod=$(get_config_prop auth_use_prod)
 log_level=$(get_config_prop log_level)
 tba_log_level=$(get_config_prop tba_log_level)
 ndb_log_level=$(get_config_prop ndb_log_level)
@@ -112,6 +100,15 @@ else
     exit 1
 fi
 
+# Setup Firebase auth emulator
+if [ -z "$auth_use_prod" ]; then
+    echo "Running with Firebase auth emulator"
+    env+=("--env_var=FIREBASE_AUTH_EMULATOR_HOST=localhost:9099")
+else
+    echo "Using upstream authentication accounts"
+    assert_google_application_credentials
+fi
+
 set -x
 dev_appserver.py \
     --admin_host=0.0.0.0 \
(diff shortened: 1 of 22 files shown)


ZachOrr:
The last thing we need here is to write a script or add support for adding a default admin user in the emulator

https://firebase.google.com/docs/auth/admin/manage-users#create_a_user

ops/dev/vagrant/dev_appserver.sh
```diff
@@ -1,21 +1,9 @@
...
-    jq -c -r -s ".[0] * (.[1] // {}) | .$prop_name | select (.!=null)" tba_dev_config.json $dev_config_file
-}
-
-function get_project_from_key {
-    local key_path="$1"
-    jq -r '.project_id' "$key_path"
-}
+source ops/dev/vagrant/config.sh
 
+auth_use_emulator=$(get_config_prop auth_use_emulator)
```
ZachOrr, line 6: - [ ] Let's default this to true if it doesn't exist. Or better - maybe we have this be a `auth_use_prod` or something to make it less weird.

src/backend/web/handlers/account.py
```diff
@@ -123,7 +124,14 @@ def login() -> Response:
     else:
         if current_user():
             return redirect(url_for("account.overview"))
-        return make_response(render_template("account_login_required.html", next=next))
+        auth_use_emulator = Environment.auth_use_emulator()
+        return make_response(
+            render_template(
+                "account_login_required.html",
+                next=next,
+                auth_use_emulator=auth_use_emulator,
```
ZachOrr, line 132: - [ ] Update the `account` tests to support this

src/backend/web/templates/account_login_required.html
```diff
@@ -5,8 +5,13 @@
 {% block meta_description %}Login Required{% endblock %}
 
 {% block inline_javascript %}
+<script src="//gstatic.com/firebasejs/8.10.0/firebase-auth.js"></script>
 <script src="/py3_javascript/tba_combined_js.firebase.min.js"></script>
-<script src="//gstatic.com/firebasejs/8.2.1/firebase-auth.js"></script>
+<script>
+  if ({{ auth_use_emulator|tojson }}) {
+    auth.useEmulator("http://localhost:9099");
```
ZachOrr, line 12: - [ ] I'm pretty sure we can't do this in `tba_firebase.js` - but we should check.

src/web.yaml
```diff
@@ -4,7 +4,7 @@ entrypoint: gunicorn -b :$PORT backend.web.main:app
 
 instance_class: F1
 automatic_scaling:
-  max_idle_instances: 1
+  max_instances: 1
```
ZachOrr, line 7: - [ ] Remove

src/backend/web/templates/base.html
```diff
@@ -186,8 +186,8 @@ <h3 class="text-center">Follow your favorite teams with myTBA</h3>
 <!-- Admin Bar -->
 {% endblock %}
 
-<script src="//gstatic.com/firebasejs/8.2.1/firebase-app.js"></script>
-<script src="//gstatic.com/firebasejs/8.2.1/firebase-performance.js"></script>
+<script src="//gstatic.com/firebasejs/8.10.0/firebase-app.js"></script>
+<script src="//gstatic.com/firebasejs/8.10.0/firebase-performance.js"></script>
```
ZachOrr, line 190: - [ ] File a ticket to upgrade these to 9.x

tba_dev_config.json
```diff
@@ -1,4 +1,5 @@
 {
+    "auth_use_emulator": true,
```
ZachOrr, line 2: - [ ] As mentioned above - let's rename this something like `auth_use_prod` and have this default to true

PR #3583, opened 2021-05-02
Port Event Teams Trusted API

phil-lopreiato:
This is a pretty big pull, but it sets up the basics for running the trusted API on py3.

 - Set up an auth decorator that can check the various cases we expect + write tests
 - Bring in a CORS library so we can allow our custom headers to be passed
 - Split the API service into blueprints for apiv3/trusted, so we can easily shared prefixes among all the endpoints and also different CORS parameters
 - Move the `ParerInputException` into `common`, so the API-specific parsers can also use it. Also of note, a custom error handler turning the JSON errors into 400s. Plus, we use `pyre`'s `safe_json` API, which provides better type safety checks on the passed data
 - And on top of all that, implement the team list endpoint


src/backend/api/handlers/trusted.py
```diff
@@ -0,0 +1,50 @@
...
+    JSONTeamListParser,
+)
+from backend.common.manipulators.event_team_manipulator import EventTeamManipulator
+from backend.common.models.event import Event
+from backend.common.models.event_team import EventTeam
+from backend.common.models.keys import EventKey
+from backend.common.models.team import Team
+
+
+# @reqiure_auth(AuthType.EVENT_TEAMS)
```
phil-lopreiato, line 17: need to implement trusted API auth decorator

src/backend/common/datafeed_parsers/json/json_team_list_parser.py
```diff
@@ -0,0 +1,24 @@
+from typing import List, Union
+
+from pyre_extensions import safe_json
+
+from backend.common.models.keys import TeamKey
+from backend.common.models.team import Team
+
+
+# Parsers probably need to live in common, since they'll also be used on the API side
```
phil-lopreiato, line 9: @ZachOrr we should agree on where these live, since there's some base code for parsers in `tasks-io` currently, but we'll also need the JSON parsers to be exposed over the trusted API

ZachOrr: I'm good moving them to common! They were in common to begin with, I moved them to tasks-io to see if we could keep them out of common. If we need to move them back, let's do it

PR #3622, opened 2021-05-15
Re-style event page headers

tervay:
This is a slight refactor to event page headers that contain info about the event (eg dates, website).

## Description
* Introduces another grid row to event page header, which has 2 columns: one for a firstinspires link, and another for the FRC API attribution.
* The attribution col is has a `text-align: right` on large screens, but does not on small screens.
* Capitalized a word (`details` -> `Details`) that looked out of place when an event does not have a website

## Motivation and Context
Users are not immediately sure which FIRST link is which. Clicking on the wrong one is common in my experience. This more clearly separates attribution and the firstinspires link.

## How Has This Been Tested?
Just manually going to localhost event web pages with a regional, offseason, cmp division, and cmp parent event (`2019nytr`, `2019nyrr`, `2019dar`, `2019cmpmi` respectively).

C9PU2DA
3y3BkTY
xKUTfef
eyiennz
mSguRMj
R6PRXpZ
mmMU2JV

## Types of changes

### DIFF
--- a/src/backend/web/static/css/less_css/less/tba/tba_base.less
+++ b/src/backend/web/static/css/less_css/less/tba/tba_base.less
@@ -147,9 +147,3 @@ body {
 .tba-auth-text-apple {
   color: white;
 }
-
-@media (min-width: 768px) {
-  .pull-right-not-xs {
-    text-align: right;
-  }
-}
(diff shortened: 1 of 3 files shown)


fangeugene:
What are people's thoughts on moving this to the footer instead?

ZachOrr:
@tervay Can you rebase + add updated screenshots here?

PR #3632, opened 2021-05-17
Fix the FRC API attribution link

ofekashery:
Fix work done in #3625.
The ``footer-separator`` class sets the ``z-index`` to ``-1``, so the link isn't clickable.

Also, it fixes the design on mobile:
**Before:**
Footer Screenshot

**After:**
Footer Screenshot


fangeugene:
Does this solve this issue? If not, can you take a look at it too? thanks @ofekashery 

image

ofekashery:
@fangeugene Thanks! Fixed.

Screenshot

phil-lopreiato (approved):
Thanks for the fix!

PR #3690, opened 2021-06-17
Port CSV backups to py3

ZachOrr:


src/backend/common/decorators.py
```diff
@@ -49,3 +57,16 @@ def decorated_function(*args, **kwargs):
             return func(*args, **kwargs)
 
     return decorated_function
+
+
+def require_task(header: str):
+    def decorator(f: Callable):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            if header not in request.headers:
```
ZachOrr, line 66: Why do we not have `X-AppEngine-TaskName` hard-coded here?

ZachOrr:
Keeping this open to inform future work - but I do not plan on doing this work myself. IMO there's too many unknowns around "why bother with the CSVs" + "how do we properly export to BigQuery" (both things I don't really use) for me to guide this work.

ZachOrr:
Closing this because I don't want to work on this - but hoping someone else can use this code as a starting point at a later date.

PR #3780, opened 2021-08-09
Disable FIRST Sync Code input box for 2021 season.

MC42:
FIRST is not issuing sync codes for the 2021 season.  This disables the form input so teams don't input it.

## Motivation and Context
This mitigates teams submitting incorrect FIRST Sync codes for 2021 events.

## How Has This Been Tested?
This has only been tested via inspect element; the HTML is valid and renders correctly.  However we may encounter issues with form parsing, as I presently have no way to verify that in a dev environment.

image

## Types of changes

### DIFF
--- a/src/backend/web/templates/suggestions/suggest_offseason_event.html
+++ b/src/backend/web/templates/suggestions/suggest_offseason_event.html
@@ -135,7 +135,7 @@
                                 <tr>
                                     <td>FIRST Sync Code</td>
                                     <td>
-                                      <input class="form-control" type="text" name="first_code" placeholder="IRI"{% if first_code %} value="{{first_code}}"{% endif %}/>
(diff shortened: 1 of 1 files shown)


src/backend/web/templates/suggestions/suggest_offseason_event.html
```diff
@@ -135,7 +135,7 @@ <h3>Suggest An Event</h3>
                                 <tr>
                                     <td>FIRST Sync Code</td>
                                     <td>
-                                      <input class="form-control" type="text" name="first_code" placeholder="IRI"{% if first_code %} value="{{first_code}}"{% endif %}/>
+                                      <input class="form-control" type="text" disabled="disabled" name="first_code" placeholder="Like 'IRI'.  Disabled for 2021 season."{% if first_code %} value="{{first_code}}"{% endif %}/>
```
ZachOrr, line 138: Could we have a Jinja if/else to disable this for 2021? We should be able to get the `current_season` in to the Jinja templates (let me know if you need help with this).

It would be great to try to improve the placeholder text, or maybe add information on the page elsewhere about what the FIRST sync code is. I'm not sure this text makes it clearer what that code is.

PR #3997, opened 2021-11-20
Update to python 3.9

phil-lopreiato:
Base changelog: https://docs.python.org/3/whatsnew/3.9.html

This also changes the dev container to use `ubuntu:rolling`, which is the newest version, instead of the LTS. This means we have access to newer python versions sooner, so it's a benefit

### DIFF
--- a/.github/workflows/comment_screenshots_on_pr.yml
+++ b/.github/workflows/comment_screenshots_on_pr.yml
@@ -13,10 +13,10 @@ jobs:
     if: ${{ github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success' }}
     steps:
       - uses: actions/checkout@v2.4.0
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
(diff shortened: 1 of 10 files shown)


phil-lopreiato:
I'm going to wait on merging this, because we still should update the version of python shipped in the dev container to match.

I'm debating changing the container to be fedora, which will make it easier to use the latest & greatest python version from official repos, but we'll have to remove all traces of py2 first (which `gcloud` stil depends on, for now)

phil-lopreiato:
This requires https://github.com/the-blue-alliance/the-blue-alliance/commit/b3a15ff460e5bb8cbb93aaa6358368948795cbec to not hang.

I don't know what changed between 3.8 and 3.9 to cause this, but it seems to fix it (and also on my local machine with 3.10)

phil-lopreiato:
I think this was caused by https://bugs.python.org/issue39812 - or at least that's my guess, and I've already sunk too much time on this to keep going :stuck_out_tongue:

Issue #3995, opened 2021-11-20
[overall tracking] Figure Out Which GAE Builtins to Keep Using


phil-lopreiato:
So, seems like Google backtracked on this...

https://cloud.google.com/appengine/docs/standard/python/migrate-to-python3#services

> Beginning in September 2021, App Engine allows apps in the Python 3 runtime to access some App Engine bundled services and APIs previously available only in the Python 2 runtime. 

This is good news, because we can rip out a bunch of the janky compatibility shims we had to write. But unfortunate that we had to writ them in the first place.

A quick brainstorm from me and @ZachOrr on what we should do:
 - memcache -> legacy bundled
 - ndb -> new library, but do a janky import for write compatibility so I can delete my janky crap
 - testbed -> legacy bundled, so I can delete my datastore implementation
 - webserver -> stick with flask
 - app_identity -> had no good replacement, keep using for the stuff we need (verifying the project ID in production)
 - taskqueue -> I need to do some science, if we can get that stuff working locally without needing Redis, let’s drop my implementations
 - urlfetch -> let’s keep using requests
 - users -> We should stick with Firebase users except for things like admin auth validation - where we should continue to use Google (since it looks at the project IAM stuff)

PR #3999, opened 2021-11-21
Remove google.appengine.ext.deferred replacement for bundled py3 deferred

ZachOrr:
Drop our `google.appengine.ext.deferred` replacement and use Google's py3 bundled `google.appengine.ext.deferred` instead.

- [x] Update docs (since the guidance has changed)

Towards https://github.com/the-blue-alliance/the-blue-alliance/issues/3995


src/backend/common/manipulators/tests/base_manipulator_test.py
```diff
@@ -7,7 +7,6 @@
 
 from backend.common.cache_clearing.get_affected_queries import TCacheKeyAndQuery
 from backend.common.consts.api_version import ApiMajorVersion
-from backend.common.deferred.clients.fake_client import FakeTaskClient
```
ZachOrr, line 10: - [ ] Update tests

src/backend/common/manipulators/tests/district_manipulator_test.py
```diff
@@ -2,7 +2,6 @@
 
 import pytest
 
-from backend.common.deferred.clients.fake_client import FakeTaskClient
```
ZachOrr, line 5: - [ ] Update tests

src/backend/common/manipulators/tests/event_manipulator_test.py
```diff
@@ -6,7 +6,6 @@
 import pytest
 
 from backend.common.consts.event_type import EventType
-from backend.common.deferred.clients.fake_client import FakeTaskClient
```
ZachOrr, line 9: - [ ] Update tests

src/requirements.txt
```diff
@@ -7,14 +7,13 @@ Flask-WTF==1.0.0
...
 google-cloud-storage==1.43.0
-google-cloud-tasks==2.7.1
 grpcio==1.42.0
 Jinja2==3.0.3
 numpy==1.21.4
 oauth2client==4.1.3
 pyre-extensions==0.0.23
 requests==2.26.0
-rq==1.10.0
+hiredis==2.0.0
```
ZachOrr, line 16: - [ ] Pull this in to a different PR (we'll drop Redis eventually here)

src/tasks_io.yaml
```diff
@@ -8,5 +8,8 @@ automatic_scaling:
   max_idle_instances: 1
 
 handlers:
+  - url: /_ah/queue/deferred.*
```
ZachOrr, line 11: - [ ] Test `/_ah/queue/deferred.*` endpoints - I'm pretty sure they're unsupported. I think only `/_ah/queue/deferred` will work with the new bundled solution, since Flask doesn't have the idea of regex routing. We might be able to patch it though.

ZachOrr: I also don't know that we necessarily need this?

PR #4008, opened 2021-11-24
GH Actions fullstack test

fangeugene:
Very basic test to ensure Vagrant container can start and can serve pages.
Also posts screenshots of the homepage & GameDay to the PR for fun.


ops/test_fullstack.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
```
fangeugene, line 1: We can do something fancier in the future (e.g. use [Puppeteer](https://github.com/puppeteer/puppeteer) to test interactive stuff rather than just use requests which misses if JS crashes or doesn't load.

PR #4015, opened 2021-11-25
Re-add deferred catchall routing

ZachOrr:
The catch-all `/_ah/queue/deferred.*` routing was removed in #3999 - however, we still depend on it in production. This PR re-adds the catch-all routing using our old `install_deferred_routes` method.

### DIFF
--- a/docs/Developing/Queues-and-defer.md
+++ b/docs/Developing/Queues-and-defer.md
@@ -37,6 +37,21 @@ app = Flask(__name__)
 app.wsgi_app = wrap_wsgi_app(app.wsgi_app, use_deferred=True)
 ```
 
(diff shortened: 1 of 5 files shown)


src/backend/common/manipulators/manipulator_base.py
```diff
@@ -340,8 +340,7 @@ def _clearCache(cls, models: Iterable[TModel]) -> None:
                 cls._clearCacheDeferred,
                 all_affected_references,
                 _queue="cache-clearing",
-                # this does not exist in Cloud Tasks
-                # _transactional=ndb.in_transaction(),
+                _transactional=ndb.in_transaction(),
```
phil-lopreiato, line 343: I would be surprised if this works, because the the library probably attempts to reference transactions in the legacy datastore service, not the cloud one we're using

PR #4018, opened 2021-11-26
Migrate to Legacy Builtin NDB Library

phil-lopreiato:
Towards https://github.com/the-blue-alliance/the-blue-alliance/issues/3995

This will simplify a few things, mainly the fact we can just have native interop between old/new apps without having to do a bunch of hackery to get datastore seriliazation/memcache to agree on a data format.

The couple things still remaining:
 - [x] figure out if we can still use the remote api locally. Answer: no, because adding `googledatastore` creates dependency hell. We can file an upstream issue for this.
 - [x] update docs


.pyre_configuration
```diff
@@ -10,10 +10,7 @@
     {"site-package": "flask_cors"},
     {"site-package": "freezegun"},
     {"site-package": "google/appengine"},
-    {"site-package": "google/cloud/datastore_v1"},
```
ZachOrr, line 13: We should also be able to remove things like `fakeredis` and `redis` here too, right? It also looks like I forgot to remove `rq`

requirements.txt
```diff
@@ -6,15 +6,13 @@ flake8-tidy-imports
...
-InMemoryCloudDatastoreStub
 pre-commit
 pyre-check==0.9.8
 pytest
 pytest-cov
 pytest-custom_exit_code
 pytest-testmon
 PyYAML==6.0
-redis==3.5.3
 hiredis==2.0.0
```
ZachOrr, line 16: We can remove `hiredis` here too

.pyre_configuration
phil-lopreiato: We still need redis momentarily (for raw memcache + flask handler caching).

Those will be easy enough to move out in separate PRs though

requirements.txt
phil-lopreiato: not yet, but soon!

ZachOrr: Also `fakeredis`

PR #4025, opened 2021-11-26
Remove Remaining Redis References

phil-lopreiato:
Just some assorted tooling/scripts. We can merge this once we finally turn off the GCE instances, just in case we need to revert before then.


ops/shell/memorystore_shell.sh
```diff
@@ -1,91 +0,0 @@
-#! /bin/bash
```
ZachOrr, line 1: Can we drop this section from the maintainer runbook as well? Maybe with a quick tl;dr of what is cached, and when it'll be cleared?

ZachOrr:
Can we also remove `pyyaml`? I'm going to drop the `deploy_queues` script, which is the only other dependent of that library. One less library to install when bootstrapping

phil-lopreiato:
Yeah, I can drop that too

ops/shell/memorystore_shell.sh
ZachOrr: We can actually remove the `Memorystore (redis)` section of the `GAE Firebase Setup` page as well!

PR #4032, opened 2021-11-28
Dispatch fmsapi_awards endpoints to py3

ZachOrr:
Requires https://github.com/the-blue-alliance/the-blue-alliance/pull/3520 to land - but will allow us to fetch awards in py3

### DIFF
--- a/dispatch.yaml
+++ b/dispatch.yaml
@@ -34,6 +34,13 @@ dispatch:
 - url: "*/clientapi/*"
   service: clientapi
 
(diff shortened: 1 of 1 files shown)


dracco1993:
With #3520 merged, this should be good too now, right?

phil-lopreiato:
I don't want to actually run these in prod just _quite_ yet. But soon!

ZachOrr:
Closing since `tasks-io` is being routed to py3

PR #4087, opened 2021-12-28
[py3] Use `pyre check` for proper exit code

fangeugene:


phil-lopreiato:
And also fixes all the remaining type errors

PR #4109, opened 2021-12-31
Port Some Basic Admin Panel Pages

phil-lopreiato:
Includes:
 - auth keys
 - task launcher
 - event list
 - event detail

Still needs:
 - [x] remove legacy sitevar calls
 - [x] write tests


src/backend/web/handlers/admin/blueprint.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@admin_routes.before_request
+def require_gae_admin() -> None:
+    """
+    Ensure that only admins can access this blueprint
+    TODO: we should probably change this to use firebase auth?
+    """
+    print(f"user: {gae_login.get_current_user()}")
+    if not gae_login.is_current_user_admin():
```
phil-lopreiato, line 23: @ZachOrr are we okay continuing to depend on this, or would we rather use firebase users and roles or something

src/backend/web/handlers/admin/tests/event_test.py
```diff
@@ -0,0 +1,44 @@
...
+from google.appengine.ext import testbed
+from werkzeug.test import Client
+
+from backend.web.handlers.tests import helpers
+
+
+@pytest.fixture
+def login_gae_user(gae_testbed: testbed.Testbed) -> None:
+    gae_testbed.init_user_stub()
+    gae_testbed.setup_env(
```
phil-lopreiato, line 14: this doesn't seem to work, which is why the test is failing

src/backend/web/handlers/admin/blueprint.py
phil-lopreiato: ah, yeah, great catch

PR #4137, opened 2022-01-05
Port website blocklist to py3

ZachOrr:
The big difference here between the previous website blocklist is that we clear the team's website *before* doing a refetch. This allows us to handle the current situation where the FRC API is returning empty responses for websites


ZachOrr:
In the future, this should probably follow the `TeamManipulator` route so we can clear caches while we're at it.

fangeugene:
> In the future, this should probably follow the TeamManipulator route so we can clear caches while we're at it.

Why does this have to be a future problem? We should avoid DB writes that arent through manipulators.

PR #4254, opened 2022-02-20
Create py3 handler for eventwizard. Untested whether Eventwizard works.

gregmarra:
## Description
Attempts to fix Eventwizard, by getting the Eventwizard HTML to render again.

Patterned after how the Gameday Handler seems to be built.

I'm assuming I am supposed to move the template file out of old_py2 now that I am using it. Not sure what the convention here is.

## Motivation and Context
Needed for Israel Preseason

## How Has This Been Tested?
Loads on my machine

image

## Types of changes


phil-lopreiato:
Only thing - on the html template, we need to change `event wizard.js` to be referenced from `py3_javascript`, which is how we namespaced this to work with the routing rules

gregmarra:
@phil-lopreiato @fangeugene @ZachOrr can you ptal and lgtm this before I merge? I don't feel confident enough with the current code to merge, even with all checks passing.

gregmarra:
thanks, will take a look at that later today

gregmarra:
I don't expect to have time to come back to this in the next week.

gregmarra:
Peeked at this again, having general JS issues.
image

PR #4302, opened 2022-03-09
Disable sending push notifications to Android

ZachOrr:
Going to disable sending notifications to Android until we can confirm Android is getting the proper payloads to not crash


phil-lopreiato:
ref https://github.com/the-blue-alliance/the-blue-alliance/issues/4303

PR #4306, opened 2022-03-09
Attempts to fix match key routing issue, but untested due to local environment issues

gregmarra:
## Description
Fixes match key validation for events with novel keys by using Event key validation inside of Match key validation.

## Motivation and Context
New event key format breaks assumptions

## How Has This Been Tested?
Not tested, local issues. Hoping tests run in cloud :)

## Types of changes


gregmarra:
I to get my local python environment to a place I can run tests locally so I can actually iterate on this one to make it work. Testing in cloud verification is not the way to go. :)

gregmarra:
Will come back to this later.

PR #4362, opened 2022-03-27
Minor cleanup of bootstrapping a full season to avoid OOM.

MC42:
See title, the full season bootstrapper ended too soon to be super useful.

## Description
When running the dev setup to bootstrap an event, smaller requests (ie 2019mrcmp) would run fine, but bootstrapping an entire season would take too long to run and terminate too soon.  

## Motivation and Context
The workaround I had found was to use the task queue as shown in the OPR calculations to make each sub-event's parsing a distinct task and fit within the memory overhead.

In addition, I switched to using a session object with flask for such a thing to avoid having too many connections open, and this seemed to improve reliability of the bootstrap process.

## How Has This Been Tested?
Tested in the py3 branch earlier today, with both no key, valid key, valid alternates of production keys to test the behavior as needed.

image

New message if the user is importing a full season of data.

image

More useful error message if the user has a bad APIv3 key.

## Types of changes

### DIFF
--- a/src/backend/common/models/notifications/alliance_selection.py
+++ b/src/backend/common/models/notifications/alliance_selection.py
@@ -1,8 +1,15 @@
+from typing import Any, Dict, Optional
+
+from pyre_extensions import none_throws
+
+from backend.common.consts.notification_type import NotificationType
+from backend.common.models.event import Event
 from backend.common.models.notifications.notification import Notification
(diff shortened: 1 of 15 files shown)


ZachOrr:
Fixed in https://github.com/the-blue-alliance/the-blue-alliance/pull/3506

PR #4363, opened 2022-03-27
Fix building container with no secrets file

ZachOrr:
If we attempt to build the container with no secrets file, the JS fails to build, and people cannot use any of the Firebase JS. This PR adds a default secrets file to the container to make sure we can compile the JS at least once.


ZachOrr:
Posing here on a related thread - login doesn't work if there's no `google_application_credentials` file set. We might consider a more holistic solution where we create a test `google_application_credentials` and populate a dummy keys file from that.

Issue #4366, opened 2022-03-28
Quintet per alliance not per robot


arimb:
For 2022 insights, quintet rate is listed as (# of occurrences)/(6*(# of matches)), aka per robot-match. It should be (# of occurrences)/(2*(# of matches)), aka per alliance-match. 

For example, currently at ISCMP:
![0120FBFA-0A33-4206-9E7A-5AA16573D264](https://user-images.githubusercontent.com/13023392/160420027-10a549f3-7daa-475d-a0b5-2fa7abe04c79.jpeg)

PR #4372, opened 2022-03-28
Fix quintet opportunities

andrewda:
## Description

I noticed that the quintet success rates seemed suspiciously low at 2022orsal, but didn't question it too much. Looks like @arimb found the reason in #4366! This changes the quintet opportunities to use `opportunities_1x` (per-alliance) rather than `opportunities_3x` (per-robot).

## Motivation and Context

Closes https://github.com/the-blue-alliance/the-blue-alliance/issues/4366

## How Has This Been Tested?

It's been a while since my last contribution so I don't have the testing environment fully setup, but this change seems to be relatively straight-forward. We may want to re-run the event stat calculation tasks for all completed 2022 events to update them, but probably not too big of a deal if that's a tedious job. 

## Types of changes

### DIFF
--- a/src/backend/common/helpers/event_insights_helper.py
+++ b/src/backend/common/helpers/event_insights_helper.py
@@ -223,8 +223,8 @@ class EventInsightsHelper:
             / opportunities_1x,
             "quintet_count": [
                 quintet_count,
-                opportunities_3x,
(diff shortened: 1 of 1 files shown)


ZachOrr:
Thank you!

Issue #4448, opened 2022-04-25
2017cmpmo event page gives a 500 error


legoktm:
## Expected Behavior

I expect to be able to see the event results/info.

## Current Behavior

![Screenshot 2022-04-25 at 01-43-54 The Blue Alliance](https://user-images.githubusercontent.com/81392/165027524-1979c239-dfcc-4993-ae4f-e387238b8bfc.png)

Compare with https://web.archive.org/web/20210121214245/https://www.thebluealliance.com/event/2017cmpmo

## Steps to Reproduce (for bugs)

1. Visit https://www.thebluealliance.com/event/2017cmpmo

## Context

Trying to reply to a Chief Delphi post...

## Your Environment

* Version used: ?
* Environment name and version (e.g. Chrome 39, node.js 5.4): Firefox 99
* Operating System and version (desktop or mobile): Fedora 34/Qubes

phil-lopreiato:
Thanks for the report; sorry this took so long to fix

PR #4489, opened 2022-05-07
Fix 2017 round robin breakdowns + tests

phil-lopreiato:
The bug here is that we need to make sure to use score breakdown keys, which `score` was not

Also add some tests to cover the uncovered years

Fixes https://github.com/the-blue-alliance/the-blue-alliance/issues/4448


src/backend/common/helpers/playoff_advancement_helper.py
```diff
@@ -84,7 +84,7 @@ class BracketItem(TypedDict):
 class PlayoffAdvancementHelper(object):
 
     ROUND_ROBIN_TIEBREAK_BEAKDOWN_KEYS: Dict[Year, List[str]] = {
-        2017: ["score"],
+        2017: ["totalPoints"],
```
phil-lopreiato, line 87: this is the actual bugfix, the rest is tests

PR #4525, opened 2022-05-21
Port avatars page to py3

cjdenio:
## Description

Makes https://www.thebluealliance.com/avatars/2022 work.

## Motivation and Context

I clicked the "view avatars" button, only to discover it was stuck in 2020!

Screen Shot 2022-05-21 at 12 42 49 PM

## How Has This Been Tested?

Er, I couldn't figure out how to set up a py2 dev environment. So the answer to the question is "D) none of the above"

## Types of changes

### DIFF
--- a/src/backend/common/models/media.py
+++ b/src/backend/common/models/media.py
@@ -9,7 +9,7 @@ from backend.common.consts.media_tag import MediaTag
 from backend.common.consts.media_type import MediaType
 from backend.common.models.cached_model import CachedModel
 from backend.common.models.event import Event
-from backend.common.models.keys import MediaKey, TeamNumber, Year
+from backend.common.models.keys import MediaKey, Year
 from backend.common.models.team import Team
 
 
@@ -281,11 +281,3 @@ class Media(CachedModel):
     def avatar_image_source(self) -> str:
         image = json.loads(self.details_json)
         return "data:image/png;base64, {}".format(image["base64Image"])
-
-    @property
-    def team_number(self) -> TeamNumber:
-        """
-        Returns the associated team number for a piece of media.
-        """
-
-        return TeamNumber(self.references[0].id()[3:])
--- a/src/backend/common/models/media.py
+++ b/src/backend/common/models/media.py
@@ -9,7 +9,7 @@ from backend.common.consts.media_tag import MediaTag
 from backend.common.consts.media_type import MediaType
 from backend.common.models.cached_model import CachedModel
 from backend.common.models.event import Event
-from backend.common.models.keys import MediaKey, Year
+from backend.common.models.keys import MediaKey, TeamNumber, Year
 from backend.common.models.team import Team
 
 
@@ -281,3 +281,11 @@ class Media(CachedModel):
     def avatar_image_source(self) -> str:
         image = json.loads(self.details_json)
         return "data:image/png;base64, {}".format(image["base64Image"])
+
+    @property
+    def team_number(self) -> TeamNumber:
+        """
+        Returns the associated team number for a piece of media.
+        """
+
+        return TeamNumber(self.references[0].id()[3:])
--- a/src/backend/web/handlers/index.py
+++ b/src/backend/web/handlers/index.py
@@ -207,9 +207,7 @@ def avatar_list(year: Optional[Year] = None) -> Response:
     avatars_future = Media.query(
         Media.media_type_enum == MediaType.AVATAR, Media.year == year
     ).fetch_async()
-    avatars = sorted(
-        avatars_future.get_result(), key=lambda a: int(a.references[0].id()[3:])
-    )
+    avatars = sorted(avatars_future.get_result(), key=lambda a: a.team_number)
 
     template_values = {
         "year": year,
--- a/src/backend/web/handlers/index.py
+++ b/src/backend/web/handlers/index.py
@@ -215,7 +215,9 @@ def avatar_list(year: Optional[Year] = None) -> Response:
         avatars_future = Media.query(
             Media.media_type_enum == MediaType.AVATAR, Media.year == year
         ).fetch_async()
-        avatars = sorted(avatars_future.get_result(), key=lambda a: a.team_number)
+        avatars = sorted(
+            avatars_future.get_result(), key=lambda a: a.references[0].id()[3:]
+        )
 
         shards = {}
         size = len(avatars) / 10 + 1
--- a/src/backend/web/handlers/index.py
+++ b/src/backend/web/handlers/index.py
@@ -2,8 +2,8 @@ from datetime import datetime, timedelta
 from typing import Any, Callable, cast, Dict, Optional, Tuple
 
 from flask import abort, Response
-from google.appengine.ext import ndb
 from google.appengine.api import memcache
+from google.appengine.ext import ndb
 
 from backend.common.consts.landing_type import LandingType
 from backend.common.consts.media_type import MediaType
@@ -206,7 +206,7 @@ def avatar_list(year: Optional[Year] = None) -> Response:
         abort(404)
 
     avatars = []
-    shards = memcache.get_multi([f'{year}avatars_{i}' for i in range(10)])
+    shards = memcache.get_multi([f"{year}avatars_{i}" for i in range(10)])
     if len(shards) == 10:  # If missing a shard, must refetch all
         for _, shard in sorted(shards.items(), key=lambda kv: kv[0]):
             avatars += shard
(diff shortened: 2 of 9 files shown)


fangeugene:
Thanks for noticing the page doesn't work! This PR actually won't work, since py2 is deployed off `master` and not `py3/old_py2`. We have a few options here. One is to submit a PR to `master` to enable 2022 and work on porting it to py3 later. The other is to just port this directly to py3. The latter is probably better since we're trying to move everything to py3 by the end of summer anyways.

Thanks for your contribution and let me know if you need any help!

cjdenio:
Ah, thanks for explaining! I'll see what I can do to get the avatars page running on py3.

fangeugene:
I think the local data bootstrapping doesn't populate avatars. Let me try to get that working so you can test locally.

src/backend/web/handlers/team.py
```diff
@@ -117,5 +120,31 @@ def team_list(page: int) -> str:
...
         "current_page": page,
+        "current_season": SeasonHelper.get_current_season()
     }
     return render_template("team_list.html", template_values)
+
+
+@cached_public
+def avatar_list(year: Optional[Year] = None) -> Response:
+    if not year:
+        return redirect(url_for('avatar_list', year=SeasonHelper.get_current_season()))
```
ZachOrr, line 131: Do we need to redirect? Or could we just set year to be the current year?

We could also do this using default arguments in the function declaration (that way it's never None)

ZachOrr (changes requested):

ZachOrr:
This is looking great though! Thanks for this

src/backend/web/handlers/team.py
cjdenio: - Yeah, I agree it makes more sense to just render the current year.

- I actually tried that, but it started yelling `AssertionError: No api proxy found for service "memcache"` at me (!!!)

ZachOrr: Lemme checkout and see what's up with the second bullet point. I'm unsure why that would be an issue. Where were you seeing the assertion failure? In tests?

src/backend/web/handlers/team.py
```diff
@@ -119,3 +122,27 @@ def team_list(page: int) -> str:
...
+    year = year or SeasonHelper.get_current_season()
+
+    valid_years = list(range(2018, SeasonHelper.get_max_year() + 1))
+    valid_years.remove(2021)  # No avatars in 2021 :(
+
+    if year not in valid_years:
+        abort(404)
+
+    avatars_future = Media.query(Media.media_type_enum
+                                 == MediaType.AVATAR, Media.year == year).fetch_async()
```
ZachOrr, line 138: Could we build out a query here? (See the existing media queries for guidance https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/src/backend/common/queries/media_query.py) That way we leverage our caching layer and abstract away any of the query stuff

src/backend/web/handlers/team.py
```diff
@@ -119,3 +122,27 @@ def team_list(page: int) -> str:
...
+    avatars = sorted(avatars_future.get_result(),
+                     key=lambda a: int(a.references[0].id()[3:]))
+
+    template_values = {
+        "year": year,
+        "valid_years": valid_years,
```
ZachOrr, line 144: Do we use this `valid_years` in the template?

src/backend/web/handlers/team.py
```diff
@@ -119,3 +122,27 @@ def team_list(page: int) -> str:
...
+    valid_years = list(range(2018, SeasonHelper.get_max_year() + 1))
+    valid_years.remove(2021)  # No avatars in 2021 :(
+
+    if year not in valid_years:
+        abort(404)
+
+    avatars_future = Media.query(Media.media_type_enum
+                                 == MediaType.AVATAR, Media.year == year).fetch_async()
+    avatars = sorted(avatars_future.get_result(),
+                     key=lambda a: int(a.references[0].id()[3:]))
```
ZachOrr, line 140: This `avatar.references[0].id()[3:]` bit of code is used in both the template and here. Maybe we throw this as a helper on the `Media` class and DRY all of this code?

ZachOrr (changes requested):

ZachOrr:
Make sure to run the linter as well! https://github.com/the-blue-alliance/the-blue-alliance/wiki/Test-Lint-Check#lint

I'd recommend setting up the `virtualenv` as well first before running the lint - since running the lint in the container won't fix the files you commit, and you can avoid trashing your machine's dependencies https://github.com/the-blue-alliance/the-blue-alliance/wiki/Repo-Setup#virtualenv-install

src/backend/web/handlers/team.py
ZachOrr: Oh right yes - I understand this. `SeasonHelper.get_current_season` hits a `Sitevar` (backed by ndb). ndb for the instance isn't setup when the constructor is called.

All this is to say - I think the `year = year or ...` code  you have is great and the right way to go here :+1:

src/backend/web/handlers/team.py
ZachOrr: OH sure right. This is interesting actually. I wonder if we could bake this in to a paginated cached query (the way we still need to build out a paginated query https://github.com/the-blue-alliance/the-blue-alliance/pull/3389)

Either way - yeah Eugene is probably right here. We should build the query/caching stuff in to the method. But it does need the caching.

src/backend/web/handlers/team.py
cjdenio: Oops, that was for a then-unfinished year switcher thing. How do you feel about this?

https://user-images.githubusercontent.com/34525547/169712363-810984c2-6c9d-4e88-8e0a-8b806434759f.mov

(I could also stick this in a separate PR if it's out of scope)

src/backend/common/models/media.py
```diff
@@ -281,3 +281,11 @@ def image_direct_url_sm(self) -> str:
...
         image = json.loads(self.details_json)
         return "data:image/png;base64, {}".format(image["base64Image"])
+
+    @property
+    def team_number(self) -> TeamNumber:
+        """
+        Returns the associated team number for a piece of media.
+        """
+
+        return TeamNumber(self.references[0].id()[3:])
```
ZachOrr, line 291: Is `self.references` guaranteed to be non-empty? I don't know much about this model (cc @fangeugene)

src/backend/web/handlers/index.py
```diff
@@ -189,3 +192,27 @@ def index_insights(template_values: Dict[str, Any]) -> str:
...
+        abort(404)
+
+    avatars_future = Media.query(
+        Media.media_type_enum == MediaType.AVATAR, Media.year == year
+    ).fetch_async()
+    avatars = sorted(avatars_future.get_result(), key=lambda a: a.team_number)
+
+    template_values = {
+        "year": year,
+        "valid_years": valid_years,
```
ZachOrr, line 214: Mentioned before - but I don't think we need to pass `valid_years` to the template

src/backend/web/handlers/index.py
```diff
@@ -189,3 +192,27 @@ def index_insights(template_values: Dict[str, Any]) -> str:
...
+    valid_years = list(range(2018, SeasonHelper.get_max_year() + 1))
+    valid_years.remove(2021)  # No avatars in 2021 :(
+
+    if year not in valid_years:
+        abort(404)
+
+    avatars_future = Media.query(
+        Media.media_type_enum == MediaType.AVATAR, Media.year == year
+    ).fetch_async()
+    avatars = sorted(avatars_future.get_result(), key=lambda a: a.team_number)
```
ZachOrr, line 210: Talked about this earlier as well - but we need to do this query in batches so we don't OOM

cjdenio:
So I re-implemented the caching logic from py2 (copied and pasted in true programmer fashion), and it seems to be working.

Screen Shot 2022-05-24 at 11 08 56 PM

src/backend/common/models/media.py
fangeugene: `Media.references` is not guaranteed to contain references to a `Team` object. It can also be `Event`. But it is guaranteed to be not null. However, `references` being not null and a reference to a `Team` is loosely guaranteed if `media_type_enum` is an avatar.

@cjdenio I see how this is potentially cleaner, but doing `avatar.references[0].id()[3:]` in the template is safer for now. Maybe in the future we can build better typing around all of this.

cjdenio:
Alright, I think this is ready for some more review!

image

bdaroz:
Just checking here.... there's only 4 avatars? I could have sworn I saw a lot more in FMS screens this year.

cjdenio:
That's all development data– I just pulled in four random avatars for testing. Hopefully there are a few more in production.

fangeugene (approved):
Awesome, ship it! Thanks for getting this ported!

fangeugene:
@cjdenio would you mind creating another PR to delete all the ported code from `old_py2` just so we can track more easily what else needs to be ported? thanks!

PR #4529, opened 2022-05-22
[WIP] Move TBANS code to tasks_io (again). Add tasks_io endpoints for TBANS

ZachOrr:
This code shouldn't be in common really - so I'm moving it out, but doing it properly this time.


fangeugene:
image

src/backend/tasks_io/models/notification_requests/fcm_request.py
```diff
@@ -44,12 +52,12 @@ def __init__(self, app, notification, tokens=None, legacy_data_format=False):
...
         self.legacy_data_format = legacy_data_format
 
-    def __str__(self):
+    def __str__(self) -> str:
         return "FCMRequest(tokens={}, notification={})".format(
             self.tokens, str(self.notification)
         )
 
-    def send(self):
+    def send(self) -> messaging.BatchResponse:
```
ZachOrr, line 60: - [ ] The return type here between `FCMRequest` and `WebhookRequest` is different. I'm surprised our linter doesn't catch this. Think about how to standardize this.

ZachOrr:
This is reminding me of the cross-service request structure I built back in 2019 for this (and later removed) https://github.com/ZachOrr/the-blue-alliance/blob/ef991510a7f7e9682b24a64e2e508c85686d01fa/helpers/tbans_helper.py

iirc the removal was because of timeout issues across services? I need to think more about this. The cross-service situation is better than regular HTTP requests, since unlike the rest of the stuff that makes requests to `tasks-io`, these are making requests synchronously (usually) whereas other requests are leveraging deferred for cross-service execution.

ZachOrr:
Alright for sure closing this out

PR #4538, opened 2022-05-26
Remove avatar list from `old_py2`

cjdenio:
https://github.com/the-blue-alliance/the-blue-alliance/pull/4525#issuecomment-1138077291


fangeugene (approved):
ty!

PR #4658, opened 2022-08-02
Add privacy url and handler

dracco1993:
## Description
Add `/privacy` url and handler

## Motivation and Context
More `py3`

## How Has This Been Tested?
Locally

image

## Types of changes

### DIFF
--- a/ops/check_venv_deps.py
+++ b/ops/check_venv_deps.py
@@ -12,9 +12,7 @@ REQUIREMENTS_FILES = [
 def test_requirements():
     missing_requirements = []
     for requirement_file in REQUIREMENTS_FILES:
-        requirements = pkg_resources.parse_requirements(
-            requirement_file.read_text()
-        )
+        requirements = pkg_resources.parse_requirements(requirement_file.read_text())
         for requirement in requirements:
             requirement = str(requirement)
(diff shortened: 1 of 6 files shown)


ops/check_venv_deps.py
```diff
@@ -12,9 +12,7 @@
 def test_requirements():
     missing_requirements = []
     for requirement_file in REQUIREMENTS_FILES:
-        requirements = pkg_resources.parse_requirements(
-            requirement_file.read_text()
-        )
+        requirements = pkg_resources.parse_requirements(requirement_file.read_text())
```
dracco1993, line 15: Uh, the auto-linter got this one.  👀😂

dracco1993:
In addition to adding the stuff for the `/privacy` page, I bumped up the cache time to `1 week` for this and the past couple of pages, since they're static/redirects; if this isn't a good idea, or I should use something else, let me know.   👍

ops/check_venv_deps.py
phil-lopreiato: Whoops!

phil-lopreiato:
Yeah, a 1w cache time makes sense for this + the others

PR #4661, opened 2022-08-02
Add brand url and handler

dracco1993:
## Description
Add `/brand` url and handler.
In addition, to clean things up, I reduced the basically-static pages into a single `static` handler, as well as added basic tests to make sure each of those routes at least responds/redirects correctly.

## Motivation and Context
More `py3`

## How Has This Been Tested?
Locally.  Also, I added tests for the `static` handler

image

## Types of changes


phil-lopreiato (approved):
:100:

Issue #4716, opened 2022-09-05
Chezy Champs Double Elim Support


phil-lopreiato:
A few things we need to update from our old `DOUBLE_ELIM_8_TEAM` playoff format:

- match names should ideally be "round N" and not winner/loser bracket
- we need to build bracket support

![image](https://user-images.githubusercontent.com/2754863/189549469-eea5696f-2a75-4d34-9805-15eee4f6b5f8.png)

phil-lopreiato:
I think this now means we're good to go!

neyre:
Hi @phil-lopreiato @AndrewKe,

Thanks so much for getting this in on such short notice. Having the clear support on TBA for the bracket is so helpful from an event organizer perspective.

Right now, the bracket is listing "Unknown" instead of alliance names. Any way we can get this corrected? I realize it might be Cheesy Arena's fault, but if there's any way to just update the data on your end that'd be wonderful.

Thanks!

NIck

phil-lopreiato:
Yeah; ack on this (I saw, but just didn't get the chance to fix it last night). Will fix!

neyre:
Thanks, Phil!

Also, if y'all want the SVG files for the dynamically updated bracket, let me know, we could provide that if it's helpful for anything. (The Cheesy Arena bracket is just a SVG image that gets dynamically populated with the relevant data and then rendered in the browser).

PR #4766, opened 2022-09-28
Show Fallback Alliance Names In Double Elim Bracket

phil-lopreiato:
Fixes the bug where we wouldn't fall back to pre-computed alliance names, if they don't happen to be present in the data model.

This is was caused all alliances to be shown as "Unknown" for Chezy Champs

Towards https://github.com/the-blue-alliance/the-blue-alliance/issues/4716

image


src/backend/common/helpers/playoff_advancement_helper.py
```diff
@@ -654,34 +654,16 @@ def _alliance_name(
...
-                alliance_selection
-            ) in alliance_selections:  # search for alliance. could be more efficient
-                picks = alliance_selection["picks"]
-                if (
-                    len(set(picks).intersection(set(team_keys))) >= 2
-                ):  # if >= 2 teams are the same, then the alliance is the same
-                    return alliance_selection.get("name")
-
-        return ""
-
```
phil-lopreiato, line 684: why this was duplicated is a mystery to me

src/backend/common/helpers/playoff_advancement_helper.py
```diff
@@ -654,34 +654,16 @@ def _alliance_name(
...
-        ) in alliance_selections:  # search for alliance. could be more efficient
+        for (n, alliance_selection) in enumerate(
+            alliance_selections
+        ):  # search for alliance. could be more efficient
             picks = alliance_selection["picks"]
             if (
                 len(set(picks).intersection(set(team_keys))) >= 2
             ):  # if >= 2 teams are the same, then the alliance is the same
-                return alliance_selection.get("name")
+                return alliance_selection.get("name", f"Alliance {n+1}")
```
phil-lopreiato, line 664: Technically the "name" field in the alliance module is optional, so we can add a fallback for when it is not found (because in Cheesy Arean's case, it doesn't populate alliance names)

https://github.com/the-blue-alliance/the-blue-alliance/blob/48ed1c9560b9adb1b2b9efab7464bb76b8d9012e/src/backend/common/models/alliance.py#L9-L16

PR #4797, opened 2022-10-12
Add support for 9xxx teams

dracco1993:
## Description
Add support for 9xxx teams

## Motivation and Context
[Someone reported it](https://www.facebook.com/groups/moardata/posts/5713396142013967/), [`frc-events` has teams over 9k](https://frc-events.firstinspires.org/2023/allteams), and, you know, those teams deserve to be seen as well.  😂
I didn't _see_ anywhere else this would need to be changed, but can make whatever additional changes are needed, if any.

## How Has This Been Tested?
It worked locally.  👀 

## Types of changes

### DIFF
--- a/src/backend/web/handlers/team.py
+++ b/src/backend/web/handlers/team.py
@@ -17,7 +17,7 @@ from backend.web.profiled_render import render_template
 from backend.web.renderers.team_renderer import TeamRenderer
 
 MAX_TEAM_NUMBER_EXCLUSIVE = (
-    9000  # Support between Team 0 and Team MAX_TEAM_NUMBER_EXCLUSIVE - 1
+    10000  # Support between Team 0 and Team MAX_TEAM_NUMBER_EXCLUSIVE - 1
 )
 TEAMS_PER_PAGE = 1000
 VALID_PAGES = range(


fangeugene:
What's the max team number this year? I think first used to reserve 9900+ for off-season teams (do they still?). We should find a way to robustly support fake team numbers.

dracco1993:
> What's the max team number this year? I think first used to reserve 9900+ for off-season teams (do they still?). We should find a way to robustly support fake team numbers.

According to `frc-events`, [9176 is now a legitimate team](https://frc-events.firstinspires.org/2023/allteams).  👀

fangeugene:
Can we go to 9500 instead of 10000 so the fake teams stay hidden?

dracco1993:
Looking into it more, changing it to `9500` seems like it wouldn't prohibit those 9900+ teams from being shown, as the value is only used to calculate the valid pages, not the range of teams to show.  We could certainly add an actual limit to the results, but that'll take a bit more code.

fangeugene:
Ah you're right. Ok I think it's fine for now then

fangeugene (approved):

PR #4810, opened 2022-10-15
Port admin homepage to py3

phil-lopreiato:
As title.

All the stats stuff is stubbed out and the links not-ported will 404, but at least here's something


src/backend/web/handlers/admin/blueprint.py
```diff
@@ -66,6 +67,32 @@ def require_gae_admin() -> None:
         abort(401)
 
 
+@admin_routes.route("/")
+def admin_home() -> str:
```
fangeugene, line 71: Wow we don't need the trailing slash anymore!

PR #4824, opened 2022-10-21
Port team admin pages to py3

AndrewKe:
## Description
Port team admin pages (`/mod` and `/mod/redeem`) to py3. 

## Motivation and Context
More py3!

## How Has This Been Tested?
Manually tested + copied over a bunch of py2 tests. 

## Types of changes


src/backend/web/handlers/suggestions/suggest_team_media_review_controller.py
```diff
@@ -169,5 +169,5 @@ def post(self) -> Response:
         # Process rejects
         self._process_rejected(reject_keys)
 
-        return_url = request.args.get("return_url", "/suggest/team/media/review")
+        return_url = request.form.get("return_url", "/suggest/team/media/review")
```
AndrewKe, line 172: I think this was a typo

ZachOrr, line 172: 👍

ZachOrr (commented):
Can you also remove the ported code from py2 so I can do a diff between the moved code?

AndrewKe:
> Can you also remove the ported code from py2 so I can do a diff between the moved code?

done!

src/backend/web/handlers/team_admin.py
```diff
@@ -0,0 +1,252 @@
...
+    now = datetime.now()
+    existing_access = TeamAdminAccess.query(
+        TeamAdminAccess.account == user, TeamAdminAccess.expiration > now
+    ).fetch()
+
+    # If the current user is an admin, allow them to view this page for any
+    # team/year combination
+    forced_team = request.args.get("team")
+    forced_year = request.args.get("year")
+    if current_user().is_admin and forced_team and forced_year:
```
ZachOrr, line 50: I'm surprised you didn't have to wrap the user in a `none_throws`

ZachOrr: Oh yeah. Sorry. L39 - change this to DRY this up like -

```
user = none_throws(current_user())
user_account_key = user.account_key
...
if user.is_admin ...:
```

src/backend/web/handlers/team_admin.py
```diff
@@ -0,0 +1,252 @@
...
+    ]
+    team_medias_future = Media.query(
+        Media.references.IN(team_keys), Media.year.IN(years)
+    ).fetch_async(50)
+    suggestions_future = (
+        Suggestion.query(Suggestion.review_state == SuggestionState.REVIEW_PENDING)
+        .filter(Suggestion.target_model.IN(ALLOWED_SUGGESTION_TYPES))
+        .filter(Suggestion.target_key.IN([k.id() for k in team_keys]))
+        .fetch_async(limit=50)
+    )
```
ZachOrr, line 81: cc @phil-lopreiato Please advise here - should we be going through and using `Queries` here? My thought is like - maybe not? Team admin page stuff can probably bypass the caching? My first thought was like "We should make sure this is all using queries" but really I think it's like "We should make sure none of this is using queries"

The issue just being like - we might try to find a way to DRY this where we can use the `TeamSocialMediaQuery` but not leverage the caching component?

src/backend/web/handlers/team_admin.py
```diff
@@ -0,0 +1,252 @@
...
+
+    team_num_to_team = {
+        team.get_result().team_number: team.get_result() for team in teams_future
```
ZachOrr, line 84: Is there some way to DRY this so we don't have to write the `team.get_result()` twice?

src/backend/web/handlers/team_admin.py
```diff
@@ -0,0 +1,252 @@
...
+    }
+    team_num_to_robot_name = {
+        int(robot.get_result().team.id()[3:]): robot.get_result().robot_name
```
ZachOrr, line 87: Same question here around the double `robot.get_result()` situation

src/backend/web/handlers/team_admin.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+@blueprint.route("/mod", methods=["POST"])
+@require_login
+def team_mod_post():
+    team_number = request.form.get("team_number")
+    if not team_number:
+        return abort(400)
+    team_number = int(team_number)
+    team = Team.get_by_id("frc{}".format(team_number))
```
ZachOrr, line 137: nit: py3 format string here? We should be able to do it with the `int`

ZachOrr (changes requested):
This is awesome! Thanks for this work. Not a TON of things here - mostly nits. We just need to figure out that one database query and I think I'm good with this.

src/backend/web/handlers/team_admin.py
phil-lopreiato: I think this is okay to leave as it, at least for a direct port? I don't think any of the other suggestion pages use queries either, although I agree we could likely write some up and have them bypass cache/storing materialized results in `CachedQueryResult`.

But, for this diff, I think it's fine to just do this

ZachOrr: Okay!

ZachOrr (approved):

PR #4894, opened 2022-11-25
Add page to show regional events in a year

phil-lopreiato:
Similar to the district detail endpoint, show a page of regional events too.

This basically reserves the `regional` code to never be used as a district. Which ... why would it


fangeugene:
FIRST Regional District coming 2030

PR #5042, opened 2023-03-01
Hide Covid information for events outside 2022

jaredhasenklein:
## Description
Shows Covid-19 Site Information only if year = 2022.

## Motivation and Context
Currently, site information populates for years > 2021, which includes 2023 and will include subsequent years. However, there are no Covid-19 site info summaries for events outside of 2022. Current links error on the FIRST website.

## How Has This Been Tested?
Hasn't

## Types of changes


jaredhasenklein:
Actually, I should say that this link is now the "site information" doc which some events have and others don't. I don't think we need to include it, but if people want to, we could rename this "site information" and keep it in.

ZachOrr (approved):

gregmarra:
I think it would be fine to remove from all years now - the information
isn’t really relevant any longer and it will reduce code complexity.

Issue #3487, opened 2021-04-17
/suggest/review tabs do not link to pages


ZachOrr:
When clicking one of the tabs on the `/suggest/review` page, the tab will not link the user to the linked page. This bug was introduced in https://github.com/the-blue-alliance/the-blue-alliance/commit/15f3d26a7d52b6be315727f4ff603a9f3733a18e

TechplexEngineer:
@ZachOrr I believe #5079 fixes this.

PR #5079, opened 2023-03-09
Fix broken links on review pending suggestions page

TechplexEngineer:
## Description
On the https://www.thebluealliance.com/suggest/review page the links are nav-pills and this logic seems to not be needed, I'm not sure if this code is needed for any other page or if it should just be removed.

## Motivation and Context
The links are .nav-pills but do not use hash routing so the previous version of the code resulted in links that only work when opened in a new tab.

## How Has This Been Tested?
Built and ran locally.

N/A

## Types of changes


gregmarra:
Oh my goodness what a huge quality of life win. This has been so annoying
to me for years and I never investigated why.

ZachOrr:
Closes #3487

ZachOrr:
Thank you!

PR #5089, opened 2023-03-11
fixed double elimination bracket template

EmreAdabag:
Fixed a typo in the double elimination bracket which had the winner of match 12 advancing to match 13 AND the finals while the winner of match 11 didn't go anywhere.

## Description
typo fix


phil-lopreiato (approved):
thanks for the fix!

Issue #4955, opened 2023-01-18
Attempting to add webhook gives 404


adamsong:
## Expected Behavior

Clicking the Add Webhook button on the account page should initiate the webhook creation process.

## Current Behavior

The Add Webhook button redirects to a 404

## Possible Solution

Either make `/webhooks/add` work, or fix the link under the button

## Steps to Reproduce (for bugs)

1. Navigate to [account page](https://www.thebluealliance.com/account)
2. Click "Add Webhook"
3. Be taken to a 404

## Context

I cannot make a webhook

## Your Environment

* Version used: Unsure how to gather this, whatever was live as of Jan. 17, 2023 at 6:06 PM EST
* Environment name and version (e.g. Chrome 39, node.js 5.4): Firefox Version 108.0.2 (64-bit)
* Operating System and version (desktop or mobile): Windows 10 Desktop

BrenekH:
I'm also running into this issue when creating or deleting a webhook. And I don't know if this is related, but I'm not getting events sent to my existing webhook except for the ping event when I trigger it myself.

Theinatorinator:
Same issue!

ZachOrr:
The webhooks routes haven't been ported to py3. I'm unsure why we're not default routing these requests to the py2 instance though...

gaborszita:
Same issue - TBA developers - when is this expected to be fixed?

gregmarra:
This should be fixed by  https://github.com/the-blue-alliance/the-blue-alliance/pull/5094 , how are things looking now?

PR #5094, opened 2023-03-13
Migrate webhook_controller.py to py3

AndrewKe:
## Description

Migrate webhook_controller.py, so that webhooks can be created and deleted on py3. 

## Motivation and Context
https://github.com/the-blue-alliance/the-blue-alliance/issues/4955

## How Has This Been Tested?
Added pytests and also tested on local dev environment that I was able to create/delete/verify webhooks. 

However, I don't have a TBANS environment setup, so I wasn't able to e2e test with an actual webhook. 

## Types of changes


gaborszita:
When is this expected to be merged? Do you think it will be merged by Friday? We have a competition coming up this weekend and TBA webhooks would be essential to our new scouting system.

src/backend/web/handlers/webhooks.py
```diff
@@ -0,0 +1,140 @@
...
+from backend.web.profiled_render import render_template
+
+blueprint = Blueprint("webhooks", __name__, url_prefix="/webhooks")
+
+
+@blueprint.route("/add", methods=["GET"])
+@require_login
+def webhook_add():
+    template_values = {
+        "error": request.args.get("error"),
```
ZachOrr, line 22: We migrated to using the session for this sort of stuff as opposed to passing it through the URL args. See https://github.com/the-blue-alliance/the-blue-alliance/wiki/Web#passing-data-between-pages

src/backend/web/handlers/webhooks.py
```diff
@@ -0,0 +1,140 @@
...
+    to_delete = ndb.Key(Account, user.account_key.id(), MobileClient, int(client_id))
+    to_delete.delete()
+
+    return redirect(url_for("account.overview"))
+
+
+@blueprint.route("/verify/<int:client_id>", methods=["GET"])
+@require_login
+def webhook_verify(client_id):
+    template_values = {"error": request.args.get("error"), "client_id": client_id}
```
ZachOrr, line 87: Same note here around using the session as opposed to sending error values through args

src/backend/web/main.py
```diff
@@ -153,6 +156,7 @@
 app.register_blueprint(suggestion_blueprint)
 app.register_blueprint(suggestion_review_blueprint)
 app.register_blueprint(team_admin)
+app.register_blueprint(webhooks)
```
ZachOrr, line 159: I think there's tests for adding these blueprints? I forget

src/backend/web/handlers/tests/test_webhooks.py
```diff
@@ -0,0 +1,120 @@
+from unittest.mock import patch
+
+from werkzeug.test import Client
+
+from backend.common.consts.client_type import ClientType
+from backend.common.helpers.tbans_helper import TBANSHelper
+from backend.common.models.mobile_client import MobileClient
+from backend.common.models.user import User
+
```
ZachOrr, line 9: Can we add a few more tests to tests cases like attempting to send a verification for a webhook that doesn't belong to a user? Seems like there's some gaps around testing for error states as well

ZachOrr (changes requested):
Looking good as a first pass! A few things here to fix and this should be able to land.

Thanks for doing this!

src/backend/web/handlers/webhooks.py
AndrewKe: updated, thanks for the tip!

src/backend/web/main.py
AndrewKe: I couldn't find any specific tests, but I think test_webhooks should verify that that the routes are registered correctly.

ZachOrr (approved):

ZachOrr:
Thank you!

PR #5073, opened 2023-03-08
Support 2023 district points

andrewda:
## Description

This adds support for 2023 district points. There was discussion on Slack of getting district point information directly from FIRST, but as that seems like a relatively substantial change, this should at least temporarily fix the 2023 issue. I've already had a number of people confused why TBA's numbers don't match what they see on the FIRST website.

## Motivation and Context

Fixes #5054 

## How Has This Been Tested?

An additional test was added on 2023mijac, which was chosen as it contains an alliance with a backup team that was awarded district points.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would change API specifications or require data migrations)

PR #5095, opened 2023-03-13
Fix 2023 district point calculations for backup teams

andrewda:
## Description

@AGPapa noticed an issue in #5073 where backup team district points were calculated using the overall match count, rather than just the proportion of matches won by the alliance. This PR fixes that issue and adds a test on 2023njfla to cover that edge case.

## Types of changes


src/backend/common/helpers/tests/data/2023mijac_district_points.json
```diff
@@ -66,9 +66,9 @@
     "frc4381": {
       "alliance_points": 16,
       "award_points": 8,
-      "elim_points": 26,
+      "elim_points": 25,
       "qual_points": 20,
-      "total": 70
+      "total": 69
```
ZachOrr, line 71: I am so confused based on https://frc-events.firstinspires.org/2023/MIJAC/rankings#district and the https://frc-events.firstinspires.org/2023/district/FIM

This lists 4381 as having a point total for this event as 70... but obviously the totals they show add up to 69. And on the rankings their totals show as 69, but the event shows 70.

I'm gonna guess this 69 is right because like - it's just addition?

ZachOrr (approved):

src/backend/common/helpers/tests/data/2023mijac_district_points.json
andrewda: Ah thank you for finding that! I thought I had just miss-typed 70 in my previous PR and it happened to work out, but yeah FIRST's calculations don't seem to add up...

PR #5112, opened 2023-03-17
Awards elim district points as matches are won

AGPapa:
## Description
Instead of awarding points to the loser of a match, we can award points to the winner of the previous match. This lets us award district points earlier, while an event is ongoing.

## Motivation and Context
A website I help run, frclocks.com, uses the blue alliance for district points. That website tracks when a team clinches a spot to the district championship or to the world championship. We would like to mark when a team clinches a spot as soon as possible - while an event is ongoing. This change would allow frclocks to do that while still using the blue alliance api, rather than implementing our own points tracking system. It will also give TBA users accurate district points as an event is ongoing.

## How Has This Been Tested?
I've been having difficulties getting vagrant to run on my machine, so I have not been able to test this as well as I would like. I would appreciate it if others could verify this. I would also appreciate help in setting up vagrant 😅 .

Screen Shot 2023-03-17 at 10 17 26 AM

This is the calculation for how points are awarded by match. No matter which route through the bracket an alliance takes it should end up in the 30/20/13/7/0/0/0/0 split.

## Types of changes

### DIFF
--- a/src/backend/common/consts/district_point_values.py
+++ b/src/backend/common/consts/district_point_values.py
@@ -47,10 +47,8 @@ class DistrictPointValues:
     F_WIN = {2015: 5.0}
 
     # Double elimination bracket points (2023 onwards)
-    DE_SF_12_LOSS = 7
-    DE_SF_13_LOSS = 13
-    DE_F_LOSS = 20
-    DE_F_WIN = 30
+    DE_SF_WIN = {1: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 0, 7: 13, 8: 13, 9: 7, 10: 7, 11: 7, 12: 6, 13: 7}
+    DE_F_WIN = 10
 
     # Chairman's Award
     CHAIRMANS_DEFAULT = 10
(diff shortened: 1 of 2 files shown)


AGPapa:
@andrewda I'm getting some test failures when I run the python tests on my machine, but I believe they are in unrelated parts of the codebase. I would appreciate you double-checking that the code I have in this PR is correct and passes all of the checks it needs to. I am able to make any changes that are needed!

andrewda:
Nice, I like the approach you took here. I'll take a look at the tests in a bit.

AGPapa:
Looks like all of the tests passed in CI. I think any issues I had on my machine could be related to how I set up the development environment. I believe we're good to go.

Let me know if other changes are needed and I can make updates. We (frclocks) would like this to be deployed before the last day of the Israel District Championship on Thursday, as this will let frclocks work during playoffs for that event and let us test our code before the other district championships begin.

AGPapa:
According to [this Q&A](https://frc-qa.firstinspires.org/qa/248) points for backup teams at DCMPs should be rounded and then multiplied by 3. The existing implementation first multiplies by 3 and then rounds. I added a commit to this PR to fix that bug.

PR #5161, opened 2023-04-08
Slow down duplicate notification sends via memcache

ZachOrr:
Will add tests for this - but the idea is that right now our `MatchManipulator_runPostUpdateHook` and `EventDetailsManipulator_runPostUpdateHook` are getting 502s from GAE and I'm unsure why. The impact of this ends up being that we're sending way too many notifications to clients.

This PR will slow down duplicate notification sends by setting a memcache key and not allowing for duplicate sends while the key is active.


src/backend/common/helpers/tbans_helper.py
```diff
@@ -53,8 +54,34 @@ class TBANSHelper:
...
+
+        memcache = MemcacheClient.get()
+        memcache.set(key, True, time_seconds)
+
     @classmethod
     def alliance_selection(cls, event: Event, user_id: Optional[str] = None) -> None:
+        memcache_key = f"{event.key_name}_alliance_selection"
+        # Always send if we're passing an individual user_id
+        if user_id is None and TBANSHelper._has_sent_notification(memcache_key):
+            return
```
ZachOrr, line 83: This really shouldn't be a `return` it should be like... only send to webhooks

ZachOrr:
I need to think about this more. This might be getting done at the wrong level. The issue is we basically need to debounce push notifications but not webhooks. Once we hit the `_send` method we kind of don't have enough easy information to debounce this sort of thing. Especially if we always want individual sends kicked of manually to go to clients.

I might have to move some code around to make this work.

ZachOrr:
This is a good idea - but going to close this for now. Hopefully we've got the bugs ironed out here.

PR #5169, opened 2023-04-11
Handles district points for four-alliance events (MICMP)

AGPapa:
## Description
This PR handles elimination district points for the Michigan State Championship.

## Motivation and Context
The blue alliance is not handling this correctly currently. It is giving 30 points to the winners, when it should give 60 points to the winners and 30 to the finalists.

## How Has This Been Tested?
A test case has been added. Without this change the test case fails.

From the 2023 game manual:
Screen Shot 2023-04-11 at 6 29 47 PM

## Types of changes

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -77,7 +77,7 @@ jobs:
       - name: Run Tests
         run: ./ops/test_py3.sh
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@3.1.2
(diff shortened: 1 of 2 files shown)


AGPapa:
The codecov version string was recently updated to a bad value in #5167. I fixed that in this PR to get the CI working. This is the error I was getting before updating that:
```
Error: Unable to resolve action `codecov/codecov-action@3.1.2`, unable to find version `3.1.2`
```

PR #5173, opened 2023-04-12
Support docker compose for local dev

dracco1993:
## Description
Add support for `docker-compose` to get local development environments running

## Motivation and Context
The current `vagrant` setup is inconsistent in working and can be confusing to debug if things go wrong.

## How Has This Been Tested?
It works for me locally.
Others should 100% try this locally as well, and let me know if there's issues.

## Types of changes


.gitattributes
```diff
@@ -1 +1,4 @@
-old_py2/* linguist-documentation
+old_py2/* linguist-documentation
+
+# Set the default behavior, in case people don't have core.autocrlf set
+*.sh text eol=lf
```
dracco1993, line 4: This is needed to make sure `sh` files have `LF` endings, since by default Windows tries to "be helpful" and convert them to `CRLF`.  🙄

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: "3"
+
+services:
+  tba:
+    # image: ghcr.io/the-blue-alliance/the-blue-alliance/tba-py3-dev:latest
+    build:
+      context: ./
+      dockerfile: ./ops/dev/docker/Dockerfile
```
dracco1993, line 8: Using this `image` is the end goal, but for now this forces the local image build, for development/testing of this actual PR.

AGPapa (dismissed):
This is great! I was not able to get vagrant up and running, but I have a local version of TBA now using this branch. Installing docker and running `docker-compose up` was all I needed to do.

+1 - I love it!

dracco1993:
> This is great! I was not able to get vagrant up and running, but I have a local version of TBA now using this branch. Installing docker and running `docker-compose up` was all I needed to do.

That's awesome!  You don't know how happy it makes me to hear that!  lol

phil-lopreiato:
Very cool! This looks straightforward enough to me, but let me test this out locally before merging

docker-entry.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
```
phil-lopreiato, line 1: move this into `/ops/dev/docker`

phil-lopreiato (changes requested):
Could you add a note in `doc/setup/Setup-Guide.md` about how to use this, with some example commands, and maybe note that it's still in beta?

docker-entry.sh
dracco1993: Moved in acf63bf

phil-lopreiato (approved):
:ship:

PR #5175, opened 2023-04-13
[py3] Port landing page admin page

ZachOrr:
Ports the Landing Page config page in the admin panel to py3

Screenshot 2023-04-12 at 9 47 19 PM

Note that this changeset is a little bit larger than necessary for a 1:1 port because right now we're storing `"True"` and `"False"` in the landing page JSON and it drives me nuts because I never know how to safely edit those fields (in fact - our build season landing page expects `!= "False"` in some spots and a boolean in others). I've made these easier to work with by changing the inputs to be checkboxes to indicate `true` or `false` and added some custom code (Flask + Jinja2 doesn't have a great way to provide a boolean for a checkbox value - just present or not) to set booleans in the json from the checkboxes.

Build season landing page with all booleans on
Screenshot 2023-04-12 at 10 00 42 PM

Build season landing page with all booleans off
Screenshot 2023-04-12 at 10 01 06 PM


src/backend/common/sitevars/landing_config.py
ZachOrr, line 1: This is dumb - but I reorganized these to be in the same logical order as they get set in the landing page config, which is how they get set in the JSON. It doesn't matter really since none of this is technically ordered, but I did it anyways.

src/backend/web/templates/admin/main_landing.html
```diff
@@ -0,0 +1,39 @@
...
+    {% if value is boolean %}
+    <td><input type="checkbox" name="{{ key }}"{% if value %} checked{% endif %}/></td>
+    {% else %}
+    <td><input name="{{ key }}" value="{{ value }}" placeholder="Value" /></td>
+    {% endif %}
+  </tr>
+  {% endif %}
+{% endfor %}
+</table>
+
```
ZachOrr, line 35: I removed the "add key" situation since it's a bit of an antipattern now that our sitevars are typed in code. The template and code are still chill with adding these on the fly, but it's not like we add key/values for the landing page and don't ship a landing page code change. So this functionality didn't seem necessary to me.

src/backend/web/templates/index/index_buildseason.html
```diff
@@ -23,7 +23,7 @@ <h2>
       <div class="btn-group game-manual">
         <a class="btn btn-default" href="http://www.firstinspires.org/resource-library/frc/competition-manual-qa-system" target="_blank" rel="noopener noreferrer"><span class="glyphicon glyphicon-file"></span> {{game_name}} Game Manual and Materials</a>
       </div>
-      {% if build_handler_show_password != 'False' %}
```
ZachOrr, line 26: Updating these to use the proper typed booleans now

src/backend/web/templates/index/index_buildseason.html
```diff
@@ -34,18 +34,12 @@ <h2>
...
       </div>
 
-      {% if build_handler_show_ri3d != 'False' %}
+      {% if build_handler_show_ri3d %}
       <p>
         <a class="btn btn-lg btn-success btn-block" href="/watch/ri3d"><span class="glyphicon glyphicon-play"></span> Watch Ri3D Live</a>
       </p>
       {% endif %}
 
-      {% if build_handler_show_teameventsearch %}
```
ZachOrr, line 43: This looks like it's been removed from the codebase, so I'm going to remove it while I'm here (a little out of scope)

dracco1993 (approved):
Worked well for me locally, and the cleanup is nice too.  👍

rHUJXrL

PR #5176, opened 2023-04-13
Run lint on py3 branch

ZachOrr:
My lint job in https://github.com/the-blue-alliance/the-blue-alliance/pull/5175 was returning dirty changes, so I ran the lint script on the py3 branch to fix things up. I'm unsure where things changed... but the changes seem pretty benign


ZachOrr:
This is especially confusing because we have both `black` and `flake8` pinned 🤷‍♂️

Issue #5005, opened 2023-02-19
Implement 2023 season specifics


fangeugene:
See https://github.com/the-blue-alliance/the-blue-alliance/wiki/New-Season

- [x] Match breakdown #5004
- [x] Ranking sort orders #5004
- [x] Event match stats #5026
- [x] Tiebreakers #5007
- [x] Bonus RP dots #5028
- [x] Insights
- [x] Predictions
- [ ] Einstein advancement

ZachOrr:
Do we still need Einstein advancements to be done?

fangeugene:
> Do we still need Einstein advancements to be done?

I don't think so because it's just normal double elim not round robin. None of the tiebreaker rules change.

PR #5203, opened 2023-04-23
Fix inverted techFoulCount tiebreaker for 2023

lethosor:
## Description
This should fix the logic used to determine the winner of 2023 matches where the "tech foul" tiebreaker was used.

## Motivation and Context
https://www.thebluealliance.com/match/2023cmptx_sf12m1 currently has the wrong winner assigned (blue, should be red) because the tiebreaker logic is awarding the win to the alliance with the larger `techFoulCount` (the number of tech fouls committed by the alliance, which award points to the opposing alliance).

This was probably overlooked when copying tiebreaker logic from 2022, where `foulPoints` (points earned from fouls committed by the opposing alliance) was the first tiebreaker.

Followup to #5005/#5007

## How Has This Been Tested?
Unit test added and appears to pass in CI.

I also verified that all existing ties broken by `techFoulCount` are showing the wrong winner on TBA (this was only the 4th "normal" match of the season affected!):

* [2023txfor_sf6m1](https://www.thebluealliance.com/match/2023txfor_sf6m1): indicated as a red win by TBA. Audience display is not visible in the video, but the bracket has blue advancing despite indicating a red win.
<details>

image

</details>

* [2023mose_sf3m1](https://www.thebluealliance.com/match/2023mose_sf3m1): indicated as a blue win by TBA. Audience display and bracket show red winning/advancing.
<details>

image
image

</details>

* [2023ilpe_sf8m1](https://www.thebluealliance.com/match/2023ilpe_sf8m1): indicated as a red win by TBA. Audience display and bracket show blue winning/advancing.

<details>

image
image

</details>

* [2023week0_f1m2](https://www.thebluealliance.com/match/2023week0_f1m2) would also have been affected, but is a finals match where tiebreakers are not used.

image

## Types of changes

### DIFF
--- a/src/backend/common/helpers/match_tiebreakers.py
+++ b/src/backend/common/helpers/match_tiebreakers.py
@@ -84,7 +84,7 @@ class MatchTiebreakers(object):
         # Since tech foul points are not provided, we use the count instead.
         if "techFoulCount" in red_breakdown and "techFoulCount" in blue_breakdown:
             tiebreakers.append(
-                (red_breakdown["techFoulCount"], blue_breakdown["techFoulCount"])
+                (blue_breakdown["techFoulCount"], red_breakdown["techFoulCount"])
             )
         else:
             tiebreakers.append(None)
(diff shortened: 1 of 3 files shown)


phil-lopreiato (approved):
Great catch, and thanks for the fix!

PR #5204, opened 2023-04-23
Removes (Houston) from 2023 Champs to match 2022 Champs

AGPapa:
## Description
There is special logic to make the two champs names correct that applied for 2023 when I believe it should not have. This makes 2023 look the same way as 2022.

## Motivation and Context
Champs for 2023 is called `Einstein Field (Houston)` when last year it was called just `Einstein Field`.

## How Has This Been Tested?
I have only run the automated tests.

## Types of changes


phil-lopreiato (approved):
Thanks for the fix!

PR #5219, opened 2023-04-27
Updates short event names for 2023 edge cases

AGPapa:
## Description
This updates how short names are created for unusual event names that came across in 2023.

It does make some changes to past years. For example, "FITDC" (FIRST in Texas District Championship) becomes "TDC".

Let me know if you have any feedback or suggestions on these changes.

After this is merged I believe that some events will need to be re-ingested to capture these changes.

## Motivation and Context
Here's an example of the short event names for frc33.
Screen Shot 2023-04-26 at 8 39 57 PM

They look ugly and are not short! This change switches "Kettering University #1 presented by Ford" to "Kettering University #1". It also changes "FIMSCPBDF - APTIV" to "MSC - APTIV" (which matches previous years). It also changes "in Michigan" to just "Michigan". 

## How Has This Been Tested?
The automated unit tests were run.

## Types of changes


src/backend/common/helpers/event_short_name_helper.py
```diff
@@ -72,8 +101,7 @@ def get_short_name(cls, name_str: str, district_code: Optional[str] = None) -> s
                 result = match.group(1).strip()
             else:
                 result = short.strip()
-            if result.startswith("FIRST"):
-                result = result[5:]
-            return result.strip()
+            result = re.sub("FIRST( in)?", "", result, flags=re.IGNORECASE)
+            return " ".join(result.split())
```
AGPapa, line 105: I’m going to double check this and make sure we’re not going to accidentally remove “first” if it’s part of a word.

AGPapa: Updated 👍

fangeugene (approved):
This is super cool! Regex matching to determine short names is kind of a mess, so thanks for making this better!

PR #5235, opened 2023-05-02
Port Basic Mobile API Registration Methods

phil-lopreiato:
Okay I think this is essentially what we'll need.

If I grab a id_token for my user, and put it over curl, then I can successfully register a client

```
[phil@fedora-x1think the-blue-alliance-py3]$ curl -H "Authorization: Bearer <snip>" -H "Content-Type: application/json" --data '{"name": "test device", "operating_system": "android", "mobile_id": "abc123", "device_uuid": "test"}' http://localhost:8080/clientapi/tbaClient/v9/register

{"code":200,"message":"Registration successful"}
```

image

This is how we get tokens in the android app, so I think this should be good:

https://github.com/the-blue-alliance/the-blue-alliance-android/blob/4313501d57594089ff01de2705bdc544a3d19eb9/android/src/main/java/com/thebluealliance/androidclient/auth/firebase/FirebaseAuthProvider.java#L64-L76

This PR also adds a fullstack test to verify that the endpoints work end-to-end. Highlights:
 - the test container will run using the auth emulator
 - [This API](https://github.com/firebase/firebase-admin-python/blob/f0865f7493a2c642d2e551efdf19da1e53c1a8c3/firebase_admin/__init__.py#L258-L261) can throw a `SessionNotFoundException`, so we need to specifically plumb through the project ID using thje environment variables we set at startup (the value will be [`demo-test`](https://github.com/the-blue-alliance/the-blue-alliance/blob/504b2de197ebb9cee3a570808e3ff535f0cf9578/ops/dev/vagrant/dev_appserver.sh#L27-L29))
 - The test use the firebase auth [REST API](https://firebase.google.com/docs/reference/rest/auth) to create a user on the emulator, and log in with its credentials to get an id token, which we can then pass in the `Authorization` header
 - Also pull the GAE `dev_appserver` logs off the container and print them into the github actions logfile, for easier debugging

### DIFF
--- a/src/dispatch.yaml
+++ b/src/dispatch.yaml
@@ -1,6 +1,9 @@
 dispatch:
   - url: "*/api/*"
     service: py3-api
+  
+  - url: "*/clientapi/*"
+    service: py3-api
 
   - url: "*/tasks/*"
     service: py3-tasks-io
--- a/ops/dev/pull-gae-logs.sh
+++ b/ops/dev/pull-gae-logs.sh
@@ -13,4 +13,4 @@ host=$(echo -e "${config}" | grep HostName | cut -d " " -f 4)
 port=$(echo -e "${config}" | grep Port | cut -d " " -f 4)
 keyfile=$(echo -e "${config}" | grep IdentityFile | cut -d " " -f 4)
 
-scp -v -P "$port" -i "$keyfile" -oStrictHostKeyChecking=no "root@$host:/var/log/tba.log" "/tmp/tba.log"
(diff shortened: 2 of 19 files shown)


src/dispatch.yaml
```diff
@@ -1,6 +1,9 @@
 dispatch:
   - url: "*/api/*"
     service: py3-api
+  
+  - url: "*/clientapi/*"
+    service: py3-api
```
fangeugene, line 6: reminder that this dispatch file is not used, the one in `master` is.

Though maybe we should copy that one here and deploy from this one.

phil-lopreiato: This is what I want for testing, where we don't have to do the default routes, but I can still run a local app build to change the service url directly to hit it on py3 and test it

PR #5261, opened 2023-05-09
Update home page links

leongmichael:
## Description

Change 2021 game and season resources to 2023 game and season resources and remove the 2021 season details button

## Motivation and Context
The years are wrong and the updated version is now consistent with the other index page templates

Current home page
image

Updated home page
image

## Types of changes


fangeugene (approved):
Ty! Though I think we switched to the wrong homepage anyways (we should be using the insights landing page) so I'll make that change as well.

PR #5289, opened 2023-05-26
District teams list: only show teams who've competed that year

### LINKED ISSUE #5106: NE team list includes inactive teams
The current list of teams in the New England district has about 220 teams, but there are only 186 registered teams. We lost a large number of teams in 2021. The links for missing teams go to a 404 page in the web site.

I don't know if other districts have the same issue.

cjdenio:
## Description

The [district teams list](https://www.thebluealliance.com/events/ne#teams) sometimes includes teams that didn't actually compete in a given year. This seems to be an issue with the FIRST API:
Screenshot 2023-05-26 at 1 59 29 PM

This change filters the list by only showing teams that have competed (or will compete) in an event that year.

## Motivation and Context

Fixes #5106!

## How Has This Been Tested?

I've verified that after this change, the NE team list matches the one on FIRST's website.

## Types of changes

### DIFF
--- a/src/backend/common/queries/team_query.py
+++ b/src/backend/common/queries/team_query.py
@@ -14,6 +14,7 @@ from backend.common.queries.dict_converters.team_converter import (
     TeamDict,
 )
 from backend.common.tasklets import typed_tasklet
+import re
 
 
 def get_team_page_num(team_key: str) -> int:
@@ -94,10 +95,13 @@ class DistrictTeamsQuery(CachedDatabaseQuery[List[Team], List[TeamDict]]):
     def _query_async(
         self, district_key: DistrictKey
     ) -> Generator[Any, Any, List[Team]]:
+        year = int(re.match(r"\d{4,}", district_key).group(0))
         district_teams = yield DistrictTeam.query(
             DistrictTeam.district_key == ndb.Key(District, district_key)
         ).fetch_async()
         team_keys = map(lambda district_team: district_team.team, district_teams)
+        event_teams = EventTeam.query(EventTeam.year == year).fetch()
+        team_keys = filter(lambda team_key: any(team_key == event_team.team for event_team in event_teams), team_keys)
         teams = yield ndb.get_multi_async(team_keys)
         return list(teams)
 


src/backend/common/queries/team_query.py
```diff
@@ -94,10 +95,13 @@ def __init__(self, district_key: DistrictKey) -> None:
     def _query_async(
         self, district_key: DistrictKey
     ) -> Generator[Any, Any, List[Team]]:
+        year = int(re.match(r"\d{4,}", district_key).group(0))
```
cjdenio, line 98: I'm assuming there's some sort of helper for parsing these but I couldn't find one.

src/backend/common/queries/team_query.py
```diff
@@ -94,10 +95,13 @@ def __init__(self, district_key: DistrictKey) -> None:
...
         district_teams = yield DistrictTeam.query(
             DistrictTeam.district_key == ndb.Key(District, district_key)
         ).fetch_async()
         team_keys = map(lambda district_team: district_team.team, district_teams)
+        event_teams = EventTeam.query(EventTeam.year == year).fetch()
```
cjdenio, line 103: I know this is a somewhat expensive query, though I'm hoping this is fine since queries are cached.

src/backend/common/queries/team_query.py
```diff
@@ -98,6 +99,18 @@ def _query_async(
             DistrictTeam.district_key == ndb.Key(District, district_key)
         ).fetch_async()
         team_keys = map(lambda district_team: district_team.team, district_teams)
+
+        matches = re.match(r"\d{4,}", district_key)
+        if matches is not None:
+            year = int(matches.group(0))
+            event_teams = EventTeam.query(EventTeam.year == year).fetch()
```
fangeugene, line 106: This query is pretty expensive. Using `.fetch(keys_only=True)` should be sufficient (although you'll need to parse out the team key from the EventTeam key).

src/backend/common/queries/team_query.py
```diff
@@ -98,6 +99,18 @@ def _query_async(
             DistrictTeam.district_key == ndb.Key(District, district_key)
         ).fetch_async()
         team_keys = map(lambda district_team: district_team.team, district_teams)
+
```
fangeugene, line 102: I can't comment on line 87, but `CACHE_VERSION` needs to be bumped.

fangeugene (commented):
Overall this seems good. Though I wonder if filtering on query is the correct thing to do, or if we should be doing more validation on write?

cc @phil-lopreiato thoughts?

src/backend/common/queries/team_query.py
fangeugene: Also, the EventTeam query can probably be done asynchronously with `DistrictTeam.query` by using `fetch_async()`

e.g.
`district_teams, event_teams = yield DistrictTeam.query(...), EventTeam.query().fetch_async(keys_only=True)`

src/backend/common/queries/team_query.py
```diff
@@ -98,6 +99,18 @@ def _query_async(
...
+        matches = re.match(r"\d{4,}", district_key)
+        if matches is not None:
```
fangeugene, line 104: I don't think we ever expect this to not match, so it might be cleaner just to throw if this parse fails instead of handling the `None` case.

fangeugene: I'm trying to see if we validate the district key anywhere (and throw a 404) or if we expect to return an empty list for invalid district keys.

fangeugene: Oh jk this is the query not the API 🤦 

yeah then this should be robust to incorrect keys and not throw an exception. ignore everything above 😛

fangeugene (commented):
Thanks for the changes! Can you fix the failing tests (and add a test that tests your changes)?

ZachOrr:
We should still work on landing this - but for tracking, the change we made to cleanup dangling `DistrictTeam`/`EventTeam` keys helped us get out of this state https://github.com/the-blue-alliance/the-blue-alliance/pull/7012

PR #5328, opened 2023-06-13
Port clientapi ping to py3

phil-lopreiato:
This ports `ping` functionality in a way that we can test using the fullstack test framework. A few things:
 - change the existing test to be a bit more idempotent to simplify local development
 - add some jest arguments to run tests serially, so we can avoid tests conflicting state on the local devserver and bump the timeout 
 - implement a basic webhook receiver on a the `local` blueprint, so we can fetch and validate the notifications we receive.


ZachOrr:
Yay! Thanks!

Issue #5388, opened 2023-07-15
Deploy dispatch.yaml from py3 and cleanup cruft


phil-lopreiato:
We currently still deploy this out of the py2 repo. GH Actions [dropped support for python2](https://www.github.com/actions/setup-python/issues/672), so the deploy pipeline no longer works.

This issue will track getting a "clean" deploy of `dispatch.yaml` from the main py3 source tree

PR #5389, opened 2023-07-15
deploy dispatch.yaml from py3 repo

phil-lopreiato:
Towards https://github.com/the-blue-alliance/the-blue-alliance/issues/5388

GH Actions [dropped support for python2](https://www.github.com/actions/setup-python/issues/672); it's not worth investing in any more forward fixes.

For the time being, we need different dispatch rules for the local devserver and prod (for as long as we still have py2 serving endpoints), so we add a separate file which the CI knows how to handle at deploy-time. The contents were copied from the [`master` branch](https://github.com/the-blue-alliance/the-blue-alliance/blob/master/dispatch.yaml)

Once we remove the last few py2 URL rules, we can kill the service in GAE, and consolidate on a single dispatch file.

### DIFF
--- a/ops/prod_dispatch.yaml
+++ b/ops/prod_dispatch.yaml
@@ -1,0 +1,60 @@
+# This dispatch file applies to prod, not local dev
+# CI will renme this file appropriately and deploy it
+# The "default" module means "py2"
+# Cleaning this up is tracked in https://github.com/the-blue-alliance/the-blue-alliance/issues/5388
+
+dispatch:
+# Pin these pages to py2
+- url: "www.thebluealliance.com/_/account/*"
+  service: default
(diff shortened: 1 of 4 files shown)


ops/prod_dispatch.yaml
```diff
@@ -0,0 +1,60 @@
...
+- url: "*/tasks/*"
+  service: py3-tasks-io
+
+# API docs (swagger)
+- url: "*/swagger/*"
+  service: py3-web
+
+# mobile client API to py3
+- url: "*/clientapi/*"
+  service: py3-api
```
phil-lopreiato, line 41: This change is inherited from https://github.com/the-blue-alliance/the-blue-alliance/commit/3dafb6697bd9d5511afaf7240eab733bd11b26a6

PR #5395, opened 2023-07-15
Allow extra RP fields for 2023

lethosor:
## Description
This allows adding `extraRp1` and `extraRp2` fields to score breakdowns through the write API. These are not used by the FRC API, so we could consider prefixing them with `tba_` as well (I brought this up as an option on Slack but we didn't reach a consensus).

## Motivation and Context
Offseasons sometimes want to add extra ranking points for various things (e.g. the additional cargo threshold at 2022cc, rainbow link bonus at 2023mirr). Supporting breakdown fields for this purpose makes it easier to submit and retrieve data about these RPs, and also allows for disambiguation if multiple extra ranking points are used.

## How Has This Been Tested?
Submitted a couple matches through the write API on a local instance. Also verified that insights are unchanged with these fields added or removed from matches. (I would like to eventually support these in insights, but that will be a more involved change.)

## Types of changes

### DIFF
--- a/src/backend/common/helpers/score_breakdown_keys.py
+++ b/src/backend/common/helpers/score_breakdown_keys.py
@@ -136,6 +136,7 @@ VALID_BREAKDOWNS: Dict[Year, Set[str]] = {
             "foulPoints",
             "rp",
             "totalPoints",
+            # fields added by TBA:
             "tba_gameData",
         ]
     ),
@@ -352,6 +353,9 @@ VALID_BREAKDOWNS: Dict[Year, Set[str]] = {
             "teleopPoints",
             "totalChargeStationPoints",
             "totalPoints",
+            # fields added by TBA:
+            "tba_extraRp1",
+            "tba_extraRp2",
         ]
     ),
 }


src/backend/common/helpers/score_breakdown_keys.py
```diff
@@ -334,6 +334,8 @@
             "endGameChargeStationRobot3",
             "endGameParkPoints",
             "extraGamePieceCount",
+            "extraRp1",
+            "extraRp2",
```
phil-lopreiato, line 338: I'm okay with adding these, but I think the `tba_` prefix will be good to separate them from ones that you should expect from the official breakdown

phil-lopreiato (changes requested):
+ the `tba_` prefix, if you don't mind

fangeugene:
Note that these extra RP won't get factored into rankings (TBA doesn't calculate rankings directly), nor show up in any UI. In other words, the only place these would be exposed would be via our API. Is that the intent?

lethosor:
Done. Yeah, the intent is just to collect this data on the API side for now, and maybe render it later. Not sure if I will have time to work on rendering or insights before 2023mirr, but hopefully after. Events should be responsible for submitting accurate rankings and accurate `rp` for each match (which get shown e.g. at https://www.thebluealliance.com/match/2022cc_qm2)

phil-lopreiato (approved):
I think at the minimum, just storing the data and putting it back in the API is reasonable. We can chat more in slack about the best way to integrate into the UI/insights/etc

PR #5401, opened 2023-07-16
Support current-season computations in insights/district rankings

phil-lopreiato:
Turns out we never updated the years in `cron.yaml` (I checked the console :stuck_out_tongue: )
image

So, add a version of the URLs here which don't take an explicit year parameter and instead default to the current season. And let's use that instead, so we don't have to remember to change the cronjobs in the future :smile:


fangeugene:
Lol nice

PR #5417, opened 2023-07-23
MatchHelper.add_match_times: use specified weekdays to disambiguate dates, if present

### LINKED ISSUE #5416: Scheduled match times submitted through write API have wrong dates if the last match isn't on the last day of the event
2023mirr ran into this - all qual matches were scheduled for Saturday (July 22), but the event ends on Sunday (July 23). Due to old match_helper logic, only the times are parsed, and the last match is assumed to be on the last day of the event: https://github.com/the-blue-alliance/the-blue-alliance/blob/9c3eaf404258b71e84826d55390dad756304adab/src/backend/common/helpers/match_helper.py#L165-L169

## Expected Behavior
Match times should match the dates the matches are scheduled for.

## Current Behavior
Match times may show the wrong dates - example from above (these matches should be on Saturday):
<details>

image

</details>

Excerpt from schedule report:
<details>

image

</details>

## Possible Solution
Looking into this myself. I think we can parse weekdays from time_string fields to determine match dates more accurately.

## Steps to Reproduce (for bugs)
1. Upload the attached schedule report through the write API
[MIRR2023_ScheduleReport_Quals.xls](https://github.com/the-blue-alliance/the-blue-alliance/files/12137965/MIRR2023_ScheduleReport_Quals.xls)

lethosor:
## Description
When ingesting match `time_string` fields (currently only done through the write API), this change will attempt to match weekday abbreviations present in those fields with event dates. e.g. a time_string containing "Fri" will be assigned to the Friday of the event, if there is one.

## Motivation and Context
Fixes #5416. Previously, dates were determined by assigning the last match to the last day of the event and working backwards, which produces the wrong results if the last match submitted doesn't occur on the last day of the event. (This was old logic that accounted for the 2012-2013 era FIRST website, which didn't include days at all).

## How Has This Been Tested?
Unit tests added. Also uploaded the schedule from #5416 successfully in a test event (below).

image
image

## Types of changes


src/backend/common/helpers/tests/match_helper_test.py
```diff
@@ -271,6 +273,118 @@ def test_upcoming_matches_all_played(test_data_importer) -> None:
...
+                alliances_json=json.dumps(alliances),
+                time_string=row["time"],
+            )
+
+            matches.append(match)
+
+    return matches
+
+
+def test_add_match_times(test_data_importer):
```
lethosor, line 310: These first two tests came from `./old_py2/tests/test_add_match_times.py`. They relied on the old `UsfirstMatchesParser`, which hasn't been ported, to read the test data, so I decided moving it over to CSV would be simpler. Other tests are using `test_data_importer.parse_match_list()` -> `MatchConverter.dictToModel_v3()` to read test matches, but that does not appear to handle the `time_string` field, and I don't need the majority of fields here anyway. But open to other ideas that don't involve writing a custom parser for the tests.

Also, is the `UsfirstMatchesParser` likely to return? I am considering deleting the old `test_add_match_times`; wondering if I should also delete the corresponding test data (html files in `old_py2/test_data/usfirst_html/`)

lethosor:
I think this one is ready to go now

src/backend/common/helpers/tests/match_helper_test.py
phil-lopreiato: The HTML pages are long gone from FIRST's website, so I doubt we'll need them ever again.

phil-lopreiato (approved):
Thanks for this patch and tests! Sorry for the delay in review...

PR #5716, opened 2024-01-31
Add component opr api endpoint

tervay:
## Description
Adds `/event/{key}/coprs` API endpoint. This **does not** touch any Swagger stuff, so we will need to add that in a followup PR.

Note that all events from 2016-2022 should have their cOPRs generated at some point; these events are aggressively cached so they haven't ever actually run since cOPRs were introduced. An admin will need to do this.

## Motivation and Context
COPRs are currently only available via web interface. This helps scouts a lot and prevents the need for web scraping.

## How Has This Been Tested?
Manually with 2 events - one with cOPRs generated, one without (returns `{}` for events without cOPRs).

image

## Types of changes

### DIFF
--- a/src/backend/api/main.py
+++ b/src/backend/api/main.py
@@ -84,7 +84,7 @@ class ModelTypeConverter(BaseConverter):
 
 
 class EventDetailTypeConverter(BaseConverter):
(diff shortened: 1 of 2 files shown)


phil-lopreiato (approved):
Nice!

PR #5742, opened 2024-02-12
Updates backup team district points for 2024

AGPapa:
## Description
The 2024 rules changed how district points work for backup teams. Previously you would take all wins for a team, divide by wins for their alliance, and then multiply that fraction by the points the whole alliance earned throughout the playoffs. With the 2024 rules you only do that for the double elimination rounds, and the finals are treated separately as 5 points per win.

## Motivation and Context
The current system doesn't handle the new 2024 rules. A website I help run, frclocks.com, uses the TBA api to get district point values, so we'd like this to be correct before Week 1 events.

## How Has This Been Tested?
I've added a test case of a 2023 Ontario district where a backup team won a finals match. That event would be treated differently in the 2023 and 2024 rules. For the 2024 version I also added another backup team in the double elim rounds to make sure the functionality worked for those teams as well.

2023 Game manual:
Screenshot 2024-02-12 at 5 38 13 PM

2024 Game manual:
Screenshot 2024-02-12 at 5 37 15 PM

## Types of changes


phil-lopreiato (approved):
Thank you for the patch and for test coverage!

PR #5757, opened 2024-02-24
Update match_breakdown_2024.html

AADad578:
fixed incorrect if statement

## Description
Changed which variable the if looks for when determining whether or not to do a detailed account of speaker note counts

## Motivation and Context
The if statement would have hidden the speaker note counts if the amp note counts are missing, not if the speaker ones are. That could cause an error (probably) and is unintentional.

## How Has This Been Tested?
N/A

## Types of changes


phil-lopreiato (approved):
good catch; thanks!

PR #5761, opened 2024-02-26
Automatically remove DistrictTeam for teams without an event in a given year

dracco1993:
## Description
Right now, once a team is listed in a district, they will get new `DistrictTeam` records for every year after that.

## Motivation and Context
The result of this is that we list teams who are no longer competing in current district listings, which is very confusing, and incorrect.

The biggest functional change here is that each time we update the `team_details` for a team, we're checking all `DistrictTeam`s for that team to make sure that each year the team has a corresponding `EventTeam` i.e. they went to at least one event.  2020/2021 was a question mark for me, but any team that actually competed has the "remote" events, so we're good there.

I've added a new spec, just to cover this specific case, as well as updated the other specs to including having a `EventTeam` for that year, since that's now required.

In all of my research, this dependency on having `EventTeam`s before updating `team_details` _SEEMS_ okay, as events are updated/added way more often, as well as before teams, but this is something that some with a better understanding of the API calls/timing of such should probably verify/confirm.

## How Has This Been Tested?
Tested locally.
Wrote additional tests for the added functionality.

## Types of changes

### DIFF
--- a/src/backend/tasks_io/handlers/frc_api.py
+++ b/src/backend/tasks_io/handlers/frc_api.py
@@ -3,7 +3,7 @@ import logging
 import re
 from typing import List, Optional, Set
 
-from flask import Blueprint, Response, make_response, render_template, request, url_for
+from flask import Blueprint, make_response, render_template, request, Response, url_for
 from google.appengine.api import taskqueue
 from google.appengine.ext import ndb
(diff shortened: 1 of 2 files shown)


phil-lopreiato (approved):
This looks right to me; thanks for the fix! ❤️

PR #5782, opened 2024-03-05
Implement 2024 playoff tiebreakers

### LINKED ISSUE #5762: 2024 playoff ties don't show winner
[2024 ISR #1](https://www.thebluealliance.com/event/2024isde1) Playoff match 12 ended in a tie, with the second tie-breaker (autonomous points) used to break the tie. However, the event page does not reflect the winner, The winning score should be bolded and the playoff bracket should show a win for the team that won the tiebreaker.

image
image

lethosor:
2024 playoff tiebreaker logic was missing.

Fixes #5762

## How Has This Been Tested?
* Unit tests
* Local import of 2024miket and 2024isde1

Note: no matches have yet had a tie broken by SCALE points, so there is no test for that.

## Types of changes


phil-lopreiato (approved):
Thank you for the patch!

PR #5805, opened 2024-03-17
Check Match.push_sent in the match manipulator

ZachOrr:
Avoids using the `match.put` which was rightly pointed out is probably an issue due to how our cache works for the models

### DIFF
--- a/src/backend/common/manipulators/match_manipulator.py
+++ b/src/backend/common/manipulators/match_manipulator.py
@@ -84,13 +84,16 @@ def match_post_update_hook(updated_models: List[TUpdatedModel[Match]]) -> None:
         event = match.event.get()
         # Only continue if the event is currently happening
         if event and event.now:
-            if match.has_been_played:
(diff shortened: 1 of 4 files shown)


src/backend/common/manipulators/match_manipulator.py
```diff
@@ -84,13 +84,16 @@ def match_post_update_hook(updated_models: List[TUpdatedModel[Match]]) -> None:
...
+            if match.has_been_played and not match.push_sent:
                 if (
                     updated_match.is_new
                     or "alliances_json" in updated_match.updated_attrs
                 ):
                     try:
                         TBANSHelper.match_score(match)
+                        # Update score sent boolean on Match object to make sure we only send a notification once
+                        match.push_sent = True
+                        MatchManipulator.createOrUpdate(match)
```
phil-lopreiato, line 96: Set `run_post_update_hooks=False` here too

PR #5822, opened 2024-03-20
Added helpful error responses to webhook test notifications

pepper-lees:
When a webhook test notification fails, the alert div below the button will now give more specific errors.

## Description

I modified apidocs_webhooks.html to display all errors in the alert div below the button instead of using `alert()` for some, and I modified apidocs.py to send JSON responses for specific errors, e.g. invalid or not specified match keys.

## Motivation and Context

Unfortunately this still doesn't help with the CSRF Expired error that occurs on the TBA prod site, since the error object from a failed authentication call returns HTML, and it'd be unsafe to set the HTML inside the alert div from a web response. However, this will help in the case where there's an error inside apidocs.py `apidocs_webhooks_notification` function.

## How Has This Been Tested?

Tested by putting valid and invalid match / event keys into the webhook testing page.

Example error message: (on my local database, I don't have events from 2014, so it shows a 'not found')
image

## Types of changes


fangeugene (approved):
Sorry this got overlooked earlier this season. This is amazing thanks so much

fangeugene (dismissed):

fangeugene (approved):

PR #5828, opened 2024-03-20
Apidocs webhooks fixes

pepper-lees:
1. The alert div below the send test notification button now provides more helpful errors
2. Fixed the 400 Unauthorized error when sending test notifications

## Description

For a while (at least a year), sending test webhook notifications has been broken. I discovered today that it was caused by the page caching, wherein the CSRF token rendered into the page belonged to whoever happened to load the page at the time it was re-rendered and cached, and would eventually become stale and invalid. I discussed with @ZachOrr and we came to the conclusion that the testing functions should be placed behind a privileged page. I gave the cached variant of the page a "Log in to send test notifications" button, which takes you to the authenticated variant of the page which has cache disabled, allowing the CSRF token being rendered to be accurate/up-to-date, fixing the 400 CSRF token expired error.

Second, I improved the error handling. If there's a user error involving an incorrect or not-specified key, the alert div below the button will show the specific error. And if there's an unexpected error (e.g. network error, or, god forbid, a CSRF issue), it'll display "Something went horribly wrong. Refresh the page and try again." in the same box instead of with an `alert()`.

## Motivation and Context

Webhook test notifications ain't workin!

## How Has This Been Tested?

I tested the error handling functionality by putting in incorrect or empty keys when sending test notifications. The CSRF token stuff works fine in my environment, but unfortunately, the webhook notifications don't actually _send_ from my development instance. I haven't modified any of the code responsible for the webhook sending code, though, so I suspect it'll work in production. But it'll have to be tested in the production environment just to be sure. However, since webhook test notifications have been broken for over a year, the risk involved in the possibility webhook test notifications not working with this code change is zero.

## Types of changes

### DIFF
--- a/src/backend/web/handlers/apidocs.py
+++ b/src/backend/web/handlers/apidocs.py
@@ -1,10 +1,9 @@
 from datetime import timedelta
 
-from flask import Blueprint, make_response, redirect, request, Response, url_for
+from flask import Blueprint, jsonify, make_response, redirect, request, Response, url_for
 from pyre_extensions import none_throws
 from werkzeug.wrappers import Response as WerkzeugResponse
 
-from backend.api.handlers.helpers.profiled_jsonify import profiled_jsonify
 from backend.common.auth import current_user
 from backend.common.consts.notification_type import (
     ENABLED_NOTIFICATIONS,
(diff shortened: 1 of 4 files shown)


src/backend/web/handlers/apidocs.py
ZachOrr, line 1: Can we get tests for these changes?

ZachOrr (commented):
This is looking cool! Pulled and looked at this locally. Let's try to get the errors fixed up and we can work on landing this

pepper-lees:
Closing, as this issue appears to have been independently fixed (much better than my implementation, too)

PR #5837, opened 2024-03-22
Remove team normalized location uses

ZachOrr:
We stopped normalizing team locations a while ago. This pull request removes the code that used to call the Google API to normalize team locations.


src/backend/common/helpers/location_helper.py
```diff
@@ -207,61 +207,6 @@ def compute_event_location_score(
 
         return score
 
-    @classmethod
```
fangeugene, line 210: Any chance we can keep this but not expose it anywhere?

ZachOrr:
Closing - a lot of this already made it's way into prod.

PR #5852, opened 2024-03-26
Work around matches with scheduled teams [1, 2,3]

phil-lopreiato:
FMS seems to be publishing unplayed finals matches with the placeholder teams now (when teamNumber would be None previously).

So now, filter those out

Two currently broken events: 

`2024vafal`
Prod:
image

Local:
image

`2024nytr`
Prod:
image

Local:
image

### DIFF
--- a/src/backend/tasks_io/datafeeds/datafeed_fms_api.py
+++ b/src/backend/tasks_io/datafeeds/datafeed_fms_api.py
@@ -321,22 +321,33 @@ class DatafeedFMSAPI:
 
     @classmethod
     def _merge_match(cls, scheduled: Dict, result: Dict) -> Dict:
+        # Over the years, both "Teams" and "teams" have been used in FRC API responses...
+        teams_key = "Teams" if "Teams" in scheduled else "teams"
+
+        # 2024, Week 4: As part of attempting to restore sync,
+        # schedules sync with teams [1, 2, 3] in to-be-played playoff matches
+        # In a case where the only teams for a match are those three, overwrite
+        # the team numbers in each station to None
+        schedule_team_numbers = [t["teamNumber"] for t in scheduled.get(teams_key)]
+        if set(schedule_team_numbers) == {1, 2, 3}:
+            scheduled["teams"] = [{**t, "teamNumber": None} for t in scheduled.get(teams_key)]
+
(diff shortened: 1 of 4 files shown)


src/backend/tasks_io/datafeeds/datafeed_fms_api.py
```diff
@@ -321,22 +321,30 @@ def _merge_match_schedule_and_results(self, schedule: Dict, matches: Dict) -> Di
...
                         ),
-                        None,
+                        (None, None),
                     )
                     if schedule_team is None:
                         # 2024, Week 3: Upstream FMS sync issues leading to schedules returned with no teams
                         # Some match results have been sync'd, so patch around this where we can
                         scheduled["teams"].append(team)
                     else:
-                        scheduled["teams"][i].update(team)
```
phil-lopreiato, line 339: For cases where an alliance uses a lineup card and uses a non-default station assignment, we want to make sure we call `update` on the corresponding team number, which can update the station assignment

(for this reason, we extract the index in the `scheduled` list above)

src/backend/tasks_io/datafeeds/parsers/fms_api/fms_api_match_parser.py
```diff
@@ -121,9 +120,19 @@ def parse(
...
             )
+
+            null_team = any(t["teamNumber"] is None for t in sorted_teams)
+            if (
+                null_team
+                and match["scoreRedFinal"] is None
+                and match["scoreBlueFinal"] is None
+            ):
+                continue
+
```
phil-lopreiato, line 131: This logic moves up, so we can more clearly skip over unplayed matches without teams configured

src/backend/tasks_io/datafeeds/parsers/fms_api/fms_api_match_parser.py
```diff
@@ -121,9 +120,19 @@ def parse(
...
             for team in sorted_teams:
                 if team["teamNumber"] is None:
-                    null_team = True
+                    continue
```
phil-lopreiato, line 134: ...which allows us to simply skip over individual teams with a `None` team number (for cases where the whole list has a mix of `None` and non-`None` teams)

PR #5899, opened 2024-04-23
Cross-compile dev image for arm64

devyntk:
Increases the docker image build time by... a fair bit, as it's cross-compiling this from amd64 to arm64. I think this is a fair substitute to make it a bit easier for folks with Mx macs get up and running easier (and also helping me get around some work laptop funkiness with SSL).


ZachOrr:
Lemme try to give this a look soon! I'm on an ARM Mac so - I suspect if we pulled the pre-compiled from the repo this would speed up my local instance. Just need to see why the checks are failing. Also cc @phil-lopreiato - I'm not necessarily the domain expert here

fangeugene (approved):

PR #5937, opened 2024-05-15
Add alliance details to team detail page

devyntk:
This PR provides details about Alliance Selection and that Alliance's progression. 

## Motivation and Context
I often find myself switching back and forth between event pages and alliance pages to determine an team's alliance, and find that the data should just be presented inline.

## How Has This Been Tested?
There seem to be no automated tests for the team renderer, so no automated tests were done. Manual tests were done with:
- District teams
- Regional teams
- Teams that were listed as backups, both in the backup field as well as a fake third pick
- Events that have specialized alliance names

image
image
image

## Types of changes

### DIFF
--- a/src/backend/common/helpers/alliance_helper.py
+++ b/src/backend/common/helpers/alliance_helper.py
@@ -1,0 +1,183 @@
+from typing import Dict, List, Optional, Tuple
+
+from pyre_extensions import none_throws
+
+from backend.common.consts.comp_level import COMP_LEVELS_VERBOSE_FULL, CompLevel
+from backend.common.consts.event_type import EventType
+from backend.common.consts.playoff_type import DoubleElimRound, PlayoffType
+from backend.common.models.alliance import (
+    EventAlliance,
+    PlayoffAllianceStatus,
+    PlayoffOutcome,
+)
+from backend.common.models.event import Event
+from backend.common.models.keys import TeamKey
+
+
+class AllianceHelper:
+    UNKNOWN_ALLIANCE_SIZE = 99
+
+    KNOWN_MODERN_EVENT_TYPE_ALLIANCE_SIZES: Dict[EventType, int] = {
+        EventType.REGIONAL: 3,
+        EventType.DISTRICT: 3,
+        EventType.DISTRICT_CMP: 3,
+        EventType.DISTRICT_CMP_DIVISION: 3,
+        EventType.CMP_DIVISION: 4,
+        EventType.CMP_FINALS: 4,
+    }
+
+    @classmethod
+    def get_known_alliance_size(
+        cls,
+        event_type: EventType,
+        year: int,
+    ) -> int:
+
+        if year < 2014:
+            # 2014 was first year of 4 team champs divisions
+            # for now, return default for legacy events
+            return cls.UNKNOWN_ALLIANCE_SIZE
+
+        return cls.KNOWN_MODERN_EVENT_TYPE_ALLIANCE_SIZES.get(
+            event_type, cls.UNKNOWN_ALLIANCE_SIZE
+        )
+
+    @staticmethod
+    def get_ordinal_pick_from_number(n: int) -> str:
+        # https://stackoverflow.com/a/20007730
+        if 11 <= (n % 100) <= 13:
+            suffix = "th"
+        else:
+            suffix = ["th", "st", "nd", "rd", "th"][min(n % 10, 4)]
+        return f"{n}{suffix} Pick"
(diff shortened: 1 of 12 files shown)


src/backend/common/models/alliance.py
```diff
@@ -1,5 +1,6 @@
 from typing import List, TypedDict
 
+from backend.common.models.event_team_status import EventTeamStatusPlayoff
```
fangeugene, line 3: I believe this create a circular import error with `src/backend/common/models/event_team_status.py`

src/backend/common/models/alliance.py
```diff
@@ -10,6 +11,7 @@ class _EventAllianceOptional(TypedDict, total=False):
     declines: List[TeamKey]
     name: str
     backup: EventAllianceBackup
+    status: EventTeamStatusPlayoff
```
fangeugene, line 14: Is this necessary? I don't see where it is being used.

src/backend/web/renderers/team_renderer.py
```diff
@@ -182,8 +183,55 @@ def render_team_details(
                     None,
                 )
 
+            alliance_size = KNOWN_EVENT_TYPE_ALLIANCE_SIZES.get(
+                event.event_type_enum, 99
+            )
+
+            alliance = None
+            alliance_pick = None
+            if event.details and event.details.alliance_selections:
```
fangeugene, line 192: It'd be nice if this could get broken out into a helper function (e.g. in a new file like `alliance_helper.py`) since there's some quite involved logic here that would be nice to unit test separately.

src/backend/common/models/alliance.py
devyntk: This is a change to make the python model match what's in the DB. Without it, the type check fails: https://github.com/the-blue-alliance/the-blue-alliance/actions/runs/9097667894/job/25006007976

This seems to be the "intended" material of the alliance data, as even the test data includes it (see https://github.com/the-blue-alliance/the-blue-alliance/blob/py3/src/backend/common/helpers/tests/data/2024test_alliances.json and other backend helper data), but I can't for the life of me figure out where the data is coming from.

src/backend/common/models/alliance.py
devyntk: The tests are current failing because of this, but I don't want to invest too much time into refactoring this to avoid the circular import until the issue of whether this is the correct data structure (see other comment) is resolved more.

src/backend/common/models/alliance.py
fangeugene: Oh i see. yeah this must've been overlooked originally somehow.

src/backend/web/templates/team_details.html
```diff
@@ -188,7 +235,7 @@ <h4>District Points Earned</h4>
                     <p class="muted">No matches played.</p>
                   {% else %}
                     <p class="muted">No matches yet. Check back after {{ comp.event.start_date|strftime("%B %d, %Y") }}.</p>
-                  {% endif %}
+                  {% endif %}
```
fangeugene, line 238: stray space?

src/backend/web/renderers/team_renderer.py
```diff
@@ -182,8 +184,32 @@ def render_team_details(
                     None,
                 )
 
+            alliance_size = AllianceHelper.get_known_alliance_size(
+                event.event_type_enum, event.year
+            )
+
+            alliance, alliance_pick = AllianceHelper.get_alliance_details_and_pick_name(
```
fangeugene, line 191: `get_known_alliance_size` is called within this helper. Would it make sense to remove the redundant call above and have `get_alliance_details_and_pick_name` also return `alliance_size`?

src/backend/common/helpers/alliance_helper.py
```diff
@@ -0,0 +1,181 @@
...
+    ) -> Tuple[Optional[EventAlliance], Optional[str]]:
+        alliance = None
+        alliance_pick = None
+
+        if event.alliance_selections is None:
+            return (None, None)
+
+        alliance_size = cls.get_known_alliance_size(event.event_type_enum, event.year)
+
+        for alliance_check in none_throws(event.alliance_selections):
```
fangeugene, line 66: if either of the `if` statements below match, we should be able to break out of the for loop right?

fangeugene (commented):
This is awesome. The new typing renaming and organization make a whole lot of sense. A few questions/comments

fangeugene (approved):
🥳

PR #5981, opened 2024-06-06
Hotfix/generate playoff status at runtime

devyntk:
## Types of changes

### DIFF
--- a/src/backend/web/renderers/team_renderer.py
+++ b/src/backend/web/renderers/team_renderer.py
@@ -188,7 +188,7 @@ class TeamRenderer:
                 AllianceHelper.get_alliance_details_and_pick_name(event, team.key_name)
             )
 
-            if alliance and "status" in alliance and "name" n alliance:
+            if alliance and "status" in alliance and "name" in alliance:
                 alliance_status = " and ".join(
                     AllianceHelper.generate_playoff_status_string(
                         alliance["status"],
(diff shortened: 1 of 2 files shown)


src/backend/web/renderers/team_renderer.py
```diff
@@ -188,10 +189,18 @@ def render_team_details(
                 AllianceHelper.get_alliance_details_and_pick_name(event, team.key_name)
             )
 
-            if alliance and "status" in alliance and "name" in alliance:
+            alliance_status = EventTeamStatusHelper._build_playoff_info(
+                team.key_name,
+                event.details,
+                event_matches,
```
fangeugene, line 195: `_build_playoff_info` expects `event_matches` needs to be matches for the whole event, but i believe this only has matches for this current team?

devyntk: Since this is only generating the status based off of that team's alliance, I think that would be fine? I'll look more into the event team status helper.

fangeugene: I think the edge case is if a team is a 3 pick and doesn't play any matches?

fangeugene: I think the reason why we were generating this on the fly in the API rather than saving it in the database was so we could change the strings without going through crazy cash clearing

devyntk: I don't think it makes sense to store the strings in a database, but the various status dicts (espcially now that this dict is both used for team event statuses and the team detail pages) may make sense to store and/or cache in some way.

PR #6046, opened 2024-06-28
Allow passing webcast `date` field to trusted API

lethosor:
## Description
This allows requests to the write API to provide dates for webcasts.

Note: 
* unsure what documentation updates are necessary. The `webcasts` field was not fully described in the API docs to begin with.
* also unsure what validation is necessary here. Suggestions do not appear to perform any validation on the dates either.

## Motivation and Context
Webcast suggestions can provide the `date` field, but write API requests cannot.

## How Has This Been Tested?
Local testing through write API, both with `url` and `type`+`channel` payloads.

## Types of changes

### DIFF
--- a/src/backend/api/handlers/tests/update_event_info_test.py
+++ b/src/backend/api/handlers/tests/update_event_info_test.py
@@ -1,6 +1,7 @@
 import json
 from typing import Dict, List, Optional
 
+import pytest
 from google.appengine.ext import ndb, testbed
 from werkzeug.test import Client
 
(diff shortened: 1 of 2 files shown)


fangeugene:
> also unsure what validation is necessary here. Suggestions do not appear to perform any validation on the dates either.

You're right, the validation around this is pretty poor. Suggestions rely on mods to ensure the date is formatted correctly. The expected date format is `YYYY-MM-DD`.

src/backend/api/handlers/tests/update_event_info_test.py
```diff
@@ -141,6 +145,38 @@ def test_update_event_info(
...
+)
+def test_invalid_webcasts_date(
+    invalid_webcast: dict,
+    taskqueue_stub: testbed.taskqueue_stub.TaskQueueServiceStub,
+    api_client: Client,
+) -> None:
+    setup_event()
+    setup_auth(access_types=[AuthType.EVENT_INFO])
+
+    # Test invalid remap_teams
```
fangeugene, line 164: Can you fix this comment? Looks copied from elsewhere.

fangeugene (commented):
Thanks! One minor nit.

fangeugene (approved):
Thanks!

PR #6049, opened 2024-07-02
Add insights API converter, queries, etc. Add "notables" insights API endpoint.;

tervay:
## Description
This adds a new API category, `/insights/`, which currently has one endpoint - `/insights/notables`. This endpoint returns a dictionary that is currently the following shape:

```json
{
  "cmp_winners": {
    ...
    "2023": [
      "frc1323",
      "frc2609",
      "frc4096",
      "frc4414"
    ],
    "2024": [
      "frc321",
      "frc1690",
      "frc4522",
      "frc9432"
    ]
  },
  "hof": {
    ...
    "2019": [
      "frc1816",
      "frc1902"
    ],
    "2021": [
      "frc503",
      "frc4613"
    ],
    "2022": [
      "frc1629"
    ],
    ...
  }
}
```

This endpoint and API route is extremely likely to change in the short term and as such, I don't recommend documenting it via OpenAPI at this time.

In order to get the HOF winners, the `CA_WINNER` insight needed to be fixed. Previously, it was only assigning one team per year. The insight's data format is now a list of team keys rather than a single team key. This does not affect anything else in the repo as literally nothing used it and it hasn't been touched in [over a decade](https://github.com/the-blue-alliance/the-blue-alliance/commit/b890c1585815a2a47a36aefbce9db70042ae2bfe).

## Motivation and Context
Insights are currently not exposed in the API at all. Exposing them can do two things:
1. Help third party tools retrieve stats that TBA already calculates, like most matches played, HOF, etc
2. Provide an avenue for the experimental PWA frontend to access insights

## How Has This Been Tested?
Tested locally with bootstrapped 2015-2024 data.

image

## Types of changes

### DIFF
--- a/src/backend/api/handlers/insights.py
+++ b/src/backend/api/handlers/insights.py
@@ -1,0 +1,85 @@
+from flask import Response
+
+from backend.api.handlers.decorators import api_authenticated
+from backend.api.handlers.helpers.profiled_jsonify import profiled_jsonify
+from backend.api.handlers.helpers.track_call import track_call_after_response
+from backend.common.consts.api_version import ApiMajorVersion
+from backend.common.consts.insight_type import LEADERBOARD_INSIGHTS, NOTABLES_INSIGHTS
+from backend.common.decorators import cached_public
+from backend.common.models.insight import Insight
+from backend.common.queries.insight_query import (
+    InsightsByNameAndYearQuery,
+    InsightsByNameQuery,
+)
+
+
+@cached_public
+def insights_leaderboards_all() -> Response:
+    track_call_after_response("insights/leaderboards/all")
+    futures = []
+    for insight_type in LEADERBOARD_INSIGHTS:
+        futures.append(
+            InsightsByNameQuery(
+                insight_name=Insight.INSIGHT_NAMES[insight_type]
+            ).fetch_dict_async(ApiMajorVersion.API_V3)
+        )
+
+    insights = []
+    for future in futures:
+        insights.extend(future.get_result())
+
+    return profiled_jsonify(insights)
+
+
+@cached_public
+def insights_leaderboards_year(year: int) -> Response:
+    track_call_after_response("insights/leaderboards", year)
+    futures = []
+    for insight_type in LEADERBOARD_INSIGHTS:
+        futures.append(
+            InsightsByNameAndYearQuery(
+                insight_name=Insight.INSIGHT_NAMES[insight_type], year=year
+            ).fetch_dict_async(ApiMajorVersion.API_V3)
+        )
+
+    insights = []
+    for future in futures:
+        insights.extend(future.get_result())
+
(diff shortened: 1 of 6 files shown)


src/backend/common/queries/dict_converters/insight_converter.py
```diff
@@ -0,0 +1,37 @@
...
+    @classmethod
+    def insightsConverter_v3(cls, insights: List[Insight]) -> List[InsightDict]:
+        return list(map(cls.insightConverter_v3, insights))
+
+    @classmethod
+    def insightConverter_v3(cls, insight: Insight) -> InsightDict:
+        return InsightDict(
+            {
+                "name": insight.name,
+                "data_json": json.loads(insight.data_json),
```
fangeugene, line 34: should this just be `"data": insight.data`?

Also the contents of data isn't well enforced / consistent across insights. Not sure how that will affect the openapi definition in the future. It might be worth going through and strongly typing the insights data (doesn't have to be in this PR, but worth keeping in mind).

src/backend/api/handlers/insights.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Returns data about teams that may be:
+    - HOF
+    - CMP winners
+    - WFA winners (todo)
+    """
+    track_call_after_response("insights/notables")
+    hof = {
+        insight_dict["year"]: insight_dict["data_json"]
+        for insight_dict in InsightAllYearsQuery(
```
fangeugene, line 25: `InsightAllYearsQuery`, as implemented, queries for all year-specific insights, not the aggregated insights across all years. Is that what you want?

src/backend/common/queries/insight_query.py
```diff
@@ -0,0 +1,43 @@
...
+    CACHE_VERSION = 1
+    CACHE_KEY_FORMAT = "insights_{insight_name}"
+    DICT_CONVERTER = InsightConverter
+
+    def __init__(self, insight_name: str) -> None:
+        super().__init__(insight_name=insight_name)
+
+    @typed_tasklet
+    def _query_async(self, insight_name: str) -> Generator[Any, Any, List[InsightDict]]:
+        insights = yield Insight.query(Insight.name == insight_name).fetch_async()
```
fangeugene, line 42: If you're looking for overall aggregated insights, you'd want to include `Insight.year == 0`

src/backend/common/queries/dict_converters/insight_converter.py
tervay: I bet we could have one generic category of "top N" metrics, which are all basically the same shape - most banners, most matches played, etc are all the same effective shape. That schema would cover most of them. Then there are others that are more rare, like HOF or CMP winners, which I bet we could generify too.

tervay:
I messed up my local copy and #6207 is now the right PR. whoops

PR #6071, opened 2024-07-17
Add proper nullables, add score breakdown 2024 to API V3 Spec

tervay:
## Description
- Adds score breakdown 2024
- Links score breakdown objects to score_breakdown field
- Adds null to possible values of fields that may be null

## Motivation and Context
Code generators that use the spec generate incorrect types and leads to a lot of headache.

## How Has This Been Tested?
I don't know how to test this

## Screenshots (if appropriate):
image
image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would change API specifications or require data migrations)

PR #6099, opened 2024-07-30
Add TS API bindings for PWA

tervay:
## Description
Adds TS bindings for fetching data from TBA utilizing [oazapfts](https://github.com/oazapfts/oazapfts). During my testing I found this library to be the most robust and flexible for generating TS OpenAPI clients.

## Types of changes

### DIFF
--- a/pwa/README.md
+++ b/pwa/README.md
@@ -10,6 +10,12 @@ First install the dependencies:
 npm i
 ```
 
+Make sure you have your TBA APIv3 Read Key set in `.env`:
+
+```sh
+VITE_TBA_API_READ_KEY="myKey"
+```
+
 Run the dev server:
 
 ```shellscript
--- a/pwa/package.json
+++ b/pwa/package.json
@@ -10,12 +10,14 @@
     "format:check": "prettier --check .",
     "format:fix": "prettier --write .",
     "start": "remix-serve ./build/server/index.js",
-    "typecheck": "tsc"
+    "typecheck": "tsc",
+    "generate-api": "oazapfts ../src/backend/web/static/swagger/api_v3.json --optimistic --argumentStyle=object > app/api/v3.ts"
   },
   "dependencies": {
     "@remix-run/node": "^2.11.1",
     "@remix-run/react": "^2.11.0",
     "@remix-run/serve": "^2.11.1",
+    "@oazapfts/runtime": "^1.0.3",
     "isbot": "^4.1.0",
     "react": "^18.2.0",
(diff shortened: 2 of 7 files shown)


pwa/README.md
```diff
@@ -4,6 +4,12 @@
 
 ## Development
 
+Make sure you have your TBA APIv3 Read Key set in `.env`:
+
+```sh
+VITE_TBA_API_READ_KEY="myKey"
```
fangeugene, line 10: I will work on getting this key set up with CI deployments

pwa/package.json
```diff
@@ -10,9 +10,11 @@
     "format:check": "prettier --check .",
     "format:fix": "prettier --write .",
     "start": "remix-serve ./build/server/index.js",
-    "typecheck": "tsc"
+    "typecheck": "tsc",
+    "generate-api": "oazapfts ../src/backend/web/static/swagger/api_v3.json --optimistic --argumentStyle=object > app/api/v3.ts"
```
fangeugene, line 14: Does this need to get regenerated after #6071 landed?

pwa/package.json
```diff
@@ -34,6 +36,7 @@
     "eslint-plugin-jsx-a11y": "^6.7.1",
     "eslint-plugin-react": "^7.33.2",
     "eslint-plugin-react-hooks": "^4.6.0",
+    "oazapfts": "^6.1.0",
```
fangeugene, line 39: I haven't used this in the past, but I've used `openapi-typescript`. Did you look into the differences between the two? I don't have strong opinions either way but `openapi-typescript` looks to be used more.

tervay: Yeah I tried out both. `openapi-typescript` is definitely more popular BUT for our specific API I think it's much more clunky:

```ts
const { data, error } = await client.GET("/team/{team_key}", {
  params: {
    path: { team_key: "frc604" },
  },
});
```

It's _really_ verbose. `oazapfts` is simply `await client.getTeam("frc604");`

Let me do one more round of digging into this to make sure I haven't missed anything with `openapi-typescript`.

tervay: Took another look and not seeing a way to do it with `openapi-typescript` that makes logical sense and is easy to use. So I think oazapfts is the way to go.

fangeugene: Fair enough. Can you fix the package.json conflicts then i can merge

pwa/package.json
tervay: yes

tervay: done

PR #6147, opened 2024-08-09
Add match table component

tervay:
Adds new tailwind-based match table. Misses a lot of edge cases (e.g. round robins) but this gets the core UI implemented.

image

## Types of changes

### DIFF
--- a/pwa/app/components/tba/matchResultsTable.tsx
+++ b/pwa/app/components/tba/matchResultsTable.tsx
@@ -1,0 +1,244 @@
+import { Link } from '@remix-run/react';
+import { type VariantProps, cva } from 'class-variance-authority';
+import type React from 'react';
+import { Fragment } from 'react';
+import { Match } from '~/api/v3';
+import {
+  cn,
+  sortMatchComparator,
+  timestampsAreOnDifferentDays,
+  zip,
+} from '~/lib/utils';
+import PlayCircle from '~icons/bi/play-circle';
+
+const cellVariants = cva('', {
+  variants: {
+    matchResult: {
+      winner: 'font-semibold',
+      loser: '',
+    },
+    allianceColor: {
+      red: 'bg-alliance-red-light',
+      blue: 'bg-alliance-blue-light',
+    },
+    teamOrScore: {
+      team: '',
+      score: '',
+    },
+  },
+  defaultVariants: {
+    matchResult: 'loser',
+    allianceColor: undefined,
+    teamOrScore: 'team',
+  },
+  compoundVariants: [
+    {
+      allianceColor: 'red',
+      teamOrScore: 'score',
+      class: 'bg-alliance-red-dark',
+    },
+    {
+      allianceColor: 'blue',
+      teamOrScore: 'score',
+      class: 'bg-alliance-blue-dark',
+    },
+  ],
+});
+
+interface CellProps
+  extends React.HTMLAttributes<HTMLDivElement>,
+    VariantProps<typeof cellVariants> {
+  dq?: boolean;
+  surrogate?: boolean;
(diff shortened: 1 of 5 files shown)


pwa/app/lib/utils.ts
```diff
@@ -8,3 +9,58 @@ export function cn(...inputs: ClassValue[]) {
...
+    return (
+      (compLevelValues[a.comp_level] ?? 0) -
+      (compLevelValues[b.comp_level] ?? 0)
+    );
+  }
+
+  return a.set_number - b.set_number || a.match_number - b.match_number;
+}
+
+export function sortMatches(matches: Match[]) {
```
fangeugene, line 31: Is this used?

pwa/app/lib/utils.ts
```diff
@@ -8,3 +9,58 @@ export function cn(...inputs: ClassValue[]) {
...
+  return matches.sort(sortMatchComparator);
+}
+
+export function timestampsAreOnDifferentDays(
```
fangeugene, line 35: This might break since timestamps are in UTC, not the event's timezone.

pwa/app/components/tba/matchResultsTable.tsx
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  if (match.comp_level === 'qm') {
+    return `Quals ${match.match_number}`;
+  }
+
+  return `Match ${match.set_number}`;
+}
+
+function maybeGetFirstTeamVideoURL(match: Match): string | undefined {
```
fangeugene, line 90: `maybeGetFirstMatchVideoURL`?

pwa/app/lib/utils.ts
tervay: Not currently, it was originally but then I just wound up using the comparator function instead. I'll remove

pwa/app/components/tba/matchResultsTable.tsx
```diff
@@ -0,0 +1,239 @@
...
+function matchTitle(match: Match): string {
+  if (match.comp_level === 'f') {
+    return `Finals ${match.match_number}`;
+  }
+
+  if (match.comp_level === 'qm') {
+    return `Quals ${match.match_number}`;
+  }
+
+  return `Match ${match.set_number}`;
```
fangeugene, line 87: Throw in a TODO to support single-elim match titles?

pwa/app/components/tba/matchResultsTable.tsx
tervay: This whole function should probably be obliterated with something more intelligent (I did not feel like doing the media type -> URL logic)

pwa/app/lib/utils.ts
tervay: at the time of writing, timezones were a future me problem

tervay: I put in a todo at the caller location in the component for now

pwa/app/components/tba/matchResultsTable.tsx
```diff
@@ -0,0 +1,241 @@
...
+function maybeGetFirstMatchVideoURL(match: Match): string | undefined {
+  if (match.videos === undefined || match.videos.length === 0) {
+    return undefined;
+  }
+
+  return `https://www.youtube.com/watch?v=${match.videos[0].key}`;
+}
+
+// tailwindgen.com is recommended if you're editing the grid classes
+export default function MatchResultsTable(props: {
```
fangeugene, line 100: Can we destructure `props` here? i.e. `MatchResultsTable({matches, title}:{ matches: Match[], title: string})`

PR #6150, opened 2024-08-09
Pin deps to specific versions rather than ^

tervay:


ofekashery:
@tervay why are we pinning the versions?

PR #6191, opened 2024-08-09
Remove motto from public view (keeps col in db)

tervay:


src/backend/common/queries/dict_converters/team_converter.py
```diff
@@ -35,7 +35,6 @@ def teamConverter_v3(cls, team: Team) -> TeamDict:
             "name": team.name if team.name else default_name,
             "website": team.website,
             "rookie_year": team.rookie_year,
-            "motto": None,
```
fangeugene, line 38: Can we keep this here? I don't want to break anyone's implementation that's expecting the field to be there even if it's null.

PR #6200, opened 2024-08-10
More API spec fixes

tervay:
god bless [jq](https://jqkungfu.com/)

### DIFF
--- a/src/backend/web/static/swagger/api_v3.json
+++ b/src/backend/web/static/swagger/api_v3.json
@@ -2318,7 +2318,8 @@
                   "type": "array",
                   "items": {
                     "$ref": "#/components/schemas/Elimination_Alliance"
-                  }
+                  },
+                  "nullable": true
                 }
               }
             }
@@ -2372,7 +2373,14 @@
             "content": {
               "application/json": {
                 "schema": {
-                  "$ref": "#/components/schemas/Event_Insights"
+                  "oneOf": [
+                    {
+                      "$ref": "#/components/schemas/Event_Insights"
+                    },
+                    {
+                      "type": "null"
+                    }
+                  ]
                 }
               }
             }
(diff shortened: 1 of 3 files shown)


fangeugene (commented):
I'm not sure if this 100% holds true. Sometimes Cheesy Arena doesn't implement all the breakdowns.

The red/blue requirement should definitely be true though.

tervay:
Oof. Ok. I'll write a script to go through every match since 2015 and see which fields ever get missed.

fangeugene:
Actually you might be able to just use `src/backend/common/helpers/score_breakdown_keys.py`

src/backend/web/static/swagger/api_v3.json
```diff
@@ -5846,6 +5846,12 @@
       "Match_Score_Breakdown_2015": {
         "type": "object",
         "properties": {
+          "required": [
```
fangeugene, line 5849: I think this is tabbed wrong. It should be outside of "properties" right?

tervay: whoops

src/backend/web/static/swagger/api_v3.json
```diff
@@ -7445,6 +7529,11 @@
           },
           "backup": {
             "type": "object",
+            "nullable": true,
```
fangeugene, line 7532: is `backup` really both nullable and not required?

tervay: Yeah.

Undefined in offseason: https://www.thebluealliance.com/api/v3/event/2022bc/alliances
Undefined in season: https://www.thebluealliance.com/api/v3/event/2022mabos/alliances

Nullable in season: https://www.thebluealliance.com/api/v3/event/2018nyut/alliances

Seems like if an event has a backup on any alliance, every other alliance object has a backup field present, which may be null or an actual backup

fangeugene: amazing

PR #6201, opened 2024-08-10
Add playoff matches. Add groundwork for generic match table rendering per event type.

tervay:
image
image

This adds the `app/components/tba/matchResultsTables/` directory which is a place we can put components for various elim bracket components. When single elim is implemented there, we should add another component that reads from `event.event_type` to use the correct component for the given `Match[]` array.


pwa/app/components/tba/matchResultsTables/quals.tsx
```diff
@@ -0,0 +1,18 @@
...
+import MatchResultsTableBase from '~/components/tba/matchResultsTables/base';
+
+export default function MatchResultsTableQuals({
+  matches,
+}: {
+  matches: Match[];
+}) {
+  return (
+    <div>
+      <div className="mb-2 text-3xl font-bold">Quals</div>
```
fangeugene, line 11: `text-2xl` to match playoffs?

pwa/app/components/tba/matchResultsTables/doubleElim.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <div key={round}>
+          <div className="mt-1.5 text-xl">Round {round}</div>
+          <MatchResultsTableBase
+            matches={matches}
+            matchTitleFormatter={(m) => `Match ${m.set_number}`}
+          />
+        </div>
+      ))}
+
+      <div className="mt-1.5 text-xl">Finals</div>
```
fangeugene, line 44: Probably want to check to see if any finals matches exist

pwa/app/routes/event.$eventKey.tsx
```diff
@@ -82,11 +84,31 @@ export default function EventPage() {
...
+  const elims = useMemo(
+    () => sortedMatches.filter((m) => m.comp_level !== 'qm'),
+    [sortedMatches],
+  );
+
+  const leftSideMatches =
+    quals.length > 0 ? (
+      <MatchResultsTableQuals matches={quals} />
+    ) : (
+      <MatchResultsTableDoubleElim matches={elims} />
```
fangeugene, line 106: Is the eventual plan to change this based on playoff type?

tervay: Yes

PR #6207, opened 2024-08-10
Add leaderboard insights to db && add insights API

tervay:
Upgrading the old insights felt like too brittle of a task, so I felt the best option was to create a couple of new insights that are guaranteed to be typed correctly. I chose `n=25` arbitrarily, this can be adjusted.

Screenshot with a limited number of local events:

image

### DIFF
--- a/src/backend/api/handlers/insights.py
+++ b/src/backend/api/handlers/insights.py
@@ -1,0 +1,30 @@
+from flask import Response
+
+from backend.api.handlers.decorators import api_authenticated
+from backend.api.handlers.helpers.profiled_jsonify import profiled_jsonify
+from backend.api.handlers.helpers.track_call import track_call_after_response
+from backend.common.consts.api_version import ApiMajorVersion
+from backend.common.decorators import cached_public
+from backend.common.models.insight import Insight
+from backend.common.queries.insight_query import InsightsByNameAndYearQuery
+
+
+@api_authenticated
+@cached_public
+def insights_leaderboards_year(year: int) -> Response:
+    track_call_after_response("insights/leaderboards", str(year))
+    futures = []
+    for insight_type in Insight.TYPED_LEADERBOARD_MATCH_INSIGHTS.union(
+        Insight.TYPED_LEADERBOARD_AWARD_INSIGHTS
+    ):
+        futures.append(
+            InsightsByNameAndYearQuery(
+                insight_name=Insight.INSIGHT_NAMES[insight_type], year=year
+            ).fetch_dict_async(ApiMajorVersion.API_V3)
+        )
+
+    insights = []
+    for future in futures:
+        insights.extend(future.get_result())
+
+    return profiled_jsonify(insights)
--- a/src/backend/common/helpers/insights_helper.py
+++ b/src/backend/common/helpers/insights_helper.py
@@ -1,8 +1,9 @@
 import itertools
 import json
 import math
+import statistics
 from collections import defaultdict
-from typing import Any, Dict, List, NamedTuple, Tuple
+from typing import Any, DefaultDict, Dict, List, NamedTuple, Tuple, TypedDict
 
 import numpy as np
 from google.appengine.ext import ndb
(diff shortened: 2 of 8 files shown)


src/backend/web/static/swagger/api_v3.json
ZachOrr, line 1: Noting here: Make sure we're updating the versions + changelog for these changes

src/backend/common/helpers/insights_helper.py
```diff
@@ -226,6 +235,58 @@ def _createInsight(self, data: Any, name: str, year: Year) -> Insight:
             data_json=json.dumps(data),
         )
 
+    @classmethod
+    def _create_leaderboard_from_dict_counts(
+        cls, dict: Dict[TeamKey, int], insight_type: int, year: int
+    ) -> Insight:
+        sorted_leaderboard_tuples = cls._sortTeamWinsDict(dict)
+        leaderboard_data = [
```
fangeugene, line 243: Can we type this (e.g. `leaderboard_data: LeaderboardData = []`) just to make sure we have a Python type for this and don't accidentally break this in the future?

src/backend/common/helpers/insights_helper.py
```diff
@@ -1088,4 +1151,48 @@ def doOverallAwardInsights(self) -> List[Insight]:
...
+                        data[team] += leaderboard_ranking["value"]
+
+            overall_insights.append(
+                cls._create_leaderboard_from_dict_counts(data, insight_type, year=0)
+            )
+
+        return overall_insights
+
+    @classmethod
+    def do_overall_leaderboard_match_insights(cls) -> List[Insight]:
```
fangeugene, line 1180: Can these two methods be combined taking in a list of insight names as an argument?

src/backend/api/handlers/insights.py
```diff
@@ -0,0 +1,54 @@
...
+from backend.common.models.insight import Insight
+from backend.common.queries.insight_query import (
+    InsightsByNameAndYearQuery,
+    InsightsByNameQuery,
+)
+
+
+@api_authenticated
+@cached_public
+def insights_leaderboards_all() -> Response:
```
fangeugene, line 17: Is the intent here to query ALL insights, or just the overall aggregate ones (`year == 0`)?

fangeugene (changes requested):
lg overall, a few minor questions/tweaks

src/backend/api/handlers/insights.py
tervay: The intent was all, but maybe that's too much data in the future?

fangeugene: What's the intended use case? Right now the individual year insights pages would presumably use `insights_leaderboards_year`, and the overall one would just use the aggregate. When do we need to fetch all of them?

tervay: I dunno. Good point.

src/backend/common/queries/insight_query.py
```diff
@@ -0,0 +1,42 @@
...
+from backend.common.models.keys import Year
+from backend.common.queries.database_query import CachedDatabaseQuery
+from backend.common.queries.dict_converters.insight_converter import (
+    InsightConverter,
+    InsightDict,
+)
+from backend.common.tasklets import typed_tasklet
+
+
+class InsightsByNameQuery(CachedDatabaseQuery[List[Insight], List[InsightDict]]):
```
fangeugene, line 13: Is this query used anywhere?

fangeugene (commented):
This is really sweet, I'm excited about this! Last question before merging.

src/backend/common/queries/insight_query.py
tervay: This is currently not used but I had it in there for maybe future use if we wanted to add endpoints for specific insights. But not super sure that's actually important right now. I'll remove and we can always put back in later.

fangeugene (approved):

PR #6213, opened 2024-08-11
Show API changelog in docs

ZachOrr:
Need assistance with this one

Show API changelog in API docs

Screenshot 2024-08-11 at 3 15 33 PM


src/frontend/apidocs/ApiDocsFrame.js
```diff
@@ -2,11 +2,60 @@ import React, { Component } from "react";
...
+          this.changes = props.info.get('x-changes');
+        }
+
+        render() {
+          const Collapse = system.getComponent("Collapse");
+
+          return (
+            <div>
+              <Original {...this.props} />
+              <h2>API Changelog <button class="swagger-ui btn" onClick={() => this.setState({ isOpened: !this.state.isOpened })}>{this.state.isOpened ? "Hide" : "Show"}</button></h2>
```
fangeugene, line 22: should be `className`

src/frontend/apidocs/ApiDocsFrame.js
```diff
@@ -2,11 +2,60 @@ import React, { Component } from "react";
...
+          );
+        }
+      }
+    }
+  }
+}
+
+class ChangesComponent extends Component {
+  render() {
+    return this.props.changes.map((changes, version) => {
```
fangeugene, line 36: This is broken because `this.props.changes` is an ImmutableJS OrderedMap, so `.map` doesn't really behave as expected.

You'll need to do this to convert the OrderedMap into a normal JS array:
```
    return this.props.changes
      .entrySeq()
      .toArray()
      .map((item) => {
        const version = item[0];
        const changes = item[1].toArray();
        return (
          <div key={version}>
            <h3>{version}</h3>
            <ul>
              {changes.map((c, i) => (
                <li key={i}>{c}</li>
              ))}
            </ul>
          </div>
        );
      });
```

Also when you render a list with `.map`, each child needs a `key` https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

fangeugene: edit: actually both of the `.toArray()` are unnecessary

src/backend/web/static/swagger/api_v3.json
```diff
@@ -2,10 +2,74 @@
...
+      "3.03.1": [
+        "Fix model of /team/{team_key}/robots to match API, add min and max for year and add min for page_num."
+      ],
+      "3.03.0": [
+        "Added Timeseries models and endpoints."
+      ],
+      "3.02.1": [
+        "Fixed the model for `/team/{team_key}/districts` to match return from API."
+      ],
+      "3.02.0": [
```
fangeugene, line 48: Why do we have a bunch of `3.0X` versions between `3.1` and `3.5?`

src/frontend/apidocs/ApiDocsFrame.js
ZachOrr: Updated this! Thanks - makes sense.

ZachOrr:
Going to close for now, since we got this in pwa

PR #6215, opened 2024-08-12
[PWA] Group events on event list page

fangeugene:


pwa/app/lib/utils.ts
```diff
@@ -11,6 +11,28 @@ export function parseDateString(date: string) {
   return new Date(date);
 }
 
+export function getEventDateString(event: Event, month: 'long' | 'short') {
```
tervay, line 14: Should this return just `${startDateString}` if the event is 1 day?

pwa/app/routes/events.$year.tsx
```diff
@@ -35,19 +36,109 @@ export async function clientLoader({ params }: ClientLoaderFunctionArgs) {
...
+      } else {
+        eventsByChampionship.set(groupName, {
+          groupName,
+          events: [event],
+        });
+      }
+    }
+
+    // FOC
+    if (event.event_type == 6) {
```
tervay, line 86: This should probably use an `EventType` ported over to TS

pwa/app/lib/utils.ts
fangeugene: It does that 
```
if (startDate == endDate) {
  return endDateString;
}
```

There reason we return the end date rather than the start date is because the end date  includes the year.

pwa/app/routes/events.$year.tsx
fangeugene: Oh yeah i hardcoded these for testing but forgot to extract them out

pwa/app/lib/utils.ts
tervay: image

Not sure what's going on here

pwa/app/routes/events.$year.tsx
tervay: on my other repo i just copy pasted the python code into chatgpt and told it what to do

pwa/app/lib/utils.ts
fangeugene: Fixed the equality check

tervay (approved):

PR #6233, opened 2024-08-13
[PWA] Dark mode

fangeugene:


pwa/package.json
```diff
@@ -34,6 +34,7 @@
     "lucide-react": "0.427.0",
     "react": "18.3.1",
     "react-dom": "18.2.0",
+    "remix-themes": "^1.5.0",
```
tervay, line 37: reinstall with `--save-exact`

PR #6246, opened 2024-08-13
[PWA] Stricter linting

fangeugene:


pwa/app/components/ui/carousel.tsx
```diff
@@ -124,8 +124,7 @@ const Carousel = React.forwardRef<
           carouselRef,
           api: api,
           opts,
-          orientation:
-            orientation || (opts?.axis === 'y' ? 'vertical' : 'horizontal'),
+          orientation,
```
fangeugene, line 127: This seems correct from a typing perspective (`orientation` seems to never be null), but it still spooks me that the original component had this extra logic @tervay

tervay:
requesting https://www.npmjs.com/package/eslint-plugin-unused-imports

tervay:
the eslint plugin ecosystem (and configs) has got to have some of the absolute worst naming schemes i've ever seen

tervay (approved):

fangeugene:
> requesting https://www.npmjs.com/package/eslint-plugin-unused-imports

Why is this necessary? don't we already check for this?

image

tervay:
> > requesting [npmjs.com/package/eslint-plugin-unused-imports](https://www.npmjs.com/package/eslint-plugin-unused-imports)
> 
> Why is this necessary? don't we already check for this?
> 
> image

oops, yes. missed this

PR #6247, opened 2024-08-13
feat(Nav): add nav component, basic icons, update styles

ehamwey:
image

## Description

Adds the nav component from prototypes and necessary shadcn components for it, updates some of the styling, and adds a bunch of icons from unplugin. 

This PR also updates the font to Roboto Flex, which I think is a little bit more refined, legible, and unique looking than Roboto, and we can use it as one of the "dangling carrots"/"new features" in the pwa. If people feel strongly about keeping Roboto as is I'm cool with that too. 

## Types of changes


pwa/package.json
```diff
@@ -21,6 +21,8 @@
     "@radix-ui/react-slot": "1.1.0",
     "@radix-ui/react-tabs": "1.1.0",
     "@radix-ui/react-tooltip": "1.1.2",
+    "@radix-ui/react-navigation-menu": "^1.2.0",
+    "@radix-ui/react-popover": "^1.1.1",
```
tervay, line 25: install as exact (`--save-exact`)

pwa/app/root.tsx
```diff
@@ -37,24 +38,25 @@ export function Layout({ children }: { children: React.ReactNode }) {
         <meta charSet="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <link rel="preconnect" href="https://fonts.googleapis.com" />
+        <link
+          href="https://fonts.googleapis.com/css2?family=Roboto%20Flex:ital,wght@0,100;0,300;0,400;0,500;0,700;0,900;1,100;1,300;1,400;1,500;1,700;1,900&display=swap"
+          rel="stylesheet"
+        />
```
tervay, line 44: IMO we should split this into a different PR to discuss if we want to or not

tervay (changes requested):

tervay:
Some git shenanigans have occurred and its probably a good idea to squash a bunch of my commits away

pwa/app/components/ui/dropdown-menu.tsx
```diff
@@ -0,0 +1,220 @@
+import * as React from 'react';
+import * as DropdownMenuPrimitive from '@radix-ui/react-dropdown-menu';
+import { Check, ChevronRight, Circle } from 'lucide-react';
+
+import { cn } from '@/lib/utils';
```
tervay, line 5: This got switched to ~ instead of @ at some point

PR #6248, opened 2024-08-13
[PWA] Test out credenza (dialog+drawer) for teams at event list

tervay:
On mobile, this is a drawer:

image

It can be dismissed by swiping.

On desktop, it's a modal:

image

The intent here is to put match schedule, etc. For now, its just a name.

### DIFF
--- a/pwa/app/routes/event.$eventKey.tsx
+++ b/pwa/app/routes/event.$eventKey.tsx
@@ -14,6 +14,8 @@ import {
   getEventAwards,
   getEventMatches,
   getEventRankings,
+  getEventTeams,
+  Team,
 } from '~/api/v3';
 import AllianceSelectionTable from '~/components/tba/allianceSelectionTable';
 import AwardRecipientLink from '~/components/tba/awardRecipientLink';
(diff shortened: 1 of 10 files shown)


pwa/app/routes/event.$eventKey.tsx
```diff
@@ -299,3 +320,53 @@ function AwardsTab({ awards }: { awards: Award[] }) {
...
+  teams.sort(sortTeamsComparator);
+
+  // Lot todo here:
+  // 1. add robot images
+  // 2. create a TeamEvent component that shows schedule, etc, which is inside CredenzaBody
+  // 3. add hover effects
+  return (
+    <div className="md:columns-2">
+      {teams.map((t) => (
+        <Credenza key={t.key}>
```
fangeugene, line 334: Would we need to do this every time we have a link to a team (e.g. for every match on an event page)? I feel like this would be more scalable if we tied the modal to the route and rely on link navigation to trigger the modals.

tervay: It depends on what you want. I personally am not really sure I'd want this modal/drawer when clicking from say rankings? But we could try it?

In any case, we would be altering `CredenzaTrigger` children, while the content stays the same. IMO splitting this out into reusable TBA-specific components can be in a future PR, but I can put it in the todo list I have there.

PR #6254, opened 2024-08-14
Add highest clean & combined score per year leaderboard

tervay:
image

yes this does create an all time highest match score for year=0, which i think is acceptable given how often i hear it as a trivia question:

image

(above data obviously small sample size of local events)

### DIFF
--- a/src/backend/common/helpers/insights_helper.py
+++ b/src/backend/common/helpers/insights_helper.py
@@ -112,6 +112,9 @@ class InsightsHelper(object):
         insights += self._calculate_leaderboard_highest_median_score_by_event(
             week_event_matches, year
         )
+        insights += self._calculate_highest_clean_and_combined_scores(
+            week_event_matches, year
+        )
 
         return insights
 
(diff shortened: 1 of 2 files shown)


src/backend/common/helpers/insights_helper.py
```diff
@@ -324,6 +327,46 @@ def _calculate_leaderboard_most_matches_played_by_team(
...
+        match_combined_scores = defaultdict(int)
+
+        for _, week_events in week_event_matches:
+            for _, matches in week_events:
+                for match in matches:
+                    if match.has_been_played:
+                        redScore = int(match.alliances[AllianceColor.RED]["score"])
+                        blueScore = int(match.alliances[AllianceColor.BLUE]["score"])
+
+                        if year >= 2016:
```
tervay, line 345: I took this logic from another function in this file, but it used `year >= 2017` -- I checked the 2016 score breakdowns and it works there too. Was there a reason it was 2017? @fangeugene

fangeugene: no clue 🤷

PR #6256, opened 2024-08-14
[PWA] Default event list year

fangeugene:


pwa/app/routes/events.($year).tsx
```diff
@@ -26,6 +27,12 @@ async function loadData(params: Params) {
     });
   }
 
+  if (events.data.length === 0) {
```
fangeugene, line 30: Maybe the API should be 404ing instead of returning zero-length arrays but oh well

pwa/app/lib/utils.ts
```diff
@@ -9,6 +11,25 @@ export function removeNonNumeric(str: string): string {
   return str.replace(/\D/g, '');
 }
 
+export async function parseParamsForYearElseDefault(
```
tervay, line 14: We don't need to do this in this PR, but we could create a provider component for "current year" and wrap the dom tree in it, and that way we can simply call a hook like `useCurrentSeason()` or the like

tervay (approved):

pwa/app/lib/utils.ts
fangeugene: I don't think that works because this is called during SSR?

PR #6272, opened 2024-08-14
[PWA] Update eslint to 8.57

tervay:
closes #6185, #6224, #6220

note that eslint-plugin-import does not support eslint9 natively, but it is possible to do with a couple extra compatibility-focused plugins: https://github.com/import-js/eslint-plugin-import/issues/2948#issuecomment-2148832701

for now just updating to the last 8.x version seems fine


pwa/.eslintrc.cjs
```diff
@@ -113,18 +122,7 @@ module.exports = {
...
-            ignore: [
-              'className',
-              'variant',
-              'size',
-              'sideOffset',
-              'shouldScaleBackground',
-            ],
-          },
-        ],
+        'react/prop-types': ['off'],
```
fangeugene, line 125: lol

PR #6285, opened 2024-08-15
[PWA] Add basic insights route

tervay:
image

not the prettiest but its a start

### DIFF
--- a/pwa/app/routes/insights.($year).tsx
+++ b/pwa/app/routes/insights.($year).tsx
@@ -1,0 +1,218 @@
+import { LoaderFunctionArgs } from '@remix-run/node';
+import {
+  ClientLoaderFunctionArgs,
+  Link,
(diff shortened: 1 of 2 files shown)


pwa/app/routes/insights.($year).tsx
```diff
@@ -0,0 +1,218 @@
...
+}
+
+export async function clientLoader({ params }: ClientLoaderFunctionArgs) {
+  return await loadData(params);
+}
+
+export const meta: MetaFunction<typeof loader> = ({ data }) => {
+  return [
+    {
+      title: `${data?.year} FIRST Robotics Events - The Blue Alliance`,
```
fangeugene, line 76: Handle the `year == 0` case?

fangeugene: Also the title is wrong

PR #6287, opened 2024-08-16
Add notables insights and API routes

tervay:
This adds "Notables" insights, which can be thought of as 'achievements' that teams may get once (maybe more?) per year. This includes like winning a division, winning HOF, winning WFA, etc.

The `context` data can be used for tooltips on the frontend. The frontend can do really whatever it wants here. We probably want to have a few "derived" leaderboards - like division winners - which can be calculated entirely on the frontend due to the small data size.

image

image


tervay:
replaced by #6712

PR #6293, opened 2024-08-16
[pwa] minor styling on insights page

tervay:
image
image

not sure if the tooltip should have all teams or just the teams that are cut off. i went with all of them in case anyone wants to copy paste a full list its a bit easier?

### DIFF
--- a/pwa/app/routes/insights.($year).tsx
+++ b/pwa/app/routes/insights.($year).tsx
@@ -26,6 +26,12 @@ import {
   TableHeader,
   TableRow,
 } from '~/components/ui/table';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipProvider,
+  TooltipTrigger,
+} from '~/components/ui/tooltip';
 import { NAME_TO_DISPLAY_NAME } from '~/lib/insightUtils';
 
 async function loadData(params: Params) {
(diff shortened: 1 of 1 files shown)


pwa/app/routes/insights.($year).tsx
```diff
@@ -195,8 +201,27 @@ function LeaderboardKeyList({
...
         </React.Fragment>
       ))}
-      {keyVals.length > cutoffSize &&
-        ` (... and ${keyVals.length - cutoffSize} more)`}
+      {keyVals.length > cutoffSize && (
+        <TooltipProvider>
+          <Tooltip>
+            <TooltipTrigger className="italic">
+              &nbsp;
+              {keyVals.length > cutoffSize &&
```
fangeugene, line 209: Is this condition necessary? isnt `keyVals.length > cutoffSize` already true since it's being checked above?

pwa/app/routes/insights.($year).tsx
```diff
@@ -195,8 +201,27 @@ function LeaderboardKeyList({
...
+                `(and ${keyVals.length - cutoffSize} others)`}
```
fangeugene, line 210: handle non-plural "others" if `len - size == 1`

PR #6294, opened 2024-08-17
[PWA] Add component OPR scatter plot with frc-colors integration

tervay:


pwa/app/routes/event.$eventKey.tsx
```diff
@@ -416,3 +428,17 @@ function TeamsTab({
...
   );
 }
+
+function InsightsTab({
+  coprs,
+  colors,
+}: {
+  coprs: EventCopRs;
+  colors: EventColors;
+}) {
```
fangeugene, line 438: This might be a good time to experiment with using react query for client-side fetching "non-critical" data (especially since this is on a different tab) so we don't block the page load waiting for colors.

PR #6295, opened 2024-08-17
[PWA] Custom express server configuration

fangeugene:
Issue: https://github.com/remix-run/remix/issues/9353

This will start becoming an issue with service workers / etc that we don't want cached indefinitely.

Used `eject-ras` [here](https://github.com/kiliman/rmx-cli#-eject-ras), with some additional changes.

Before:
image

After:
image

### DIFF
--- a/pwa/package.json
+++ b/pwa/package.json
@@ -6,12 +6,12 @@
   "scripts": {
     "build": "remix vite:build",
     "gcp-build": "",
-    "dev": "remix vite:dev",
+    "dev": "node ./server.js",
     "lint": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
     "lint:fix": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint . --fix",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
-    "start": "remix-serve ./build/server/index.js",
+    "start": "cross-env NODE_ENV=production node ./server.js",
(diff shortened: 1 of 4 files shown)


pwa/package.json
```diff
@@ -6,12 +6,12 @@
...
     "build": "remix vite:build",
     "gcp-build": "",
-    "dev": "remix vite:dev",
+    "dev": "node ./server.js",
     "lint": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint .",
     "lint:fix": "eslint --ignore-path .gitignore --cache --cache-location ./node_modules/.cache/eslint . --fix",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
-    "start": "remix-serve ./build/server/index.js",
+    "start": "cross-env NODE_ENV=production node ./server.js",
```
tervay, line 14: Now that we are calling node more directly, can we pin a specific node version in package.json engine field?

pwa/server.js
```diff
@@ -0,0 +1,57 @@
+import { createRequestHandler } from '@remix-run/express';
+import compression from 'compression';
+import express from 'express';
+import morgan from 'morgan';
```
tervay, line 4: well at least we can do our own logging now

tervay (changes requested):

pwa/package.json
fangeugene: 20.16.0 seems to be the latest on 20

tervay (approved):

PR #6297, opened 2024-08-17
Add event/key/team_media API endpoint

tervay:
This is needed for the frontend so that we can show team robot photos without making 40-80 requests.

image

### DIFF
--- a/pwa/app/api/v3.ts
+++ b/pwa/app/api/v3.ts
@@ -1,6 +1,6 @@
 /**
  * The Blue Alliance API v3
- * 3.9.1
+ * 3.9.2
  * DO NOT MODIFY - This file has been generated using oazapfts.
  * See https://www.npmjs.com/package/oazapfts
  */
@@ -827,6 +827,8 @@ export type Media = {
   };
   /** True if the media is of high quality. */
   preferred?: boolean;
+  /** List of teams that this media belongs to. Most likely length 1. */
+  team_keys: string[];
   /** Direct URL to the media. */
   direct_url?: string;
   /** The URL that leads to the full web page for the media, if one exists. */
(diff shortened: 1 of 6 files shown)


src/backend/common/queries/dict_converters/media_converter.py
```diff
@@ -40,6 +40,7 @@ def mediaConverter_v3(cls, media: Media) -> MediaDict:
             "preferred": True if media.preferred_references != [] else False,
             "view_url": None,
             "direct_url": None,
+            "team_keys": [r.id() for r in media.references],
```
fangeugene, line 43: `references` may be any key, so this should probably filter for references that are team keys.

pwa/app/api/v3.ts
```diff
@@ -827,6 +827,8 @@ export type Media = {
   };
   /** True if the media is of high quality. */
   preferred?: boolean;
+  /** List of teams that this media belongs to. Most likely length 1. */
```
fangeugene, line 830: > Most likely length 1.
😆

src/backend/api/handlers/event.py
```diff
@@ -147,6 +148,19 @@ def event_teams_statuses(event_key: EventKey) -> Response:
     return profiled_jsonify(statuses)
 
 
+@api_authenticated
+@validate_keys
+@cached_public
+def event_teams_media(event_key: EventKey) -> Response:
+    track_call_after_response("event/teams/media", event_key)
+
+    query = EventTeamsPreferredMediasQuery(event_key=event_key).fetch_dict(
```
fangeugene, line 157: this only gives the preferred media, aka the 1-2 good robot pics for the team. Is that the intent?

tervay: Yeah thats the intent.

fangeugene: Should the endpoint be renamed/documented to reflect this? Why not just make this use all the media, since `preferred` is an available property anyways?

tervay: I can make it all.

PR #6304, opened 2024-08-17
Get all awards when computing award insights

tervay:


src/backend/common/helpers/insights_helper.py
```diff
@@ -124,33 +124,12 @@ def doAwardInsights(self, year: Year) -> List[Insight]:
...
-            Award.year == year,
-            Award.award_type_enum == AwardType.FINALIST,
-            Award.event_type_enum.IN({EventType.CMP_DIVISION, EventType.CMP_FINALS}),
-        ).fetch_async(10000, keys_only=True)
-
         award_futures = ndb.get_multi_async(
-            set(blue_banner_award_keys_future.get_result()).union(
-                set(cmp_finalist_award_keys_future.get_result())
+            Award.query(
+                Award.year == year, Award.event_type_enum.IN(SEASON_EVENT_TYPES)
```
fangeugene, line 129: This probably needs to keep the event type filter so we don't get preseason / offseasons?

fangeugene: jk im blind

src/backend/common/helpers/insights_helper.py
```diff
@@ -124,33 +124,12 @@ def doAwardInsights(self, year: Year) -> List[Insight]:
...
-        cmp_finalist_award_keys_future = Award.query(
-            Award.year == year,
-            Award.award_type_enum == AwardType.FINALIST,
-            Award.event_type_enum.IN({EventType.CMP_DIVISION, EventType.CMP_FINALS}),
-        ).fetch_async(10000, keys_only=True)
-
         award_futures = ndb.get_multi_async(
-            set(blue_banner_award_keys_future.get_result()).union(
-                set(cmp_finalist_award_keys_future.get_result())
+            Award.query(
```
fangeugene, line 128: Actually on second thought we might want to do this closer to how we handle matches, where we fetch all the events and call `event.matches`.

fangeugene: I can take a look at this tonight

PR #6306, opened 2024-08-17
[PWA] Add images to event teams tab

tervay:
~Performance is pretty bad. Need to figure out improvements~

<details>

image

</details>

### DIFF
--- a/pwa/app/routes/event.$eventKey.tsx
+++ b/pwa/app/routes/event.$eventKey.tsx
@@ -24,12 +24,14 @@ import {
   Award,
   Event,
   Match,
+  Media,
   Team,
   getEvent,
   getEventAlliances,
   getEventAwards,
   getEventMatches,
   getEventRankings,
+  getEventTeamMedia,
   getEventTeams,
 } from '~/api/v3';
 import AllianceSelectionTable from '~/components/tba/allianceSelectionTable';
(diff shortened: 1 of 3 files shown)


fangeugene:
Can we do this with tanstack query on client side only

tervay:
Performance issues are scrolling related, not network related

tervay:
Performance issues resolved by loading the small image (`{}s.jpg`) rather than the full size image

pwa/app/routes/event.$eventKey.tsx
```diff
@@ -416,3 +435,33 @@ function TeamsTab({
...
+
+    if (media.type === 'instagram-image') {
+      return `https://www.thebluealliance.com/instagram_oembed/${media.foreign_key}`;
+    }
+
+    return media.direct_url;
+  }
+
+  return maybePreferred === undefined
+    ? imageUrl(pics[0])
```
fangeugene, line 465: I don't see a lot of value doing pics[0] here. I think if there aren't any preferred then we just return undefined.

pwa/app/routes/event.$eventKey.tsx
```diff
@@ -375,11 +385,20 @@ function TeamsTab({
...
-                alt=""
-                className="size-20"
-              />
+              {(() => {
+                const maybeImage = getTeamRobotPic(
+                  media.filter((m) => m.team_keys.includes(t.key)),
+                );
+
+                return maybeImage === undefined ? null : (
+                  <img
```
fangeugene, line 394: On prod this is a carousel. Do we want to replicate that?

tervay: IMO we don't need to. I don't think it adds a ton of value. We could always add the carousel to the team-event modal.

PR #6314, opened 2024-08-19
Don't count wildcards as awards

tervay:


fangeugene (dismissed):
formatting going crazy

fangeugene (approved):

PR #6345, opened 2024-08-21
[PWA] Add stats block to team pages

tervay:
I plan to add to this in the future, with things like award count, playoff record, qual record, etc (which will be expandable and hidden by default).

Also need to figure out a better way to center shorter blocks like 368s.

<details>

image

image

image

image

</details>

### DIFF
--- a/pwa/app/lib/matchUtils.ts
+++ b/pwa/app/lib/matchUtils.ts
@@ -1,4 +1,4 @@
-import { Event, Match } from '~/api/v3';
+import { Event, Match, WltRecord } from '~/api/v3';
 import { PlayoffType } from '~/lib/api/PlayoffType';
 
 const COMP_LEVEL_SORT_ORDER = {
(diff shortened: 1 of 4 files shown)


pwa/app/lib/matchUtils.ts
```diff
@@ -67,3 +67,34 @@ export function matchTitleShort(match: Match, event: Event): string {
...
+      (m.winning_alliance === 'blue' &&
+        m.alliances.red.team_keys.includes(teamKey)),
+  ).length;
+
+  const tied = matches.filter(
+    (m) =>
+      (m.alliances.blue.team_keys.includes(teamKey) ||
+        m.alliances.red.team_keys.includes(teamKey)) &&
+      m.winning_alliance === '' &&
+      m.actual_time !== null,
```
fangeugene, line 96: better to check that score is not `-1` to see if a match has been played. Not all matches have `actual_time` populated.

I'd create a helper called `hasBeenPlayed(match)` that checks if both red and blue scores are not `-1`.

pwa/app/lib/utils.ts
```diff
@@ -78,3 +78,8 @@ export function zip<T extends unknown[]>(...arrays: (T | undefined)[]): T[] {
 
   return zipped;
 }
+
+export function stringifyRecord(record: WltRecord): string {
+  const wl = `${record.wins}-${record.losses}`;
+  return record.ties > 0 ? `${wl}-${record.ties}` : wl;
```
fangeugene, line 84: I think keeping the 0 around is better for consistency.

fangeugene (changes requested):
lg, couple nits

tervay:
Minor styling changes. felt like the `Event Results` text was obvious and unnecessary

<details>
<summary>Details</summary>

image

image

image

image
</details>

tervay:
Moved socials down to the bottom, right above the separator

<details>
<summary>Details</summary>

image

image

image

image

</details>

fangeugene (approved):

PR #6350, opened 2024-08-22
[PWA] Add expandable team annual stats section

tervay:
I expect in the future to add a toggle to include/exclude offseason data on the left side. For now its just official data (awards / matches / high score).

<details>

image

image

image

image

image

</details>

### DIFF
--- a/pwa/app/lib/matchUtils.ts
+++ b/pwa/app/lib/matchUtils.ts
@@ -1,4 +1,4 @@
-import { Event, Match, WltRecord } from '~/api/v3';
+import { Event, Match, MatchAlliance, WltRecord } from '~/api/v3';
 import { PlayoffType } from '~/lib/api/PlayoffType';
 
 const COMP_LEVEL_SORT_ORDER = {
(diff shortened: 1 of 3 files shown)


pwa/app/routes/team.$teamNumber.tsx
```diff
@@ -366,21 +363,211 @@ export default function TeamPage(): JSX.Element {
...
       <dt className="text-gray-500">{label}</dt>
-      <dd className="order-first text-3xl font-semibold tracking-tight text-gray-900">
+      <dd className="order-first text-2xl font-semibold tracking-tight text-gray-900">
         {value}
       </dd>
     </div>
   );
 }
+
+// For some reason you can't use the <Stat /> component inside the Trigger with asChild
```
fangeugene, line 545: Is `asChild` necessary? I just tried wrapping `<Stat />` without `asChild` and it seems to work?

fangeugene (changes requested):

pwa/app/routes/team.$teamNumber.tsx
tervay: No idea what I did wrong originally, but yeah seems to work when I try again

tervay:
<details>
<summary>Details</summary>

image

image


</details>

pwa/app/lib/matchUtils.ts
```diff
@@ -72,33 +72,121 @@ function matchHasBeenPlayed(match: Match) {
...
+      match: Match;
+      score: number;
+      alliance: MatchAlliance;
+    }
+  | undefined {
+  if (matches.length === 0) {
+    return undefined;
+  }
+
+  const highScoreMatch = matches
```
fangeugene, line 166: What happens here if teamKey doesn't exist in matches? Does `[0]` break? maybe it makes sense to filter by teamKey first, check for length, and then map for the high score?

fangeugene (approved):

PR #6355, opened 2024-08-23
[PWA] Base match page

fangeugene:
image

### DIFF
--- a/pwa/app/lib/matchUtils.ts
+++ b/pwa/app/lib/matchUtils.ts
@@ -18,7 +18,7 @@ const COMP_LEVEL_SHORT_STRINGS = {
 };
 
 export function isValidMatchKey(key: string) {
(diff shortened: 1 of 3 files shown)


pwa/app/lib/matchUtils.ts
```diff
@@ -17,6 +17,12 @@ const COMP_LEVEL_SHORT_STRINGS = {
   qm: 'Quals',
 };
 
+export function isValidMatchKey(key: string) {
+  return /^[1-9]\d{3}[a-z]+[0-9]*\_(?:qm|ef\d{1,2}m|qf\d{1,2}m|sf\d{1,2}m|f\dm)\d+$/.test(
```
tervay, line 21: Can we add a couple sample unit test cases just to be sure, sometimes regex engines are different between languages

tervay (approved):

PR #6356, opened 2024-08-23
[PWA] Event key validator

fangeugene:


pwa/app/lib/eventUtils.ts
```diff
@@ -1,5 +1,9 @@
 import { Event } from '~/api/v3';
 
+export function isValidEventKey(key: string) {
+  return /^[1-9]\d{3}(\d{2})?[a-z]+[0-9]{0,3}$/.test(key);
```
tervay, line 4: A small count of sample test cases would be good here like in the match one

tervay (approved):

PR #6363, opened 2024-08-24
[PWA] Add footer

dracco1993:
<details><summary>Details</summary>
<p>

Large:
rjiR7sB

Medium:
zDmiTpt

Small:
c3aGZye

</p>
</details> 

The only two things I didn't implement yet, both of which didn't seem of critical importance, are:
- The correct timestamp for page generation (needs to get pulled from the backend on render?  Or from a local cache date, if data is pulled from there?)
- Facebook like/share (too lazy/do we still care about this?)

Some non-breaking QOL improvements to the dev container are also included, in the first couple of commits.


fangeugene:
> > * The correct timestamp for page generation (needs to get pulled from the backend on render?  Or from a local cache date, if data is pulled from there?)
> 
> Honestly I don't think this is super necessary. I don't think it has much value. I think it can be replaced by current git commit.
> 
> I also wonder if we can give the footer a minor visual refresh - maybe a little wider and a little shorter.

Agree git commit would be nice. But the time is also useful mostly for debugging.

dracco1993:
I can give flattening things out a go, and adding the commit.

I still think the timestamp has value, like @fangeugene mentioned, for debugging.

pwa/app/components/tba/footer.tsx
```diff
@@ -0,0 +1,107 @@
...
+    <div className="flex flex-col space-y-3 p-5 text-center">
+      <div className="justify-center">
+        <hr className="w-full" />
+      </div>
+      <div>
+        <div className="flex flex-row justify-center">
+          <BiInfoCircleFill className="mx-1 mt-1" />
+
+          <div className="mt-0">
+            Data provided by the{' '}
```
fangeugene, line 29: I don't think we actually want this on every page, only on team pages and specific event pages that are fetched from the FRC API

fangeugene (changes requested):

pwa/app/components/tba/footer.tsx
ZachOrr: In prod we do a check as well to see if it's an API-supported event - so something >= 2006

Rkortlandt:
Hey should these things be in the footer? 
20250322_00h10m41s_grim

fangeugene:
done in #8655

PR #6364, opened 2024-08-24
[PWA] Add year support to team pages

tervay:
Also fixes a missing sort on match tables that was driving me crazy

image


fangeugene (approved):
need to handle `/history` in the future as well.

PR #6424, opened 2024-09-05
Increase `Event.within_a_day` threshold to 2 days before event

lethosor:
## Motivation and Context
A few events this summer have had cache issues towards the start of the event, either from the event dates not including a setup day or from unknown cache compounding (per Eugene, could be any combination of memcache, Google edge cache, Cloudflare, and browser).

## How Has This Been Tested?
Not tested.

### DIFF
--- a/src/backend/common/helpers/tests/event_helper_test.py
+++ b/src/backend/common/helpers/tests/event_helper_test.py
@@ -487,9 +487,9 @@ def test_end_date_or_distant_future(
 @pytest.mark.parametrize(
     "current_date,expected_event_keys",
     [
-        (datetime.datetime(2019, 3, 1), [f"event_{i}" for i in range(1, 3)]),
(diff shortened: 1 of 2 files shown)


fangeugene:
I'd rather not make these changes since `Event.within_a_day` is used beyond just cache.

Maybe we can drop the default cache time to be 6 or 12 hours instead of 24 hours?

ZachOrr:
Agreed with Eugene here - we should not change the `Event.within_a_day` logic, as this has plenty of downstream affects in the codebase (like logic for when to dispatch notifications).

The response caching logic is really what needs to be updated here https://github.com/the-blue-alliance/the-blue-alliance/blob/6ccfe4766885ef915cc03f7469496cc34062553f/src/backend/web/handlers/event.py#L282-L285

Breaking the `max-age=86400` stuff shouldn't be a concern - we'll just need to update the tests accordingly for whatever new cache time we decide on.

PR #6452, opened 2024-09-13
[PWA] UI Improvements

ofekashery:
- Some UI improvements and cleanups
- Improve navigation styles
- Improve about page style & content
- Fix page title styles
- Add sticky sidebar in team page

### Before
CleanShot 2024-09-14 at 09 47 31@2x

### After
CleanShot 2024-09-14 at 09 48 30@2x


pwa/app/components/tba/nav.tsx
```diff
@@ -8,6 +8,7 @@ import BiPeopleFill from '~icons/bi/people-fill';
 import BiStarFill from '~icons/bi/star-fill';
 import BiThreeDotsVertical from '~icons/bi/three-dots-vertical';
 import IonCalendar from '~icons/ion/calendar';
+import SearchIcon from '~icons/lucide/search';
```
tervay, line 11: Sorry to be super nitpicky here; does [`icons/bi/search`](https://icon-sets.iconify.design/bi/search/) fit the UI better? I only ask because we use bootstrap icons in a lot of places (but not all, sometimes there are better alternatives)

Not a blocking comment, just curious what you think

pwa/app/tailwind.css
```diff
@@ -64,6 +64,10 @@
   * {
     @apply border-border;
   }
+  html,
+  body {
+    @apply scroll-pt-14 antialiased;
```
tervay, line 69: What does this fix? Again not a blocking comment, asking because I don't know

tervay:
Could you provide some screenshots of mobile just to verify they work? Any mobile through devtools is fine.

Once we get PR-based test deployments working this becomes less annoying

pwa/app/components/tba/nav.tsx
ofekashery: It's all good :) Yes, I tried the `bi` icons but they seem to me to be very old, thin and not compatible for the new interface. That's why I'm in favor of switching to another icons pack - just because we used Bootstrap in the original web doesn't mean the new design has to be completely same and not modern.
 
You might see the difference in #6453.

pwa/app/tailwind.css
ofekashery: Sure, the `scroll-pt-14` makes the browser know that there is a header with an offset of 56px - each scrolling to an element on the page (like via `#` anchor) will be `+56px`.

For example, scrolling to `#2024isde1` before will show a cut element:
CleanShot 2024-09-15 at 07 28 45@2x

And this is the same scrolling with the class applied:
CleanShot 2024-09-15 at 07 30 05@2x

ofekashery: The `antialiased` makes the font look better on certain displays, like in maoOS.

ofekashery:
@tervay Sure. Actually I didn't make any direct changes to the mobile navbar, I think we still need to make a big change there and implement a new menu for mobile.

CleanShot 2024-09-15 at 20 48 05@2x

ofekashery:
@tervay I'd appreciate your review :)

tervay (approved):

PR #6453, opened 2024-09-13
Improve event/team details styles

ofekashery:
Improved styling to be more clean and crisp. Added spacing and replaced icons.
### Before
CleanShot 2024-09-14 at 00 54 55@2x

### After
CleanShot 2024-09-14 at 00 55 10@2x

### DIFF
--- a/pwa/app/components/tba/detailEntity.tsx
+++ b/pwa/app/components/tba/detailEntity.tsx
@@ -1,0 +1,13 @@
+interface DetailEntityProps {
+  icon: React.ReactNode;
+  children?: React.ReactNode;
+}
+
+export default function DetailEntity({ icon, children }: DetailEntityProps) {
+  return (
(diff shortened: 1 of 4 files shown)


ofekashery:
@tervay

tervay:
Sorry for the delay, been traveling + started new job. @ofekashery 

1. I'd like to keep the bootstrap icons in place for the time being. I think we should keep them the same as prod for now, and ultimately figure out a real design direction later.
2. I think we should remove the `font-medium`, it feels like more eye strain to me personally. Maybe a lesser weight than 500?
3. The spacing changes I think are good.

ofekashery:
@tervay Thanks for the feedback!

1. I don't agree here. Our goal is to move toward something cleaner and more modern, no to copy one-to-one.
2. Got it, I've removed the font-medium.

tervay:
Open to change, we can try it

PR #6532, opened 2024-10-04
Create history API endpoint

tervay:
image

I'd like to add a total match record here but I'm not sure the best way to do that. Typically for WLT we just iterate over the matches but that seems highly dumb here. Likely best to do some sort of db count/groupby? But I don't know how.

I don't know why regenerating the api moved some types around but whatever.

### DIFF
--- a/src/backend/api/handlers/team.py
+++ b/src/backend/api/handlers/team.py
@@ -18,6 +18,7 @@ from backend.common.consts.api_version import ApiMajorVersion
 from backend.common.consts.media_tag import get_enum_from_url
 from backend.common.decorators import cached_public
 from backend.common.models.event_team import EventTeam
+from backend.common.models.history import History
 from backend.common.models.keys import EventKey, TeamKey
 from backend.common.models.team import Team
 from backend.common.queries.award_query import (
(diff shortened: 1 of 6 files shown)


src/backend/api/handlers/team.py
```diff
@@ -65,6 +66,16 @@ def team(team_key: TeamKey, model_type: Optional[ModelType] = None) -> Response:
     return profiled_jsonify(team)
 
 
+def team_history(team_key: TeamKey) -> Response:
```
tervay, line 69: ill add decorators once i'm done testing locally

src/backend/api/handlers/team.py
```diff
@@ -65,6 +66,16 @@ def team(team_key: TeamKey, model_type: Optional[ModelType] = None) -> Response:
...
+    track_call_after_response("team/history", team_key)
+
+    events = TeamEventsQuery(team_key=team_key).fetch_dict(ApiMajorVersion.API_V3)
```
fangeugene, line 72: These should be called with `fetch_dict_async` so they can be done in parallel.

src/backend/common/models/history.py
```diff
@@ -0,0 +1,9 @@
+from typing import List, TypedDict
+
+from backend.common.models.award import Award
+from backend.common.models.event import Event
+
+
+class History(TypedDict):
+    events: List[Event]
+    awards: List[Award]
```
fangeugene, line 9: is the intent to add some sort of `stats` field in the future here as well?

src/backend/api/handlers/team.py
```diff
@@ -65,6 +66,16 @@ def team(team_key: TeamKey, model_type: Optional[ModelType] = None) -> Response:
...
+    awards = TeamAwardsQuery(team_key=team_key).fetch_dict(ApiMajorVersion.API_V3)
+
+    history: History = History(events=events, awards=awards)
```
fangeugene, line 75: I don't think the explicit `: History` is necessary because the type can be deduced automatically. I think?

src/backend/common/models/history.py
tervay: Likely

fangeugene (approved):

PR #6585, opened 2024-10-15
Add apidocs with scalar

### LINKED ISSUE #6171: PWA Feat: Swagger API V3 page

GrahamSH-LLK:
## Description
Adds apidocs with Scalar. I think it's way nicer to use than Swagger, but I can easily swap it out if somebody objects. 
## Motivation and Context

Fixes #6171
## How Has This Been Tested?
Tested locally in chrome
## Types of changes


tervay:
This is cool, I hadn't seen Scalar before.

Could you post some screenshots and/or a screen recording so we can see how it looks? (PR deploys aren't set up yet)

GrahamSH-LLK:
image
image
image

tervay:
Neat.

We can give this a shot for now since PWA makes no current promises of functionality. If it works out, great, if it doesn't, we can revert.

tervay:
Make sure to fix the format & lint errors

ZachOrr:
This rules. Is this something we can get in the py3 `web` site as well - or is it not worthwhile

pwa/package.json
```diff
@@ -35,6 +35,7 @@
     "@radix-ui/react-tooltip": "1.2.7",
     "@react-router/express": "7.6.0",
     "@react-router/node": "7.6.0",
+    "@scalar/api-reference-react": "^0.7.13",
```
tervay: Can you just pin this to exactly 0.7.13?

tervay (changes requested):
Looks like there are multiple swagger specs that got committed in different paths

ZachOrr:
I've dropped a commit on this branch with some changes -

1) I'm loading the APIv3 Swagger json from py3. This can change eventually and we can serve it from the PWA - but for now, I'm loading it from py3
2) I simplified the API Swagger route. The existing layout of `apidocs.$doc.($version)` was cool and allowed for us to load any of our Swagger docs. However, I want to keep the `/apidocs/v3` endpoint (for now) since that's a known endpoint for the APIv3 docs. I do think we can create a component that re-uses the same code for all of our Swagger docs, but one step at a time. This just adds the APIv3 code.
3) I added a `XChangesPlugin` to show our changelog. The API changelog is actually pretty nice and useful, and I want to make sure we don't just drop surfacing it for our API consumers. Long-term I want to do something like https://github.com/the-blue-alliance/the-blue-alliance/pull/6213 so we're not just barfing a big string that has no formatting.

Finally, I updated the APIv3 schema and make some structural changes
1) I removed the "list" tag. "list" is a dumb tag. Lots of our endpoints return lists. The other ones return singular objects. We don't have an "object" tag for the endpoints that only return one. Anyways - no more ranting here.
2) I updated the tags to be friendlier to humans. So "event" -> "Event"
3) I wrote a quick Python script to sort the models. It's wild to me this Scalar component allows us to sort schema properties, tags, operations, but NOT models? The models not being alphabetical was driving me crazy - so I fixed this. This should probably be an upstream script to sort the models and make sure they stay sorted.

I'm also not determined that sorting the operations alphabetically is the best answer. I think we ought to sort them logically. It's just that right now - they're not really organized, so giving them SOME organization was better than none.

Screenshot 2025-12-02 at 6 11 01 PM

Screenshot 2025-12-02 at 6 33 23 PM

fangeugene:
@ZachOrr i merged in `main` and resolved a bunch of issues. Unfortunately your changes to the json file would not cleanly merge so i just reset it to `main`. I think those changes can go in a separate PR.

ZachOrr:
@fangeugene Sounds good! I think the biggest change is just sorting the models by name - which is going to be super messy to rebase against. That's an easy fix though - and IMO we might want a CI job to ensure those models stay sorted. So we can always follow up with any API spec changes.

fangeugene:
Scrolling auto-updates the url hash and it's not playing well with tanstack router. Scroll position ends up jittering constantly. Doesn't seem to be a solution...

same issue: https://github.com/scalar/scalar/issues/4292

I made a changelog display though
image

PR #6598, opened 2024-10-21
Store week number from FRC API

fangeugene:


phil-lopreiato (approved):
Seems alright to me!

We'll just have to make sure to handle the weird 2016 Week 0.5 case if/when we backfill

gregmarra:
@fangeugene should we close this one?

PR #6606, opened 2024-10-22
Implement search on PWA

tervay:
PhdxNsykHC

Since our search methodology is not really firmly decided yet, this introduces a straightforward generic search API + search component that can be easily modified.

Search consists of two things: a data provider and a data filterer:

```ts
export interface SearchDataProvider {
  provide(): Promise<SearchIndex>;
}

export interface SearchDataFilterer {
  filter(data: SearchIndex, query: string): SearchIndex;
}
```

Classes should be written to implement these interfaces; the in-use ones are set in the searchbar component:

```ts
const provider = useMemo(() => new ProdAPIProvider(), []);
const filterer = useMemo(() => new FuzzysortFilterer(), []);
```

It still needs work on prioritizing active teams, lower number teams, official events, recent events, etc.


pwa/app/components/tba/searchbar.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const inputRef = useRef<HTMLInputElement | null>(null);
+
+  // Toggle the menu when ⌘K or ctrl-k is pressed
+  useEffect(() => {
+    function listener(e: KeyboardEvent) {
+      if (e.key === 'k' && (e.ctrlKey || e.metaKey)) {
+        e.preventDefault();
+        setIsOpen((o) => !o);
+        inputRef.current?.focus();
```
ofekashery, line 59: It should be `focus` on open but `blur` on close.

ofekashery: Maybe consider use `<Popover />`, it'll handle you all the focus things like auto focus, blur, and focus trapping.

pwa/app/components/tba/searchbar.tsx
```diff
@@ -0,0 +1,135 @@
...
+    document.addEventListener('keydown', listener);
+
+    return () => {
+      document.removeEventListener('keydown', listener);
+    };
+  }, []);
+
+  return (
+    <Command className="relative" shouldFilter={false}>
+      <CommandInput
```
ofekashery, line 72: I think it should be with `h-9`. This `h-11` height is really big.

pwa/app/components/tba/searchbar.tsx
tervay: It uses https://ui.shadcn.com/docs/components/command, tried with popover but it was kinda gross. This is also gross though. Open to refactoring it in the future.

pwa/app/components/tba/searchbar.tsx
```diff
@@ -0,0 +1,157 @@
...
+        }
+      }
+    }
+
+    document.addEventListener('keydown', listener);
+
+    return () => {
+      document.removeEventListener('keydown', listener);
+    };
+  }, []);
```
fangeugene, line 74: Follow linter

fangeugene (changes requested):

fangeugene (approved):

PR #6683, opened 2024-11-12
Add search_index API route

tervay:
image

This is used to get search results on the frontend. This will (should) be heavily cached and only really changes between September and January.


src/backend/web/static/swagger/api_v3.json
```diff
@@ -4318,6 +4318,61 @@
           }
         ]
       }
+    },
+    "/search_index": {
```
ZachOrr, line 4322: What's the reason for not using the typeaheads the same way we use them on web?

tervay: I had talked about this approach with @fangeugene last week offline and we figured it'd be the best approach.

I am totally fine with using the typeaheads - we'd have to add TypeaheadQuery and other infra to expose it to the API. But this (as in, this PR) approach with utilizing existing Team/Event queries seems quite a bit simpler?

I'm not set on any particular method, this one seems like less code to maintain in the long run I guess? 🤷

ZachOrr: I have no preference one way or another - just trying to understand the decision!

src/backend/api/handlers/helpers/model_properties.py
```diff
@@ -44,6 +44,16 @@
...
 ]
 
+search_team_properties = [
+    "key",
+    "nickname",
+]
+
+search_event_properties = [
+    "key",
+    "name",
```
fangeugene, line 54: on prod, we use short name. We have two options:
1) Include both name and short name and choose on the client side which to use (don't love this idea because it bloats this model)
2) Make `short_name` not nullable in the API (defaulting to `name` if it's null) and use that here.

src/backend/api/handlers/search.py
```diff
@@ -0,0 +1,53 @@
...
+from backend.common.consts.api_version import ApiMajorVersion
+from backend.common.decorators import cached_public
+from backend.common.helpers.season_helper import SeasonHelper
+from backend.common.models.team import Team
+from backend.common.queries.event_query import EventListQuery
+from backend.common.queries.team_query import TeamListQuery
+
+
+@api_authenticated
+@cached_public
```
fangeugene, line 20: Add a TODO to bump the cache time on this up to a day. It's fine to keep this low while we are testing.

src/backend/api/handlers/search.py
```diff
@@ -0,0 +1,53 @@
...
+        partial_team_list = future.get_result()
+        team_list += partial_team_list
+
+    event_list = []
+    for future in event_futures:
+        partial_event_list = future.get_result()
+        event_list += partial_event_list
+
+    event_list = filter_event_properties(event_list, ModelType("search"))
+    team_list = filter_team_properties(team_list, ModelType("search"))
```
fangeugene, line 51: we might want to split events and teams into two endpoints if the cache for the combination doesn't fit in cache. but it's hard to test this locally without the full data we can test in prod 🙃

fangeugene (commented):
aside from the short name issue, i think this is good to go. We'll need to test this on prod and there's a small chance we might need to revert to precalculating this (like we do for prod typeahead) but it's hard to well without the full set of data.

src/backend/api/handlers/helpers/model_properties.py
tervay: Are you sure prod uses short_name? Here's 2024 Chezy (abbreviated):

```json
{
  "address": "960 W Hedding St, San Jose, CA 95126, USA",
  "city": "San Jose",
  "country": "USA",
  "district": null,
  "division_keys": [],
  "end_date": "2024-09-29",
  "event_code": "cc",
  "event_type": 99,
  "event_type_string": "Offseason",
  "first_event_code": null,
  "first_event_id": null,
  "key": "2024cc",
  "location_name": "Bellarmine College Preparatory",
  "name": "Chezy Champs",
  "short_name": "",
}
```

fangeugene: Prod uses short name but defaults to name. Look at any of the divisions "sponsored by X"

PR #6707, opened 2024-11-16
[pwa] fix quals being included in round1 of match tables

tervay:
image


pwa/app/components/tba/matchResultsTable.tsx
```diff
@@ -163,6 +170,10 @@ export default function MatchResultsTable(props: MatchResultsTableProps) {
 
   return (
     <>
+      <>
+        <div className="mt-1.5 text-lg">Quals</div>
+        <MatchResultsTableGroup {...props} matches={quals} />
```
fangeugene, line 175: I think this breaks in cases without quals matches (e.g. einstein)

fangeugene (approved):

PR #6712, opened 2024-11-16
Add notable insights and corresponding api route

tervay:
This adds "Notables" insights, which can be thought of as 'achievements' that teams may get once (maybe more?) per year. This includes like winning a division, winning HOF, winning WFA, etc.

The `context` data can be used for tooltips on the frontend. The frontend can do really whatever it wants here. We probably want to have a few "derived" leaderboards - like division winners - which can be calculated entirely on the frontend due to the small data size.

<details>

image

image
</details>


src/backend/common/helpers/insights_helper.py
```diff
@@ -80,7 +102,9 @@ def doMatchInsights(self, year: Year) -> List[Insight]:
             Event.query(Event.year == year).order(Event.start_date).fetch(1000)
         )
         events_by_week = EventHelper.group_by_week(official_events)
-        week_event_matches = []  # Tuples of: (week, events) where events are tuples of (event, matches)
+        week_event_matches = (
```
tervay, line 105: No I don't know why this keeps happening

src/backend/web/static/swagger/api_v3.json
```diff
@@ -8322,6 +8383,55 @@
...
+              "type": "object",
+              "properties": {
+                "entries": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "context": {
+                        "type": "array",
+                        "description": "A list of events this team achieved the notable at. This type may change over time.",
```
fangeugene, line 8401: What do you mean by "change over time"? How do you see this changing?

tervay: It may add more context - for example for WFFAs, we may want to add awardee. Not sure. It's more of a disclaimer than an intent to change

fangeugene (approved):

PR #6801, opened 2024-12-09
Add /api/v3/district/<abbr>/history endpoint

tervay:
Currently the only way to get history for a district is to hit `/api/v3/districts/<year>` for each year 2009 to `current_year` which is pretty annoying. This endpoint can help reduce client wait times.

<details>

image

image

</details>

This can be used to implement districts in `/local/bootstrap` once it lands:

1. Check if the `/local/bootstrap` input key is solely alphabetical characters (eg `ne`).
2. If so, assume its a district (since it won't be of form `frc1234` or `2024abcd`).
3. Get district history

### DIFF
--- a/src/backend/api/handlers/district.py
+++ b/src/backend/api/handlers/district.py
@@ -12,10 +12,29 @@ from backend.api.handlers.helpers.profiled_jsonify import profiled_jsonify
 from backend.api.handlers.helpers.track_call import track_call_after_response
 from backend.common.consts.api_version import ApiMajorVersion
 from backend.common.decorators import cached_public
-from backend.common.models.keys import DistrictKey
-from backend.common.queries.district_query import DistrictQuery, DistrictsInYearQuery
+from backend.common.models.keys import DistrictAbbreviation, DistrictKey
+from backend.common.queries.district_query import (
+    DistrictAbbreviationQuery,
+    DistrictQuery,
+    DistrictsInYearQuery,
+)
 from backend.common.queries.event_query import DistrictEventsQuery
 from backend.common.queries.team_query import DistrictTeamsQuery
+
+
+@api_authenticated
+@cached_public
+def district_history(district_abbreviation: DistrictAbbreviation) -> Response:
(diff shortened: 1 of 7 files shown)


src/backend/api/handlers/district.py
```diff
@@ -12,12 +12,32 @@
...
+    """
+    Returns a list of District objects with the given district abbreviation. Accounts for abbreviation changes.
+    """
+    track_call_after_response("district", district_abbreviation)
+
+    districts = DistrictAbbreviationQuery(
+        abbreviation=district_abbreviation
+    ).fetch_dict(ApiMajorVersion.API_V3)
+
+    districts.sort(key=lambda x: x["year"])
```
fangeugene, line 37: Is there a reason not to do the sort in the query?

src/backend/api/main.py
```diff
@@ -290,21 +294,27 @@ class EventDetailTypeConverter(BaseConverter):
...
-    "/event/<string:event_key>/awards/update",
-    methods=["POST"],
-    view_func=update_event_awards,
-),
-trusted_api.add_url_rule(
-    "/event/<string:event_key>/info/update",
-    methods=["POST"],
-    view_func=update_event_info,
-),
+(
```
fangeugene, line 297: why the extra nested parens?

src/backend/common/models/district.py
```diff
@@ -1,5 +1,5 @@
 import re
-from typing import List, Set
+from typing import Dict, List, Set
```
fangeugene, line 2: Remove

fangeugene (changes requested):
Few nits

src/backend/api/main.py
tervay: i have no idea how this occurred

tervay: oh, there are trailing commas on the method calls, and vscode is formatting them as a tuple

fangeugene (approved):

PR #6841, opened 2024-12-15
Add district/key/awards endpoint

tervay:
image

Currently the only way to get these is making a per-event request which is inefficient, especially at the size of something like Michigan.

Also renames getDistrict to getDistrictHistory, and clarifies that district rankings are nullable. (Example: `2025ne` rankings before 2025 has begun: https://www.thebluealliance.com/api/v3/district/2025ne/rankings)

### DIFF
--- a/src/backend/web/static/swagger/api_v3.json
+++ b/src/backend/web/static/swagger/api_v3.json
@@ -3945,7 +3945,7 @@
           "event",
           "list"
         ],
-        "description": "Gets a list of events in the given district.",
+        "description": "Gets a list of awards in the given district.",
         "operationId": "getDistrictAwards",
         "parameters": [
           {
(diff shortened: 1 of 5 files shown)


src/backend/web/static/swagger/api_v3.json
```diff
@@ -3938,6 +3938,68 @@
...
       }
     },
+    "/district/{district_key}/awards": {
+      "get": {
+        "tags": [
+          "district",
+          "event",
+          "list"
+        ],
+        "description": "Gets a list of events in the given district.",
```
fangeugene, line 3948:
```suggestion
        "description": "Gets a list of awards in the given district.",
```

fangeugene (changes requested):
lg minus the doc typo

fangeugene (approved):

PR #6844, opened 2024-12-16
Overwrite event_type/award_type types in generated api

tervay:
One problem we have is that we have to cast `event_type as EventType` in PWA, because eslint gets upset that the enum doesn't completely cover all `number` values (even though we know it should from the API).

Unfortunately every method of named, numeric enums in OpenAPI is janky and does not create a great developer experience.

This introduces a bash script that should be run to regenerate the API which:

1. Runs `oazapfts`
2. Replaces `award_type: number` with `award_type: AwardType`
3. Replaces `event_type: number` with `event_type: EventType`
4. Adds imports to the top of the file
5. Formats the file

<details>
<summary>Alternative solution</summary>

We can add this to the schema:

```json
{
// ...
"schemas": {
  "EventType": {
    "type": "integer",
    "description": "Enumeration of event types.",
    "enum": [
      0,
      1,
      2,
      3,
      4,
      5,
      6,
      7,
      99,
      100,
      -1
    ],
    "x-enum-descriptions": {
      "0": "REGIONAL",
      "1": "DISTRICT",
      "2": "DISTRICT_CMP",
      "3": "CMP_DIVISION",
      "4": "CMP_FINALS",
      "5": "DISTRICT_CMP_DIVISION",
      "6": "FOC",
      "7": "REMOTE",
      "99": "OFFSEASON",
      "100": "PRESEASON",
      "-1": "UNLABELED"
    }
  }
}
```

We can then enable `--useEnumType` in `oazapfts` which will generate:

```ts
export enum EventType {
  Regional = 0;
  District = 1;
  DistrictCmp = 2;
  CmpDivision = 3;
  // ....
}
```

Note that this also makes enums for everything else in the api spec, which are very common in score breakdowns, and things get a bit bloated. It *is* a viable solution, though.
</details>

<details>
<summary>Alternative 2</summary>

Just disable [the eslint rule](https://typescript-eslint.io/rules/no-unsafe-enum-comparison) but it wasn't working and I hate eslint

</details>

### DIFF
--- a/pwa/generate-api.sh
+++ b/pwa/generate-api.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+npx oazapfts ../src/backend/web/static/swagger/api_v3.json --argumentStyle=object > app/api/v3.ts
+sed -i 's/award_type: number/award_type: AwardType/g' app/api/v3.ts
+sed -i 's/event_type: number/event_type: EventType/g' app/api/v3.ts
+sed -i '1i\
+import { AwardType } from '"'"'~/lib/api/AwardType'"'"';\
+import { EventType } from '"'"'~/lib/api/EventType'"'"';' app/api/v3.ts
+npm run format:fix
--- a/pwa/app/api/v3.ts
+++ b/pwa/app/api/v3.ts
@@ -7,6 +7,9 @@
 import * as Oazapfts from '@oazapfts/runtime';
 import * as QS from '@oazapfts/runtime/query';
 
+import { AwardType } from '~/lib/api/AwardType';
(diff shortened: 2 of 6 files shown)


pwa/generate-api.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+npm run generate-api
```
fangeugene, line 2: Should we replace `npm run generate-api` in `package.json` with this script so there's no confusion/accidentally running the wrong script?

fangeugene (approved):

PR #6845, opened 2024-12-16
[pwa] Add /event/districtAbbr/year page

### LINKED ISSUE #6163: PWA Feat: District page

tervay:
<details>

image

image

image

image

image

image

</details>

Lots of stuff we can still add here, but this is a start.

Closes #6163


pwa/app/routes/district.$abbreviation.($year).tsx
fangeugene, line 1: Is the intent for this this to replace the current `/events/<district_key>/<year>` routes on prod? It would be nice if we could keep the same url structure so existing links don't break.

pwa/app/routes/district.$abbreviation.($year).tsx
```diff
@@ -0,0 +1,555 @@
...
+        );
+
+  const actuallyActiveTeams =
+    actuallyActiveRankings === null
+      ? teams.data
+      : teams.data.filter((team) =>
+          actuallyActiveRankings.find((r) => r.team_key === team.key),
+        );
+
+  const matchesByEvent = await Promise.all(
```
fangeugene, line 94: Do we actually need this on the top-level district page? Seems like a large call and I'm worried about blocking the whole page load waiting for these to resolve (e.g. 30+ events in michigan) for a matches played leaderboard. I guess we can monitor this and move the fetching to be client-only in the future if necessary?

pwa/app/routes/district.$abbreviation.($year).tsx
```diff
@@ -0,0 +1,555 @@
...
+    rankings.status !== 200 ||
+    teams.status !== 200 ||
+    events.status !== 200 ||
+    awards.status !== 200
+  ) {
+    throw new Response(null, { status: 404 });
+  }
+
+  // The api returns a lot of teams that previously played in the district but didn't in the given year
+  const actuallyActiveRankings =
```
fangeugene, line 80: Can/should this filtering be done not in the fetching code, but in the render?

pwa/app/routes/district.$abbreviation.($year).tsx
tervay: We can move it to that URL. I do think it's less confusing since districts have more information than just events, though. Up to you

fangeugene: I think `/events` might generalize better, so we can have a `/events/regionals` for regional points.

pwa/app/routes/district.$abbreviation.($year).tsx
tervay: Nah we definitely don't need this. I'd like to just be able to just do this on the backend. I can take this out for now, I forgot I left it in there from playing around.

pwa/app/routes/district.$abbreviation.($year).tsx
tervay: IMO it should stay in the fetch because it's a workaround for a backend bug. The API shouldn't be returning the data that it does, and until we fix it, I think we can mediate it here.

fangeugene (approved):

PR #6980, opened 2025-01-20
Update swagger types for team event status objects

bherbst:
## Description

Update swagger specs:
 * Set `playoff_average` to use the `double` format
 * Set `sort_orders` to use the `float` format

## Motivation and Context

I've been working on updating the the API models for the Android app, and ran into a few unexpected types in the team event status models. I'm sure I'll find more as I keep going, but this was a good starting point.

## How Has This Been Tested?

I ran the JSON file through https://editor.swagger.io/ and it checked out, fields now display as expected.

## Types of changes

- [X] Breaking change (fix or feature that would change API specifications or require data migrations)

This could be a breaking change if others are generating API models based on the Swagger specs, but the newly specific types align with the underlying Python types. The actual behavior of these APIs is also not changing.

### DIFF
--- a/src/backend/web/static/swagger/api_v3.json
+++ b/src/backend/web/static/swagger/api_v3.json
@@ -5114,7 +5114,8 @@
                 "type": "array",
                 "description": "Ordered list of values used to determine the rank. See the `sort_order_info` property for the name of each value.",
                 "items": {
-                  "type": "number"
+                  "type": "number",
+                  "format": "float"
                 }
               },
               "record": {
@@ -5248,7 +5249,8 @@
             ]
           },
           "playoff_average": {
-            "type": "integer",
+            "type": "number",
+            "format": "double",
             "nullable": true,
             "description": "The average match score during playoffs. Year specific. May be null if not relevant for a given year."
           }
@@ -5300,7 +5302,8 @@
                   "nullable": true,
                   "description": "Additional year-specific information, may be null. See parent `sort_order_info` for details.",
                   "items": {
-                    "type": "number"
+                    "type": "number",
+                    "format": "float"
                   }
                 },
                 "record": {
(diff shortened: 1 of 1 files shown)


tervay:
Can you regenerate the PWA API?

```bash
cd pwa/
./generate-api.sh
```

bherbst:
> Can you regenerate the PWA API?
> 
> ```shell
> cd pwa/
> ./generate-api.sh
> ```
> 
> Although I don't think it will change anything, best to make sure.

Correct - didn't change anything. Well, at least nothing meaningful in `app/api/v3.ts` - it did reformat ~ 40 files, but I'm guessing that's not an intended side effect 😆

bherbst:
>Do we need a double, or can we use a float for all of these? A double seems like a lot precision (unclear how much precision we have on the backend - I can dig later)

That's a good question, and after a little digging I think my inexperience with Python is showing.

I saw the underling types ([playoff_average](https://github.com/the-blue-alliance/the-blue-alliance/blob/4662396de7ccbba28d55ea215258db519760e1bc/src/backend/common/models/alliance.py#L29), [sort_orders](https://github.com/the-blue-alliance/the-blue-alliance/blob/4662396de7ccbba28d55ea215258db519760e1bc/src/backend/common/models/event_team_status.py#L28)) were `float`, but it turns out Python's `float` is really just a `double` in most other languages 🤷‍♂️ 

So it probably _should_ be a double - though a double is going to have ~ double the precision of a float so to your point float might be more representative of what we actually have here.

ZachOrr:
Agreed - it looks like in Python these will all be doubles - so let's type them as doubles. Also agreed that floats is probably more representative. 

Going to do some poking at the API before we get this landed, but my guess is all doubles will be fine.

bherbst:
Sounds good - PR is updated

ZachOrr (approved):

ZachOrr:
Thanks for this!

PR #7012, opened 2025-01-25
Do cleanup for district keys and event keys for teams with empty responses

ZachOrr:
Two parts to this pull request -

First - if we do a `get_team_details` and we don't get a team back for a particular year our parsing is failing. This seems like an oddity with how we're doing returns from our parsers (see also https://github.com/the-blue-alliance/the-blue-alliance/pull/6916 - which that parsing seems more right). This fixes the API return from our parsers in the case of an empty array.

The second part of this PR is to clean up `DistrictTeam`/`EventTeam` for teams that do not exist. Currently the code to cleanup these dangling teams is behind a check to make sure we get a `Team` model back from the API. In the case that we stop getting a Team back from the API, existing `DistrictTeam` and `EventTeam` models will not be cleaned up.

This data deletion *feels* safe in this case since it's scoped to a given year - although we do need to clean up 2024 data, so we'll need to run a one-off job for that.

### DIFF
--- a/src/backend/tasks_io/datafeeds/datafeed_fms_api.py
+++ b/src/backend/tasks_io/datafeeds/datafeed_fms_api.py
@@ -116,7 +116,7 @@ class DatafeedFMSAPI:
         result = self._parse(api_response, FMSAPITeamDetailsParser(year))
         if result:
             models, _ = result
-            return next(iter(models), None)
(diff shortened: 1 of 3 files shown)


ZachOrr:
Will work on fixing these tests later - likely my fix is under-thought

PR #7013, opened 2025-01-25
Add district-specific insights

tervay:
Creates district-specific insights by adding a `district_abbreviation` field to the `Insight` model. The keys have a suffix of `_{district_abbreviation}` if it's present; they're untouched if no district is attached.

District abbreviation is used instead of district key because otherwise there would not be a straightforward way to do overall insights (year=0), as there is no `0fim` district key for example.

### DIFF
--- a/src/backend/common/helpers/tests/insights_helper_test.py
+++ b/src/backend/common/helpers/tests/insights_helper_test.py
@@ -8,6 +8,8 @@ from backend.common.consts.event_type import EventType
 from backend.common.consts.insight_type import InsightType
 from backend.common.helpers.insights_helper import InsightsHelper
 from backend.common.models.award import Award
+from backend.common.models.district import District
+from backend.common.models.district_team import DistrictTeam
 from backend.common.models.event import Event
 from backend.common.models.insight import Insight
 from backend.common.models.team import Team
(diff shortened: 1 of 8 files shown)


src/backend/common/helpers/tests/insights_helper_test.py
```diff
@@ -113,6 +115,28 @@ def setup(ndb_stub):
...
     ).put()
 
+    # Setup district teams
+    District(
+        id="2022ne",
+        year=2022,
+        abbreviation="ne",
+    ).put()
+    DistrictTeam(
+        team=ndb.Key(Team, "frc604"),  # CA districts!
```
fangeugene, line 125: 604 is in new england now

src/backend/common/models/insight.py
```diff
@@ -141,6 +141,8 @@ class Insight(CachedModel):
     created = ndb.DateTimeProperty(auto_now_add=True, indexed=False)
     updated = ndb.DateTimeProperty(auto_now=True, indexed=False)
 
+    district_abbreviation: Optional[str] = ndb.StringProperty(required=False)
```
fangeugene, line 144: Can this be typed as a `DistrictAbbreviation`

PR #7014, opened 2025-01-25
Add District and DistrictTeam bootstrapping for local dev

tervay:
This doesn't bootstrap district rankings because I can't figure out how. But you can kick those by visiting http://localhost:8080/tasks/math/do/district_rankings_calc/key.

### DIFF
--- a/src/backend/web/local/bootstrap.py
+++ b/src/backend/web/local/bootstrap.py
@@ -6,6 +6,9 @@ from google.appengine.ext import deferred, ndb
 
 from backend.common.manipulators.award_manipulator import AwardManipulator
 from backend.common.manipulators.district_manipulator import DistrictManipulator
+from backend.common.manipulators.district_team_manipulator import (
+    DistrictTeamManipulator,
+)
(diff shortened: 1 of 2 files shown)


src/backend/web/local/bootstrap.py
```diff
@@ -201,5 +242,12 @@ def bootstrap_key(cls, key: str, apiv3_key: str) -> Optional[str]:
             for event_key in event_keys:
                 deferred.defer(cls.update_event, event_key, apiv3_key)
             return f"/events/{key}"
+        elif key.isalpha():
```
fangeugene, line 245: Can we validate the district key name instead of allowing all letters?

tervay: We do not, but we can add one if you want.

phil-lopreiato (approved):
Adding a hardcoded list of district abbreviations looks alright to me, as long as it's not used on the prod path (since we dynamically add new ones as we receive them from FIRST). Since it's only used here during local bootstrap, this should be okay

PR #7016, opened 2025-01-27
Update api_v3 spec to OpenAPI 3.1

tervay:
This doesn't really change much, mostly just preventative maintenance. I only changed things mentioned in here  https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0, which are:

1. Changed `example:` to `examples: []`
2. Removed `nullable: true` in favor of either `types: []` (for non-ref types) or `anyOf: [$ref, null]`

The PWA spec change is only syntactic, no actual type change.

----

edit: temp draft due to git shenanigans


fangeugene (dismissed):

tervay:
closing because i broke my version control too badly, will revisit soon

PR #7072, opened 2025-02-06
[pwa] Remove vite-pwa

tervay:
It's blocking package upgrades and it doesn't appear to be maintained anymore: https://github.com/vite-pwa/remix/issues/15


pwa/package.json
```diff
@@ -75,8 +75,6 @@
     "@types/react-dom": "18.3.1",
     "@typescript-eslint/eslint-plugin": "8.23.0",
     "@typescript-eslint/parser": "8.23.0",
-    "@vite-pwa/assets-generator": "0.2.6",
```
fangeugene, line 78: Do we need to remove the assets generator? if so we should also remove the `generate-pwa-assets` script above.

fangeugene (approved):

PR #7118, opened 2025-02-17
DO NOT MERGE finding out what happens if we upgrade protobuf

ZachOrr:


ZachOrr:
Closing this and going to report my findings -

When unpinning our protobuf<4 dependency in both TBA and our app-engine fork, all our tests pass. Again - this could be something with hitting upstream services, but using the simulators and whatnot locally I can still access ndb, use deferred tasks, etc.

We do still need our deferred patch in our fork - the current state of this PR shows that if we change to using the upstream repo (the only difference being our has our deferred compat patch) then our tests do start failing.

PR #7119, opened 2025-02-17
Add 2025 coprs

tervay:
image
image


ZachOrr (dismissed):

ZachOrr:
Actually before we merge - just a question about the component OPRs. Is the count here the interesting metric, or is it points? Or does it not matter because you can use count to get to some point OPR per level?

tervay:
Done (added total coral points - total algae points already exists from the score breakdown)

PR #7144, opened 2025-02-20
IGNORE DO NOT MERGE revert GAE pinning changes for science

ZachOrr:
Ignore this for now - was doing some science to see if I could figure out if there was the right changes in dev_appengine (or if we could patch dev_appengine somehow) as opposed to patching the deferred stuff. Trying to figure out which is less code. Putting this up to get the test suite to run.


ZachOrr:
Both failures here look like explicit failures we have in place to guard against this - which at least gives me some ideas for bench top-ing I can do against dev_appserver2

PR #7169, opened 2025-02-24
Add webcast date field to eventwizard2 UI

NixFey:
## Description

This PR adds an optional field to the event info tab of eventwizard2 to allow for specifying a date for a webcast.

## Motivation and Context

This change is to support a subset of FIRST in Michigan events, who have streams which are only active on certain days. This functionality was already available in the trusted API, but not surfaced in the event wizard.

## How Has This Been Tested?

In local environment, adding and removing webcasts via the event wizard both with and without dates, confirming changes are as expected by using the read API. Local environment is set up using the docker compose method.

Node tests and lints have been run locally and passed.

image

image

## Types of changes


phil-lopreiato (approved):
LGTM!

PR #7189, opened 2025-02-27
Refactor out match leaderboards to separate helper

tervay:
The `insights_helper` file was starting to get really unwieldy, and its only going to grow as we (I) add more insights to the site.

This refactors out the match-based leaderboards (more to come in future PRs) into a separate helper file. It also introduces a couple test data files for 2025, a shared insights util file, as well as a 2025-only leaderboard for most coral scored in a match.

A key behavioral difference here compared to the past is that it recomputes insights for *all years* when run. This is a tradeoff to use what I view as a rather nice architecture - each insight takes in a set of `[matches, teams, events]` (whatever it may be, that's what the `Arguments` class is for) and returns a single `Insight`. This `Arguments` object can be filtered and grouped beforehand - which allows us to create event-specific, district-specific, or team-specific insights much more easily (should we choose to in the future - I'd like to add district-specific ones).


src/backend/common/helpers/insights_helper_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    args_list = make_leaderboard_args()
+
+    insights: List[Insight] = []
+    for args in args_list:
+        # Only compute overall and the given year
+        if args.year not in [0, year]:
+            continue
+
+        for fn in fns:
+            if maybe_insight := fn(args):
```
ZachOrr, line 71: nit: Can we call this just `insight`?

ZachOrr (changes requested):

src/backend/tasks_cpu/handlers/insights.py
```diff
@@ -82,6 +85,46 @@
...
+)
+def enqueue_leaderboard_year_insights(kind: str, year: Year) -> Response:
+    taskqueue.add(
+        url=url_for("insights.do_leaderboard_year_insights", kind=kind, year=year),
+        method="GET",
+        target="py3-tasks-cpu",
+        queue_name="default",
+    )
+
+    return make_response(f"enqueued {kind} leaderboard insights for year {year}")
```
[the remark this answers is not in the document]
ZachOrr: Not sure if everyone can see this - just got a CodeQL thing for escaping these

Screenshot 2025-02-28 at 10 28 07 PM

src/backend/common/helpers/insights_leaderboard_match_helper.py
```diff
@@ -0,0 +1,117 @@
...
+                InsightsLeaderboardMatchHelper._highest_match_clean_score,
+                InsightsLeaderboardMatchHelper._highest_match_clean_combined_score,
+                InsightsLeaderboardMatchHelper._2025_most_coral_scored,
+            ],
+        )
+
+    @staticmethod
+    def _highest_match_clean_score(
+        arguments: LeaderboardInsightArguments,
+    ) -> Optional[Insight]:
```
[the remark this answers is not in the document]
tervay: Not this one specifically, but I did it so that it matches the function argument in `make_insights_from_functions`, so that all the insight functions are the same. Some of them may return None.

src/backend/tasks_cpu/handlers/insights.py
```diff
@@ -82,6 +87,50 @@ def do_year_insights(kind: str, year: Year) -> Response:
     return make_response("")
 
 
+@blueprint.route("/backend-tasks-b2/math/do/insights/leaderboards/<kind>/<int:year>")
```
ZachOrr, line 90: Does this need to get setup on a cron job as well?

src/backend/tasks_cpu/handlers/insights.py
```diff
@@ -82,6 +87,50 @@ def do_year_insights(kind: str, year: Year) -> Response:
...
+def enqueue_leaderboard_year_insights(kind: LeaderboardKeyType, year: Year) -> Response:
+    taskqueue.add(
+        url=url_for("insights.do_leaderboard_year_insights", kind=kind, year=year),
+        method="GET",
+        target="py3-tasks-cpu",
+        queue_name="default",
+    )
+
+    return make_response(
+        f"enqueued {escape(kind)} leaderboard insights for year {year}"
```
ZachOrr, line 117: This has the same escape CodeQL

src/backend/tasks_cpu/handlers/insights.py
tervay: Future PR

ZachOrr (approved):

PR #7245, opened 2025-03-04
[PWA] Upgrade to tailwind v4

tervay:
Verified things look the same. Verified the animations work.

- Removed `eslint-plugin-tailwindcss`, as it does not support Tailwind 4 yet (https://github.com/francoismassart/eslint-plugin-tailwindcss/pull/385)
- Removed `tailwindcss-animate`, as it does not support Tailwind 4 yet (https://github.com/jamiebuilds/tailwindcss-animate/pull/63), replaced with hand-ported version (see PR)
- Ran [automatic upgrader](https://tailwindcss.com/docs/upgrade-guide), `npx @tailwindcss/upgrade`.


fangeugene (approved):
yolo

PR #7250, opened 2025-03-05
[PWA] Improve event search results

tervay:
This improves the event search results by ordering more recent events first. A reminder that in fuzzysort, a 1.0 score is a perfect match, a 0.5 score is a good match, and a 0.0 score is no match.

Example from documentation: https://github.com/farzher/fuzzysort?tab=readme-ov-file#advanced-usage

Tested locally.

### DIFF
--- a/pwa/app/lib/search/fuzzysortFilterer.ts
+++ b/pwa/app/lib/search/fuzzysortFilterer.ts
@@ -29,6 +29,18 @@ function searchEvents(
     limit,
     keys: ['key', 'name'],
     threshold: 0.5,
+    scoreFn: (r) => {
+      // For current_year events, return score * 2
+      // For current_year-1 events, return score * (2 - 1 / (current_year - 1992))
+      // ...
+      // Down to score * 1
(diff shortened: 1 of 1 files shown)


pwa/app/lib/search/fuzzysortFilterer.ts
```diff
@@ -29,6 +29,17 @@ function searchEvents(
...
+    scoreFn: (r) => {
+      // For current_year events, return score * 2
+      // For current_year-1 events, return score * 1.95
+      // ...
+      // Down to score * 1, which will work for 20 years of events
+      const eventYear = Number.parseInt(r.obj.key.slice(0, 4));
+      const currentYear = new Date().getFullYear();
+      const yearDiff = currentYear - eventYear;
+
+      return r.score * Math.max(1, 2 - yearDiff / 20);
```
fangeugene, line 41: is the 20 going to be an issue? should we make the denominator `currentYear - 1992` instead of something hardcoded?

fangeugene (approved):

PR #7262, opened 2025-03-06
[PWA] 2025 RP dots

fangeugene:


pwa/app/components/tba/rpDot.tsx
```diff
@@ -23,10 +22,7 @@ function RpDot({
       <Tooltip>
         <TooltipTrigger asChild>
           <svg
-            className={cn('h-[4px] absolute top-[2px] left-[3px] w-[4px]', {
-              'ml-0': rpIndex === 0,
-              'ml-[6px]': rpIndex === 1,
-            })}
+            className={`h-[4px] absolute top-[2px] left-[3px] w-[4px] ml-[${6 * rpIndex}px]`}
```
tervay, line 25: This may not work consistently. The tailwind compiler checks for classes as string literals (I think), and doesn't support string interpolation. It may work due to these classes being loaded elsewhere? I'm not sure I fully I understand it, but that's why I did the object approach before.

pwa/app/lib/rankingPoints.ts
```diff
@@ -96,6 +98,15 @@ export function isScoreBreakdown2024(
   );
 }
 
+export function isScoreBreakdown2025(
+  scoreBreakdown: MatchScoreBreakdown,
+): scoreBreakdown is MatchScoreBreakdown2025Alliance {
+  return (
+    (scoreBreakdown as MatchScoreBreakdown2025Alliance).autoBonusAchieved !==
```
tervay, line 105: I would use coral or barge just in case FIRST reuses autoBonusAchieved in a future game

tervay (approved):

PR #7278, opened 2025-03-07
[PWA] Add component OPR scatter plot w/ FRC-Colors

tervay:
<details>

image

image

image

COPRs with low values (eg RP achievement rate):

image

image

</details>

The shadcn charts wrapper has a lot of linting errors by default. I've fixed a handful and disabled the rest that are not really immediately addressable.


pwa/app/routes/event.$eventKey.tsx
```diff
@@ -168,8 +182,9 @@ export default function EventPage() {
     awards,
     teams,
     teamMedia,
+    colors,
     coprs,
-  } = useLoaderData<typeof loader>();
+  } = useLoaderData<Awaited<ReturnType<typeof loadData>>>();
```
tervay, line 187: I don't know why, but after I added `colors` here, `coprs` became `unknown`. I tried everything I could think of to keep the generic type the same (`typeof loader`) but this is the only thing I could get working.

pwa/app/components/tba/charts/coprScatterChart.tsx
```diff
@@ -0,0 +1,296 @@
...
+} from '~/components/ui/select';
+import { camelCaseToHumanReadable } from '~/lib/utils';
+
+interface Datapoint {
+  teamKey: string;
+  valueX: number;
+  valueY: number;
+}
+
+function generateWholeNumberedGridlinePoints(
```
tervay, line 40: I'm not sure if these should be extracted to a utils file. I think we can do that if/when we need them again for another chart in the future, rather than this PR.

pwa/app/routes/event.$eventKey.tsx
```diff
@@ -96,17 +98,27 @@ async function loadData(params: Route.LoaderArgs['params']) {
...
+  ] = await Promise.all([
+    getEvent({ eventKey: params.eventKey }),
+    getEventMatches({ eventKey: params.eventKey }),
+    getEventAlliances({ eventKey: params.eventKey }),
+    getEventRankings({ eventKey: params.eventKey }),
+    getEventAwards({ eventKey: params.eventKey }),
+    getEventTeams({ eventKey: params.eventKey }),
+    getEventTeamMedia({ eventKey: params.eventKey }),
+    getEventCopRs({ eventKey: params.eventKey }),
+    getEventColors({ eventKey: params.eventKey }),
```
fangeugene, line 120: Doesn't have to be done in this PR, but eventually we should probably look at not blocking the whole page load on every single one of these requests (especially 3rd party ones like colors).

pwa/app/components/tba/charts/coprScatterChart.tsx
```diff
@@ -0,0 +1,296 @@
...
+}: {
+  coprs: EventCopRs;
+  colors: EventColors;
+}) {
+  const [selectedXCopr, setSelectedXCopr] = useState('teleopPoints');
+  const [selectedYCopr, setSelectedYCopr] = useState('autoPoints');
+  const [data, setData] = useState<Datapoint[]>([
+    { teamKey: '604', valueX: 2, valueY: 3 },
+    { teamKey: '2713', valueX: 4, valueY: 5 },
+  ]);
```
fangeugene, line 105: Were these left over from testing?

fangeugene (approved):

fangeugene (approved):

PR #7329, opened 2025-03-11
Refactor notable insights into separate helper

tervay:
continuation of #7248

Note that `src/backend/common/helpers/tests/insights_helper_test.py` is deleted because all of its tests have been migrated into the various insight helper test files that now exist.


src/backend/common/helpers/insights_notable_helper.py
```diff
@@ -0,0 +1,126 @@
...
+                InsightsNotableHelper._calculate_notables_world_champions,
+                InsightsNotableHelper._calculate_notables_division_winners,
+                InsightsNotableHelper._calculate_notables_division_finals_appearances,
+            ],
+        )
+
+    @staticmethod
+    def _create_notable_insight(
+        teams: Dict[TeamKey, List[EventKey]] | DefaultDict[TeamKey, List[EventKey]],
+        insight_type: int,
```
ZachOrr, line 30: Can we type this? (maybe even just like `InsightType: TypeAlias = int` in the insights type constants?

ZachOrr (changes requested):

ZachOrr (approved):

PR #7336, opened 2025-03-12
Show Pit Locations from Nexus

phil-lopreiato:
As title :smile: 

This diff implements:
 - the basic code to fetch from/interface with the [nexus api](https://frc.nexus/api)
 - plumbing that fetches pit locations: daily for all events + hourly for live events
 - rendering to show it on the event page and team detail page

image

image


src/backend/tasks_io/handlers/nexus_api.py
```diff
@@ -0,0 +1,63 @@
...
+    nexus_pit_locations = nexus_pit_locations_future.get_result()
+
+    for eventteam in eventteams:
+        team_key = eventteam.team.string_id()
+        if nexus_pit_locations and team_key in nexus_pit_locations:
+            eventteam.pit_location = nexus_pit_locations[team_key]
+        else:
+            eventteam.pit_location = None
+
+    EventTeamManipulator.createOrUpdate(eventteams)
```
fangeugene, line 57: I think there's a chance for a race condition with the existing event team stuff, but it probably doesn't matter?

src/backend/tasks_io/handlers/frc_api.py
```diff
@@ -371,15 +372,17 @@ def event_details(event_key: EventKey) -> Response:
     if not Event.validate_key_name(event_key):
         return make_response(f"Bad event key: {Markup.escape(event_key)}", 400)
 
-    df = DatafeedFMSAPI(save_response=True)
+    fms_df = DatafeedFMSAPI(save_response=True)
+    nexus_df = DatafeedNexus()
```
fangeugene, line 376: Is there a point in this given `src/backend/tasks_io/handlers/nexus_api.py` exists? alternatively only updating event teams here would avoid the race condition below.

src/backend/tasks_io/handlers/nexus_api.py
phil-lopreiato: Yeah, potentially. I don't think it'll come up in practice though, because usually team registration is pretty static by the time events are `within_a_day`. And even if this loses the race, it should get picked up on the next hourly run of this cronjob

src/backend/tasks_io/handlers/frc_api.py
phil-lopreiato: mmm, yeah, guess I probably should split this out into its out daily cronjob for all events

src/backend/web/renderers/team_renderer.py
```diff
@@ -321,6 +334,12 @@ def render_team_details(
             "max_year": SeasonHelper.get_max_year(),
             "hof": hall_of_fame,
             "team_district_points": team_district_points,
+            "has_any_pit_location": any(
+                comp["event"].official
```
fangeugene, line 338: Do we care about the official filter? I think some offseasons run nexus. (also I think `.official` has lost its real meaning, i think now it just means FMS sync is enabled)

phil-lopreiato: I think this is something we can relax in the offseason (and maybe ask @eforbes for an API to fetch all offseason events), so we don't have to optimistically try to hit every offseason looking for data.

PR #7343, opened 2025-03-12
Fix leaderboards OOMing

tervay:
Previously leaderboard calculations would store all matches `len(valid_years)` times, since for each one it was calculating the given year and also overall. Now, overall is a separate year (0) and matches are retrieved via generator given a list of events, rather than stored in `arguments`.

### DIFF
--- a/src/backend/common/helpers/insights_helper_utils.py
+++ b/src/backend/common/helpers/insights_helper_utils.py
@@ -1,7 +1,7 @@
 import json
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Any, Callable, DefaultDict, Dict, List, Optional, Tuple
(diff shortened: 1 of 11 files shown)


src/backend/common/helpers/insights_helper_utils.py
```diff
@@ -23,49 +23,34 @@
...
-        all_matches.extend(year_matches)
-        args_list.append(
-            LeaderboardInsightArguments(
-                matches=year_matches,
-                events=year_events,
-                awards=year_awards,
-                year=year,
-            )
-        )
+    year_events = [e for e in all_events if (year == 0 or e.year == year)]
```
fangeugene, line 45: isnt `all_events` already this year's events?

fangeugene (dismissed):

fangeugene (approved):

PR #7345, opened 2025-03-12
[PWA] Add react query, client-side load awards

tervay:
Adds client-side requests for data to the event page via react-query. The `shouldPreviewXYZ` fields are to minimize how often the UI "jumps" around when something is rendered. The only jumping that occurs as a result of this PR is offseason pages with awards (e.g. 2024cc), and the team count badge going from `-` to `40` (or whatever the team count is).

### DIFF
--- a/pwa/app/routes/event.$eventKey.tsx
+++ b/pwa/app/routes/event.$eventKey.tsx
@@ -1,3 +1,4 @@
+import { useQuery } from '@tanstack/react-query';
 import { ColumnDef } from '@tanstack/react-table';
 import { range } from 'lodash-es';
 import { useMemo, useState } from 'react';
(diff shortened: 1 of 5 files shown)


pwa/app/routes/event.$eventKey.tsx
```diff
@@ -393,12 +402,17 @@ export default function EventPage() {
...
-            <>
-              <CoprScatterChart colors={colors} coprs={coprs} />
-              <ComponentsTable coprs={coprs} year={event.year} />
-            </>
-          )}
+          {coprsQuery.data &&
+            colorsQuery.data?.status === 200 &&
+            Object.keys(coprsQuery.data).length > 0 && (
+              <>
+                <CoprScatterChart
```
fangeugene, line 409: should we try to render the chart with default colors if the color query fails?

fangeugene (approved):

PR #7528, opened 2025-04-07
Fix coral leaderboard to not double-count auton

tervay:
It currently says 2024oncmp2_f1m1 has 58 coral scored, but in reality it's only 48 (10 in auto, then 48 were scored at the end of tele).


src/backend/common/helpers/insights_leaderboard_match_helper.py
```diff
@@ -91,25 +91,26 @@ def _2025_most_coral_scored(
         if arguments.year != 2025:
             return None
 
+        def get_coral_count(sb: Dict) -> int:
+            return sum(
```
fangeugene, line 95: is `sum([])` better than `+`?

tervay: makes no difference, just less repetitive imo

fangeugene (approved):

PR #7529, opened 2025-04-07
Add max_team_page to Status endpoint

tervay:
Also formats the api_v3 spec to be more consistent with the rest of itself.

### DIFF
--- a/src/backend/api/handlers/status.py
+++ b/src/backend/api/handlers/status.py
@@ -5,8 +5,10 @@ from flask import Response
 from backend.api.handlers.decorators import api_authenticated
 from backend.api.handlers.helpers.profiled_jsonify import profiled_jsonify
 from backend.api.handlers.helpers.track_call import track_call_after_response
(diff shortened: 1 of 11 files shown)


src/backend/api/handlers/status.py
```diff
@@ -28,4 +29,10 @@ def status() -> Response:
     status["is_datafeed_down"] = ApiStatusFMSApiDown.get()
     status["down_events"] = down_events_list if down_events_list is not None else []
 
+    max_team_key = Team.query().order(-Team.team_number).fetch(1, keys_only=True)[0]
+    max_team_num = int(max_team_key.id()[3:])
+    max_team_page = int(max_team_num / 500)
```
phil-lopreiato, line 34: Might be worth making the page size a shared constant somewhere, and use it both here & in the API (so we can make sure they reference the same thing)

phil-lopreiato (approved):

fangeugene (approved):

Issue #4445, opened 2022-04-23
Match/Video Endpoint to do FIRST/TBA Key Translations


MC42:
from slack:

> we should make a video endpoint api that does all our internal event key/match number translations, so it's easier for first to push us videos

> Since this work I've been doing manually for Jake this year, as it's come up

phil-lopreiato:
The two that got us this weekend were the CMP event code exceptions and Einstein match numbering.

So we should just make the API just do the same logic as the match parser and save us all the trouble

bdaroz:
Reported closed by https://github.com/the-blue-alliance/the-blue-alliance/pull/7574#issuecomment-2830618204

PR #7574, opened 2025-04-17
Resolve CMP division overrides in truted API

phil-lopreiato:
This will take away an annual headache we have to do; this will allow FIRST to post videos using the full division name keys (which match frc-events) and not need to maintain their own list of overrides.


ofekashery:
Closes #4445.

PR #7585, opened 2025-04-19
Upgrade to OpenAPI 3.1

tervay:
No functional changes here. There seems to be a small bug in oazapfts that drops the description for `backup` in the new spec, but it doesn't actually affect anything.


ZachOrr:
The tool we're using to generate these typescript types just doesn't seem to support the nullable types fwiw

I'm not sure we should bend our spec to conform to lack of tooling. I'd rather we post-process in our generate script to get it to conform for the tool, or PR support in this tool

PR #7586, opened 2025-04-19
LiveEventPanel/gameday: fix issue where lingering unplayed matches could break sort order

lethosor:
## Description

2025cur still has f1m3 in Firebase (unsure why - possibly because sf3m2 happened). This is missing the comp_level/set_number/match_number fields (c/s/m in Firebase), so this match's sort order on the frontend gets calculated as `NaN`, which breaks the sorting of everything. Because matches are returned from Firebase in alphabetical order, in practice this means that semifinals end up displayed last.

This does not fix the issue of sf3m2 sorting after sf13m1 (a remnant of the old single-elim bracket) - that is a separate (maaaybe related) issue that likely warrants a bit more discussion.

## How Has This Been Tested?
Local with `2025cur`. Also confirmed no regressions on `2025dal`.

Before:

<details>

image

image

</details>

After:

<details>

image

image

</details>

## Types of changes

### DIFF
--- a/src/frontend/gameday2/selectors/index.js
+++ b/src/frontend/gameday2/selectors/index.js
@@ -89,6 +89,8 @@ export const getEventMatches = createSelector(
       // if (match.pt) {
       //   time = match.pt
       // }
+      if (!match.c || !match.m || !match.s || !compLevelsPlayOrder[match.c])
+        return 0;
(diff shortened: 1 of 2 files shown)


phil-lopreiato (approved):
Thanks!

PR #7602, opened 2025-04-23
Add district-specific insights

tervay:
Adds district-specific insights.

There are a handful of district-specific spreadsheets floating around that all basically do the same thing - a generic overview of their district's stats, all pulled from TBA and compiled. It would make sense to simply integrate these into TBA; this does that.

Because of json field size constraints, district insights are split into 2 insights: team data and district data. Team info contains team-specific info like average district points earned by a given team. District info contains info like retention, growth, event count, etc.

These are not added to cron here; they will be in a future PR once these insights are known to function on prod.


phil-lopreiato (approved):
sure!

PR #7648, opened 2025-05-04
Spin up Firebase Emulator via Docker Compose

ZachOrr:
Putting up a WIP while I finish up the rough edges here


ZachOrr:
Fixed in https://github.com/the-blue-alliance/the-blue-alliance/pull/8476

PR #7735, opened 2025-05-22
Downgrade sentry

tervay:
[@sentry/](https://github.com/getsentry/sentry-javascript/issues/16327)


.github/dependabot.yml
```diff
@@ -1,4 +1,6 @@
 version: 2
+ignore:
```
tervay, line 2: ignore until its fixed

fangeugene (approved):

PR #7737, opened 2025-05-23
Upgrade to OpenAPI 3.1 and change PWA client library

tervay:
hey-api client library has better support for OpenAPI 3.1 and better parity with the Swift client that @ZachOrr is using. It also can generate Zod validators for realtime data validation as well as react-query queries.


pwa/app/routes/event.$eventKey.tsx
```diff
@@ -111,22 +112,18 @@ async function loadData(params: Route.LoaderArgs['params']) {
...
-    getEvent({ eventKey: params.eventKey }),
-    getEventMatches({ eventKey: params.eventKey }),
-    getEventAlliances({ eventKey: params.eventKey }),
+    getEvent({ path: { event_key: params.eventKey } }),
+    getEventMatches({ path: { event_key: params.eventKey } }),
+    getEventAlliances({ path: { event_key: params.eventKey } }),
   ]);
 
-  if (event.status == 404) {
+  if (event.data == undefined) {
```
fangeugene, line 120: Is there no way to get the http response code?

tervay: There is, but it doesn't type narrow like the previous API:
```ts
if (event.response.status !== 200) {
  throw new Response(null, {
    status: 404,
  });
}

event.data; // Event | undefined
```

PR #7758, opened 2025-06-02
[PWA] Add devbox

tervay:
Adds [devbox](https://www.jetify.com/devbox) instructions and setup to PWA. This allows developers to share a reproducible shell environment.

Being evaluated as a test before being further integrated with the py3 codebase.


.github/workflows/pull_request.yml
```diff
@@ -250,6 +250,31 @@ jobs:
...
+        run: ./ops/pwa/create_dotenv.sh --env
+        env:
+          TBA_API_READ_KEY: TEST_KEY
+      - name: Install dependencies
+        run: devbox run -- npm ci
+        working-directory: pwa
+      - name: Run tests
+        run: devbox run -- npm test
+        working-directory: pwa
+      - name: Upload coverage to Codecov
```
fangeugene: Not sure what happens if we double upload to codecov. I think we can just remove this?

fangeugene (approved):

PR #7803, opened 2025-06-09
[PWA] Persist user state on login

tervay:
image


tervay:
Closing in favor of better solution

PR #7846, opened 2025-06-14
[PWA] Add authenticated loaders

tervay:
Adds authenticated SSR and the ability to add favorites on /account.

Favorites adding/list are just a temporary showcase of functionality; they'll be moved to a more correct spot in the future.

Note: @fangeugene will have to add the correct environment variables from firebase prod before merging this.


tervay:
we are no longer using react router; closing. will recreate for tanstack eventually

PR #7847, opened 2025-06-16
Add types to media.details in api spec

tervay:
Validated on all team media from 2012-2025

### DIFF
--- a/src/backend/web/static/swagger/api_v3.json
+++ b/src/backend/web/static/swagger/api_v3.json
@@ -9258,8 +9258,126 @@
           },
           "details": {
             "type": "object",
-            "additionalProperties": true,
-            "description": "If required, a JSON dict of additional media information."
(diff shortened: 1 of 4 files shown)


src/backend/web/static/swagger/api_v3.json
```diff
@@ -9258,8 +9258,126 @@
           },
           "details": {
             "type": "object",
-            "additionalProperties": true,
-            "description": "If required, a JSON dict of additional media information."
+            "description": "If required, a JSON dict of additional media information.",
+            "oneOf": [
```
fangeugene, line 9262: Do we actually have the detail types other than the one with `base64Image`?

tervay: avatars
CAD (grabcad/onshape share the same properties)
instagram (maybe old instagram?)
old CD

PR #7891, opened 2025-06-28
Move backend-tasks-b2 to py3-tasks-io

tervay:
We are encountering a lot of timeouts on backend-tasks-b2:

image

@fangeugene:

>i think the problem is that the py3-tasks-cpu service doesn't autoscale (it's pinned to one instance) so if it's handling something already, any incoming request times out.


fangeugene:
I didn't actually resolve this issue yet, the task queue thing I did was separate.

PR #7968, opened 2025-07-11
[PWA] Rewrite match table components

tervay:
Includes some experimental style changes. Previous components were difficult to extend to other things like match pages.

To do in a follow up PR:
- DQ tooltip
- Underline on team page
- Surrogate underline matching prod styling

<details><summary>
Screenshots</summary>
<p>
image
image
image

</p>
</details>


pwa/app/components/tba/match/breakers.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { timestampsAreOnDifferentDays } from '~/lib/utils';
+
+export interface BreakerResult {
+  shouldBreak: boolean;
+  text?: string;
+  whereToInsertBreak?: 'before' | 'after';
+}
+
+export type ShouldInsertBreakCallback = (
+  match: Match,
```
fangeugene: can this be done with destructuring instead of positional arguments and eliminate the problem with unused vars?

pwa/app/lib/matchUtils.ts
```diff
@@ -65,14 +68,12 @@ export function matchTitleShort(match: Match, event: Event): string {
       PlayoffType.ROUND_ROBIN_6_TEAM,
       PlayoffType.CUSTOM,
       null,
-    ].includes(event.playoff_type)
+    ].includes(playoffType)
   ) {
     return `${COMP_LEVEL_SHORT_STRINGS[match.comp_level]} ${match.set_number} Match ${match.match_number}`;
```
fangeugene: should this just be the base case instead of the `return '';` below?

fangeugene (approved):

PR #8028, opened 2025-08-04
[APIv3] Expose event team remapping

fangeugene:


src/backend/web/static/swagger/api_v3.json
```diff
@@ -5569,6 +5570,17 @@
               "null"
             ],
             "description": "String representation of the `playoff_type`, or null."
+          },
+          "remap_teams": {
+            "type": [
+              "object",
+              "null"
+            ],
+            "description": "Map of temporary \"off-season demo\" team numbers to pre-rookie and B teams. Both keys and values are team keys in the format 'frc####'. Key is the old team key, value is the new team key.",
```
lethosor: Is it worth spelling out that (only) the value can have a letter suffix?

fangeugene: Does this make more sense?

lethosor: Yep I think so

tervay (approved):

PR #8354, opened 2025-11-07
Have CI commit regenerated API changes

tervay:
Whenever hey-api gets bumped, we need to manually regen the API; this should automatically do that for us


fangeugene:
Is there a way to test if this actually works?

PR #8370, opened 2025-11-14
Add /event/<key>/html/* routes

tervay:
Adds 3 new routes:

- /event/<key>/html/teams
- /event/<key>/html/schedule
- /event/<key>/html/flat_schedule

These offer easier uses for `=IMPORTHTML()` in gSheets. I successfully tested importing them by using a local tunnel (since GSheets runs `IMPORTHTML` logic on the server and I needed to route it to a localhost).

<details><summary>Details</summary>
image

image

image

image

</details>

### DIFF
--- a/src/backend/web/handlers/event.py
+++ b/src/backend/web/handlers/event.py
@@ -3,7 +3,6 @@ import csv
 import io
 import json
 import re
-from collections import OrderedDict
 from datetime import datetime, timedelta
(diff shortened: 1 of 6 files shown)


fangeugene:
Discussed offline -- this should probably be moved to the API service. Probably no auth needed, whatever lowers the bar for accessibility.

PR #8385, opened 2025-11-18
Added download CSV + Copy to Clipboard functions to Scouting page

cadandcookies:
## Description

This adds two buttons:
A "Download CSV" button, that downloads an appropriately named CSV file from the associated textarea on the Scouting page.
A "Copy to Clipboard" button that copies the contents of the textarea to the user's clipboard. 

## Motivation and Context

These changes combined are a small UX experience improvement for users using this page for the intended purpose of copying into scouting google docs or excel sheets.

The js for this is entirely clientside. It is unclear to me if the approach I took of adding this to the template as an inline script is preferable compared to adding it to the TBA bundled javascript. Happy to revise that bit of the implementation if that is desired.

As implemented, this functionality should be functional on ~95% of browsers per MDN/caniuse. Browsers prior to 2020 may not support the clipboard functionality, browsers prior to 2017 might not support the download functionality. If we want those workarounds included, I can add those as well.

## How Has This Been Tested?

Confirmed all tests and lint locally, though I think since this is all in the templates it may not actually be being linted.
Tested on Firefox, Edge, and Chrome on Windows 11 with TBA running in Docker on WSL. 

image

## Types of changes

Disclosure: I did use GH Copilot as part of this PR, but all code was reviewed by (and almost all significantly edited by) me.

### DIFF
--- a/src/backend/web/templates/event_details.html
+++ b/src/backend/web/templates/event_details.html
@@ -468,7 +468,7 @@
           <div class="form-group">
             <div class="clearfix">
               <label class="control-label" style="display:inline-block; margin:0; padding-top:6px;">Team List</label>
-              <button type="button" class="btn btn-default btn-xs pull-right" style="margin-left:4px;" onclick="downloadCSVFromTextarea('team-list-csv', 'team_list.csv')"><span class="glyphicon glyphicon-download"></span> Download CSV</button>
+              <button type="button" class="btn btn-default btn-xs pull-right" style="margin-left:4px;" onclick="downloadCSVFromTextarea('team-list-csv', '{{ event.key_name }}_team_list.csv')"><span class="glyphicon glyphicon-download"></span> Download CSV</button>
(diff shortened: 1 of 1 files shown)


src/backend/web/templates/event_details.html
```diff
@@ -466,16 +466,28 @@ <h2 id="media">Media</h2>
           <h2>Scouting</h2>
           <p>Copy the CSV data below for scouting purposes:</p>
           <div class="form-group">
-            <label>Team List</label>
-            <textarea class="form-control" rows="20" readonly style="font-family: monospace; resize: vertical;">{{ team_list_csv|safe }}</textarea>
+            <div class="clearfix">
+              <label class="control-label" style="display:inline-block; margin:0; padding-top:6px;">Team List</label>
+              <button type="button" class="btn btn-default btn-xs pull-right" style="margin-left:4px;" onclick="downloadCSVFromTextarea('team-list-csv', 'team_list.csv')"><span class="glyphicon glyphicon-download"></span> Download CSV</button>
```
fangeugene: Do you think it would make sense to include the event key as part of the filename like `<event_key>_team_list.csv`?

cadandcookies: Oh yeah that's a great idea! I've added that!

PR #8391, opened 2025-11-22
No longer geocode `Event` models. Remove unused code for `Team` geocoding. Update `Event` API models to use FRC-API `address` and `venue`.

ZachOrr:
Less aggressive that originally suggested - this PR removes setting the event timezone from geocoded location for official events (will still work for unofficial offseasons). Updates the Event API converter to use the FRC API or inputted user location as opposed to the geocoded location. Will fallback to geocoded location in the case the Event is missing location information.

### DIFF
--- a/src/backend/common/queries/dict_converters/event_converter.py
+++ b/src/backend/common/queries/dict_converters/event_converter.py
@@ -20,7 +20,7 @@ class EventConverter(ConverterBase):
     EVENT_DATE_FORMAT_STR = "%Y-%m-%d"
 
     SUBVERSIONS = {  # Increment every time a change to the dict is made
-        ApiMajorVersion.API_V3: 7,
+        ApiMajorVersion.API_V3: 8,
     }
 
     @classmethod
--- a/src/backend/common/queries/dict_converters/team_converter.py
+++ b/src/backend/common/queries/dict_converters/team_converter.py
@@ -10,7 +10,7 @@ TeamDict = NewType("TeamDict", Dict)
 
 class TeamConverter(ConverterBase):
     SUBVERSIONS = {  # Increment every time a change to the dict is made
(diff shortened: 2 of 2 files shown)


ZachOrr:
Note to Zach: The `>= 2026` conditional is confusing and pretty silly.

What I'm really trying to protect against here is legacy events that have geocoded data but not `address` or `location_name`. For new events, we almost certainly have both.

It might be worth writing a quick script to look at the production database and seeing how many cases there are of those. We might be able to backfill from the geocoded data to the Event `address` and `location_name` properties, which would allow us to remove this year conditional. Then we can just change how it works for all events without having to worry there's some old events in the database that ONLY have geocoded location data (which is the case).

src/backend/common/queries/dict_converters/event_converter.py
```diff
@@ -79,6 +79,10 @@ def eventConverter_v3(cls, event: Event) -> EventDict:
         }
         event_dict.update(cls.constructLocation_v3(event))
 
+        # Use FRC API address + venue, fallback to geocoded location
```
fangeugene, line 82: Need to bump subversion above

src/backend/common/queries/dict_converters/team_converter.py
```diff
@@ -39,7 +40,7 @@ def teamConverter_v3(cls, team: Team) -> TeamDict:
             # "home_championship": team.championship_location,  # TODO: event not ported yet
             "school_name": team.school_name,
         }
-        team_dict.update(cls.constructLocation_v3(team))
+        team_dict.update(cls.constructLocation_v3(cast(Locatable, team)))
```
fangeugene, line 43: Need to bump subversion above

fangeugene (approved):

PR #8427, opened 2025-11-27
[pwa] Implement GameDay

fangeugene:
Will implement match ticker later.


pwa/app/root.tsx
```diff
@@ -215,14 +222,22 @@ export function Layout({ children }: { children: React.ReactNode }) {
...
       <body>
-        <Nav />
-        <div className="container mx-auto px-4 pt-14 text-sm">
-          <div vaul-drawer-wrapper="" className="bg-background">
-            <QueryClientProvider client={queryClient}>
-              {children}
-            </QueryClientProvider>
-          </div>
-        </div>
+        {isFullscreen ? (
```
tervay: Is this actually necessary?

fangeugene: I think so? we don't want to share the navbar

tervay (approved):

PR #8516, opened 2025-12-03
Update typehints for modern py3

ZachOrr:
Our typehints are pretty old. There's been a lot of improvements with [PEP 585](https://peps.python.org/pep-0585/) and [PEP 695](https://peps.python.org/pep-0695/) that make typing easier.

This PR updates our type hints to use the built in typehints that don't require imports, and a few other things.


ZachOrr:
Closing for now. I still wanna do this work - but this PR doesn't need to be open for that. Having the diff online should be enough.

PR #8517, opened 2025-12-03
[pwa] Migrate from react router to tanstack router

fangeugene:
Main advantage is the ability to do modal routing. Also seems to have better overall typesafety and debug tools.
[comparison](https://tanstack.com/router/latest/docs/framework/react/comparison)


tervay (approved):
full sends only

PR #8518, opened 2025-12-04
Add CD Thread collection page

tervay:
Very empty page for now. Will eventually add match info, award info, location, etc. Just getting this in for now.

image

### DIFF
--- a/src/backend/common/cache_clearing/get_affected_queries.py
+++ b/src/backend/common/cache_clearing/get_affected_queries.py
@@ -152,6 +152,7 @@ def media_updated(affected_refs: TAffectedReferences) -> List[TCacheKeyAndQuery]
     reference_keys = _filter(affected_refs["references"])
     years = _filter(affected_refs["year"])
     media_tags = _filter(affected_refs["media_tag_enum"])
+    media_types = _filter(affected_refs["media_type_enum"])
 
     team_keys = list(filter(lambda x: x.kind() == "Team", reference_keys))
     event_team_keys_future = (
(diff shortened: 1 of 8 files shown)


src/backend/web/handlers/team_threads.py
```diff
@@ -0,0 +1,44 @@
...
+from backend.common.queries.media_query import MediaTypeQuery
+from backend.web.profiled_render import render_template
+
+
+@cached_public
+def team_threads(year: Year) -> str:
+    if year not in SeasonHelper.get_valid_years():
+        abort(404)
+
+    all_cd_threads = MediaTypeQuery(MediaType.CD_THREAD).fetch_async().get_result()
```
fangeugene: Should the query take in `year` as an argument instead of filtering below?

src/backend/common/manipulators/manipulator_base.py
```diff
@@ -403,9 +403,23 @@ def _update_attrs(
             if not update_manual_attrs and attr in old_model.manual_attrs:
                 continue
 
-            if len(getattr(new_model, attr)) > 0 or not auto_union:
-                if getattr(new_model, attr) != getattr(old_model, attr):
-                    setattr(old_model, attr, getattr(new_model, attr))
+            attr_value = getattr(new_model, attr)
```
fangeugene: I'm not sure why these changes are necessary. `media_type_enum` isn't a `list_attr` so it should never make it here?

src/backend/common/cache_clearing/get_affected_queries.py
```diff
@@ -177,6 +178,9 @@ def media_updated(affected_refs: TAffectedReferences) -> List[TCacheKeyAndQuery]
                 queries.append(
                     media_query.TeamTagMediasQuery(reference_key.id(), media_tag)
                 )
+            for media_type in media_types:
```
fangeugene: I don't think this should be nested under the `if reference_key.kind() == "Team":` check. This should for all reference keys.

src/backend/common/manipulators/manipulator_base.py
tervay: this was an LLM hallucination that i thought i deleted, oops

fangeugene (approved):

PR #8538, opened 2025-12-05
Use `isinstance` for type checking. Update some test comparisions

ZachOrr:
Nothing crazy. Just noticed when working against pylance that `isinstance` would cast while `type() is xyz` wouldn't.

### DIFF
--- a/src/backend/api/api_trusted_parsers/json_matches_parser.py
+++ b/src/backend/api/api_trusted_parsers/json_matches_parser.py
@@ -124,7 +124,9 @@ class JSONMatchesParser:
                             f"Bad team: '{team_key}'. Must follow format 'frcXXX'."
                         )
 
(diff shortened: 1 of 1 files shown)


src/backend/api/api_trusted_parsers/json_matches_parser.py
```diff
@@ -124,7 +124,7 @@ def parse[T: (str, bytes)](matches_json: T, year: Year) -> Sequence[ParsedMatch]
                             f"Bad team: '{team_key}'. Must follow format 'frcXXX'."
                         )
 
-                if details["score"] is not None and type(details["score"]) is not int:
+                if not isinstance(details["score"], int):
```
ZachOrr: This logic actually changed - needs a fix. `None` should be valid - but with this check, it won't be.

PR #8571, opened 2025-12-09
Fix Team api spec (add motto, fix requireds)

tervay:
excessively inconsequential change but for the sake of correctness. `motto` was missing from the apispec even though it's being returned. also fixes a couple fields that were able to be undefined but actually are never undefined. tested on all teams


fangeugene (approved):

ZachOrr:
IMO we should just make sure motto stops getting returned and remove it from the API. It's been deprecated for years now - feels like we can remove it.

PR #8645, opened 2025-12-13
[pwa] Use relative paths for links

ofekashery:
Although `gameday/$eventKey` is not yet supported.


pwa/app/routes/account.tsx
```diff
@@ -74,11 +74,7 @@ function Account(): React.JSX.Element {
         <p>
           Build on top of The Blue Alliance! When logged in, you can generate
           API keys and manage webhooks. Our{' '}
-          <a
-            href="https://www.thebluealliance.com/apidocs"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
+          <a href="/apidocs" target="_blank" rel="noopener noreferrer">
```
fangeugene: use Link component?

ofekashery: Updated, not sure how different it is next with the `_blank`.

PR #8652, opened 2025-12-13
[pwa] Improve event results styles

ofekashery:
The match table is now more compact and readable. Adjusted sizing and spacing to better fit more matches on screen, and improved colors for better contrast.

I think there is still a lot of work and improvements to be made here, it's just the first step.

## Before and after

----

### DIFF
--- a/pwa/app/api/tba/read/@tanstack/react-query.gen.ts
+++ b/pwa/app/api/tba/read/@tanstack/react-query.gen.ts
@@ -368,149 +368,24 @@ const createQueryKey = <TOptions extends Options>(
   return [params];
 };
 
-export const getStatusQueryKey = (options?: Options<GetStatusData>) =>
-  createQueryKey('getStatus', options);
+export const getDistrictDcmpHistoryQueryKey = (
+  options: Options<GetDistrictDcmpHistoryData>,
+) => createQueryKey('getDistrictDcmpHistory', options);
 
 /**
- * Returns API status, and TBA status information.
+ * Gets a list of DCMP events and awards for the given district abbreviation.
  */
-export const getStatusOptions = (options?: Options<GetStatusData>) =>
-  queryOptions<
-    GetStatusResponse,
-    GetStatusError,
-    GetStatusResponse,
-    ReturnType<typeof getStatusQueryKey>
-  >({
-    queryFn: async ({ queryKey, signal }) => {
-      const { data } = await getStatus({
-        ...options,
-        ...queryKey[0],
-        signal,
-        throwOnError: true,
-      });
-      return data;
-    },
-    queryKey: getStatusQueryKey(options),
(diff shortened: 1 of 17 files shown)


pwa/app/components/tba/match/matchRows.tsx
```diff
@@ -94,110 +94,114 @@ export function MatchRow({
   const isPlayed =
     match.alliances.red.score !== -1 && match.alliances.blue.score !== -1;
 
+  /* Desktop: 1x11 grid, Mobile: 2x6 grid */
```
fangeugene, line 97: This new grid looks a lot denser on desktop but seems less dense on mobile. Is that intentional? Can mobile be made denser as well?

Image

Image

pwa/app/components/tba/match/matchRows.tsx
```diff
@@ -94,110 +94,114 @@ export function MatchRow({
...
+        className="row-span-2 flex items-center justify-center rounded-tl-lg
+          xl:col-span-1 xl:row-span-1 xl:rounded-l-lg"
       >
-        {/* Play Button */}
-        <div
-          className="row-span-2 flex items-center justify-center rounded-tl-lg
-            bg-gray-100 xl:col-span-1 xl:row-span-1 xl:rounded-l-lg"
+        {maybeVideoURL && (
+          <Link to={maybeVideoURL} className="mx-2">
+            <PlayCircleIcon />
```
fangeugene, line 116: @ofekashery do you have any suggestions on whether this link to the first match video is a good UX pattern? Personally, I don't think playing lottery with the first video is great, especially considering the match modal loads so quickly and shows all the videos. Maybe the entire row can be one big button that opens the modal? Curious to hear your thoughts.

pwa/app/components/tba/match/matchRows.tsx
ofekashery: Thanks! Fixed it now.

pwa/app/components/tba/match/matchRows.tsx
ofekashery: Maybe we can open the same match modal too, I think it's okay.

ofekashery: Or
1 video → open youtube in new tab
\>1 videos → open match modal

fangeugene: Inconsistent behavior could be confusing though

fangeugene: We can figure it out later

ofekashery: Maybe create a temporary playlist with all match records like our Watch all matches button?

fangeugene (approved):

PR #8655, opened 2025-12-13
[pwa] Add footer

ofekashery:

### DIFF
--- a/pwa/app/components/tba/navigation/footer.tsx
+++ b/pwa/app/components/tba/navigation/footer.tsx
@@ -98,7 +98,7 @@ export const Footer = () => {
 
         <div
           className="relative isolate flex justify-between border-t
-            border-gray-600/10 py-6 text-sm max-md:flex-col md:items-center"
+            border-gray-600/10 py-4 text-sm max-md:flex-col md:items-center"
         >
           <div>
             Thanks to our platinum sponsor
--- a/pwa/app/tailwind.css
+++ b/pwa/app/tailwind.css
@@ -55,7 +55,7 @@
 
   --header-height: 4rem;
   --footer-inset-top: 4rem;
-  --footer-min-height: calc(8rem + 2px); /* we have 2 borders of 1px each */
+  --footer-min-height: calc(7rem + 2px); /* we have 2 borders of 1px each */
 
   --animate-accordion-down: accordion-down 0.2s ease-out;
   --animate-accordion-up: accordion-up 0.2s ease-out;


ofekashery:
Resolves #6363, #6363.

pwa/app/routes/legal_.privacy.tsx
```diff
@@ -2,7 +2,7 @@ import { Link, createFileRoute } from '@tanstack/react-router';
 
 import { publicCacheControlHeaders } from '~/lib/utils';
 
-export const Route = createFileRoute('/privacy')({
+export const Route = createFileRoute('/legal_/privacy')({
```
fangeugene: Is there a particular reason for this? If possible we shouldn't be changing links from what's on production.

pwa/app/components/tba/navigation/footer.tsx
```diff
@@ -0,0 +1,89 @@
+import { Link } from '@tanstack/react-router';
+import { Fragment } from 'react/jsx-runtime';
+
+const links = [
```
fangeugene: Tanstack router's typesafety is broken here. e.g. if you misspell `/about` to `/abot` it doesn't show a typechecking error.

Something like this could work:

```
import type { FileRouteTypes } from '~/routeTree.gen';

interface InternalLink {
  label: string;
  to: FileRouteTypes['fullPaths'];
}

interface ExternalLink {
  label: string;
  href: string;
}

type NavigationLink = InternalLink | ExternalLink;

const links: NavigationLink[] = [...
```

pwa/app/components/tba/navigation/footer.tsx
```diff
@@ -0,0 +1,89 @@
...
+        </div>
+
+        <div
+          className="relative isolate flex justify-between border-t
+            border-gray-600/10 py-6 text-sm max-md:flex-col md:items-center"
+        >
+          <p>© {new Date().getFullYear()} The Blue Alliance</p>
+
+          <p className="text-xs text-gray-600">
+            Generated on {renderTime}. Commit:{' '}
```
fangeugene, line 118: I think "Generated on" is a bit meaningless, since the timestamp will be when the footer renders and it shouldn't re-render even if data is fetched in other components. However i think this looks good and we can keep it around, but maybe think about how to hook into the data fetching so the time is more reflective of the data. That can be a separate PR. cc @tervay

fangeugene (changes requested):
Looks awesome! just a few nits.

Also it looks a bit weird when the content is shorter than the screen height since the background goes from white to grey to white.

Image

fangeugene:
Also, anywhere we can squeeze the andymark sponsor logo in there somewhere?

pwa/app/components/tba/navigation/footer.tsx
ofekashery: Thanks, added :)

pwa/app/routes/legal_.privacy.tsx
ofekashery: I think it makes more sense this way but I'm fine with reverting.

pwa/app/components/tba/navigation/footer.tsx
ofekashery: Yes, we can rethink this later.

ofekashery:
@fangeugene just added the AM logo here:

ofekashery:
I also added a GitHub icon so that it will be much clearer for developers that TBA is open source.

PR #8656, opened 2025-12-13
[pwa] 2015 score breakdown

tervay:
image

everybody's movin here in autonomous

### DIFF
--- a/pwa/app/components/tba/match/scoreBreakdown2015.tsx
+++ b/pwa/app/components/tba/match/scoreBreakdown2015.tsx
@@ -1,0 +1,217 @@
+import { MatchScoreBreakdown2015 } from '~/api/tba/read';
+import {
+  ScoreBreakdownAllianceCell,
+  ScoreBreakdownLabelCell,
+  ScoreBreakdownRow,
+  ScoreBreakdownTable,
+} from '~/components/tba/match/scoreBreakdown';
+import {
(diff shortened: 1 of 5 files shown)


pwa/app/components/tba/match/scoreBreakdown2015.tsx
```diff
@@ -0,0 +1,217 @@
...
+        <ScoreBreakdownLabelCell shade="dark">
+          Total Teleop
+        </ScoreBreakdownLabelCell>
+        <ScoreBreakdownAllianceCell color="blue" shade="dark">
+          {scoreBreakdown.blue.teleop_points}
+        </ScoreBreakdownAllianceCell>
+      </ScoreBreakdownRow>
+
+      {/* Foul Points */}
+      <ScoreBreakdownRow
```
fangeugene, line 185: Why not use `<FoulDisplay />`?

fangeugene (commented):
Starting to backport with the best game, I see.

pwa/app/components/tba/match/scoreBreakdown2015.tsx
tervay: 2015 only has 1 type of foul and FoulDisplay is built to handle 2 types

fangeugene: Classic 2015.

PR #8660, opened 2025-12-14
[pwa] Auth + user fetching

fangeugene:
Super basic implementation of checking if a user is signed in and fetching their favorites. Got rid of the auth emulator because it won't work with prod data. it does mean `VITE_FIREBASE_API_KEY` and `VITE_FIREBASE_AUTH_DOMAIN` must be properly set in `.env` for local development, but those aren't really secrets and are exposed in JS builds anyways.


pwa/app/components/tba/auth/auth.tsx
```diff
@@ -0,0 +1,67 @@
+import {
+  type AuthProvider,
+  type User,
+  onAuthStateChanged,
+  signInWithPopup,
+  signOut,
+} from 'firebase/auth';
+import * as React from 'react';
```
tervay: i prefer to import just the functions rather than `* as React`

pwa/app/components/tba/auth/signInButton.tsx
```diff
@@ -15,13 +15,9 @@ export default function SignInButton({
...
   const handleSignIn = () => {
-    signInWithPopup(auth, provider)
-      .then((result) => {
-        console.log('User signed in:', result.user);
-      })
-      .catch((error: unknown) => {
-        console.error('Error during sign-in:', error);
-      });
+    signInWithPopup(auth, provider).catch((error: unknown) => {
+      console.error('Error during sign-in:', error);
```
tervay: use `createLogger()`

ofekashery (approved):

pwa/app/firebase/firebaseConfig.tsx
```diff
@@ -1,5 +1,6 @@
+import { isServer } from '@tanstack/react-query';
```
ofekashery, line 1: Can we add the public firebase config to default.env?

ofekashery (approved):

PR #8663, opened 2025-12-15
[pwa] Service worker precache static assets

fangeugene:


pwa/scripts/generate-sw.ts
```diff
@@ -0,0 +1,52 @@
+import { generateSW } from 'workbox-build';
+
+async function buildServiceWorker() {
+  try {
+    console.log('🔧 Generating service worker with Workbox...');
```
tervay, line 5: can we have eslint ignore the pwa/scripts/ dir for this no-console rule?

PR #8706, opened 2025-12-19
[pwa] Improve dark theme

ofekashery:
I replaced all `gray` shades with `neutral`. The light theme remains unchanged, while dark mode benefits from improved visual consistency. I also redesign the theme switcher button on footer.

## Preview

----

----

### DIFF
--- a/pwa/app/components/tba/navigation/footer.tsx
+++ b/pwa/app/components/tba/navigation/footer.tsx
@@ -58,7 +58,8 @@ function ThemeToggle() {
 
   return (
     <button
-      className="inline-flex items-center rounded-full border p-1"
+      className="group inline-flex cursor-pointer items-center rounded-full
+        border p-1"
       aria-label="Toggle Theme"
       onClick={() => setTheme(value === 'light' ? 'dark' : 'light')}
     >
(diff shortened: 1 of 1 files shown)


pwa/app/components/tba/navigation/footer.tsx
```diff
@@ -51,49 +45,37 @@ const links: NavigationLink[] = [
...
-        >
-          <Monitor className="mr-2 size-4" />
-          System
-        </DropdownMenuItem>
-      </DropdownMenuContent>
-    </DropdownMenu>
+    <button
+      className="inline-flex items-center rounded-full border p-1"
+      aria-label="Toggle Theme"
+      onClick={() => setTheme(value === 'light' ? 'dark' : 'light')}
```
fangeugene, line 64: is there a way to set this back to "system"?

ofekashery: I don’t think "system" option is very important here. Most users care about explicitly choosing light or dark, and keeping the control simple feels like a better experience.

If you want to set it back to follow the system settings, you can simply just set the theme you prefer instead. In practice, it achieves the same result.

fangeugene: fair enough.

PR #8751, opened 2026-01-07
Added navigation select to team stats page, added blue banner count

NimbleValley:
## Description

Added select dropdown for navigating between team history, stats, and specific years on the stats pages. Added default selections of Stats and History on their respective pages. Added blue banner count to team history page.

## Motivation and Context
Navigating to the team stats page leads to a dead end, as the select disappeared. The Stats page select is defaulted to Stats and the History had no default value selected, now it is History. Oftentimes users navigate to history to see their blue banners, so a count was added above the banners for quick reference.

## How Has This Been Tested?
Only web page was tested, page was rendered on mobile and desktop

image

image

## Types of changes
- [x ] New feature (non-breaking change which adds functionality)

### DIFF
--- a/pwa/app/routes/team.$teamNumber.history.tsx
+++ b/pwa/app/routes/team.$teamNumber.history.tsx
@@ -131,6 +131,7 @@ function TeamHistoryPage(): React.JSX.Element {
               params: { teamNumber: String(team.team_number), year: value },
             });
           }}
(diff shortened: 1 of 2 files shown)


tervay:
@NimbleValley Sorry for the delay - build season.

Could you format the code? and then it'll be good to go.

PR #8805, opened 2026-01-17
Add 2026 score breakdowns

fangeugene:


bdaroz:
API has match breakdown, but no Web UI

image

PR #8812, opened 2026-01-19
Fix Playwright tests to skip gracefully when API key unavailable

### LINKED ISSUE #8811: Skip optional playwright tests in CI when secrets unavailable
Playwright tests fail in CI because they don't have a `VITE_TBA_API_READ_KEY`

Example: https://github.com/the-blue-alliance/the-blue-alliance/actions/runs/21104664959/job/60693977255

Make those tests not run in CI.

# Trailhead From Gemini
To mark a Playwright test as optional or skip it specifically when secrets are missing (like in a CI environment), you can use the `test.skip()` annotation with a condition.

Since your CI environment lacks the necessary secrets, you can check for the existence of the relevant environment variable at the beginning of your test or `describe` block.

### Option 1: Skip a single test

If only one test requires the secret, add this line at the top of the test function:

```javascript
test('my test that needs secrets', async ({ page }) => {
  // Skip if the specific environment variable is missing
  test.skip(!process.env.MY_SECRET_KEY, 'Skipping because MY_SECRET_KEY is not set');
  
  // Test logic follows...
});

```

### Option 2: Skip a whole file or group

If the entire file requires secrets, you can place the skip logic inside a `test.describe` block:

```javascript
test.describe('Tests requiring secrets', () => {
  test.beforeAll(() => {
    test.skip(!process.env.MY_SECRET_KEY, 'Missing secrets; skipping these tests in CI');
  });

  test('example secret test', async ({ page }) => {
    // ...
  });
});

```

### Option 3: Detect CI specifically

If you want to be explicit about skipping only when in GitHub Actions, most CI providers set a `CI` environment variable to `true`:

```javascript
test('conditional CI skip', async ({ page }) => {
  // Runs locally, but skips in GitHub Actions
  test.skip(!!process.env.CI && !process.env.MY_SECRET_KEY, 'Skipping in CI because secrets are unavailable');
});

```

### Why this works:

* **Locally:** `process.env.MY_SECRET_KEY` will be present (assuming you have it in your `.env` or shell), so the condition is false and the test runs.
* **In CI:** The variable is undefined, the condition becomes true, and Playwright marks the test as **skipped** rather than **failed**. This allows your CI pipeline to pass while still keeping the test in your codebase for local development.

gregmarra:
## Problem:
Playwright tests were failing in CI when the VITE_TBA_API_READ_KEY secret was not configured (e.g., in forked repositories). The tests require a valid TBA API key to fetch data from routes, but the vite config enforced the key as required, and tests would fail rather than skip when credentials were unavailable.

## Solution:
1. Made VITE_TBA_API_READ_KEY optional in vite.config.ts
   - Changed from Schema.string() to Schema.string.optional()
   - Allows build to proceed without the API key

2. Created reusable test utility (pwa/tests/utils.ts)
   - Added skipWithoutApiKey() function to check for valid API key
   - Skips tests when key is missing, empty, or placeholder value
   - Provides clear skip message for CI visibility

3. Updated all Playwright test files to use skip logic
   - index.spec.ts: Home page tests (navigates to /)
   - routes.spec.ts: Comprehensive route smoke tests
   - team.spec.ts: Team page tests
   - theme.spec.ts: Theme toggle tests (navigates to /)

## Impact:
- Tests now marked as "skipped" instead of "failed" in CI
- CI pipelines pass on forks without secrets configured
- Tests still run normally when API key is available
- No changes required to GitHub Actions workflows

Fixes #8811

### DIFF
--- a/pwa/playwright.config.ts
+++ b/pwa/playwright.config.ts
@@ -44,5 +44,6 @@ export default defineConfig({
     command: 'npm run build && npm run start',
     url: 'http://localhost:3000',
     reuseExistingServer: !process.env.CI,
+    timeout: 120_000,
   },
 });


gregmarra:
Well that didn't fully work, did it.

phil-lopreiato:
I'm going to just delete this via https://github.com/the-blue-alliance/the-blue-alliance/pull/8814 until we can get the skip logic working, but we should bring it back once we do!

PR #8814, opened 2026-01-19
remove PWA E2E Test from PRs

phil-lopreiato:
This is the other way to fix https://github.com/the-blue-alliance/the-blue-alliance/issues/8811 :stuck_out_tongue: 

Since this is getting noisy, and we shouldn't be having "expected" failures on diffs.


gregmarra:
Can't fail a test that doesn't exist!

PR #8923, opened 2026-02-03
Add Tablesort, Viewing Multiple Statistics, and CSV Download to Event Insights Tab

cadandcookies:
## Description

This change adds several features to the Event "Insights" page.
- It allows users to select multiple (or all) statistics to view
- It adds tablesort functionality to the table
- It adds the ability to download the table as a csv file

When large numbers of columns are selected, the table horizontally scrolls, which isn't my favorite pattern but seemed like the least bad option.

This is a purely UI change. 

## Motivation and Context

I added these features in an effort to make the Event Insights page more useful for users.

In particular, this should make this page more useful to users who are suing TBA as an ad-hoc scouting system without using the API. 

The tablesorter integration is a bit uglier than I would have liked it to be, but this ended up being the cleanest way I could find to get it to do all the things I wanted it to. Definitely open to pointers if I missed anything obvious here.

## How Has This Been Tested?

I have tested this locally on several events on my local system (Win 11/WSL, Firefox, Chrome).

Tests/lint were run, but I don't think those actually hit this.

image

## Types of changes

Disclosure: This change was assisted by Github Copilot's Agent mode with various models. All changes have been manually reviewed by me.

### DIFF
--- a/src/backend/web/templates/event_partials/event_copr_table.html
+++ b/src/backend/web/templates/event_partials/event_copr_table.html
@@ -23,7 +23,8 @@
     <button type="button" class="btn btn-default pull-right" id="copr_csv" style="margin-left:8px;">Download CSV</button>
 </div>
 
-<table id="coprTable" class="table table-condensed table-striped table-center table-hover" style="table-layout: fixed;">
+<div style="overflow-x: auto;">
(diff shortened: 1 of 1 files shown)


tervay:
lgtm! thanks nick!

PR #8965, opened 2026-02-06
Add missing uploadDate to VideoObject structured data

gregmarra:
## Summary
- Google Search Console flagged VideoObject markup as missing the required `uploadDate` field, preventing video rich results from appearing in search
- Adds `uploadDate` to the VideoObject JSON-LD using `match.actual_time`, falling back to `match.time`, then `event.start_date`

## Test plan
- [ ] Load a match page with videos and inspect the JSON-LD `<script>` tag to verify `uploadDate` is present
- [ ] Validate with [Google Rich Results Test](https://search.google.com/test/rich-results) after deploy
- [ ] Confirm the Search Console error resolves after Google re-crawls


gregmarra:
Feels like a spiritually good time, as it is the time the thing actually happened.

PR #8978, opened 2026-02-07
Fix GSC errors for missing startDate, location, name on match pages

gregmarra:
## Summary
- Google Search Console flagged 14 match pages with invalid Event structured data: missing `startDate`, `name`, and `location`
- Root cause: the `superEvent` reference in match JSON-LD only had `@type` and `@id`, so Google treated it as a separate incomplete SportsEvent entity
- Populated `superEvent` with the event's actual data (`name`, `startDate`, `endDate`, `location`, `url`)
- Added `startDate` fallback to `event.start_date` when `match.time` is missing
- Changed match location to use `event.venue_or_venue_from_address` (matching the event page template) to cover older events with only `venue_address`

Verified fix locally using `gsm-gsc inspect` against the GSC API — the "Unnamed item" SportsEvent errors are resolved.

## Test plan
- [x] New `test_schema_org_match_sports_event` test validates both the main match SportsEvent and superEvent have all required fields
- [x] All existing match detail tests pass (12/12)
- [x] All existing event schema tests pass
- [x] Local dev server renders valid JSON-LD on match pages
- [ ] CI passes

### DIFF
--- a/src/backend/web/handlers/tests/match_detail_test.py
+++ b/src/backend/web/handlers/tests/match_detail_test.py
@@ -1,3 +1,5 @@
+import json
+
 from bs4 import BeautifulSoup
 from freezegun import freeze_time
 from werkzeug.test import Client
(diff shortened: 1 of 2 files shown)


gregmarra:
CI failures are unrelated.

PR #9020, opened 2026-02-12
Add OS_ANDROID_FCM client type for push notifications

### LINKED ISSUE #8974: Add OS_ANDROID_FCM client type to enable push notifications for new Android app
## Problem

The new Android app (Compose rewrite on `the-blue-alliance-android-2026` branch) supports FCM push notifications, but the server's `FCM_CLIENTS` set only includes `OS_IOS` and `WEB` — `OS_ANDROID` is not included. This means subscription-based notifications (match scores, upcoming matches, etc.) are never sent to Android clients.

We can't simply add `OS_ANDROID` to `FCM_CLIENTS` because old Android clients still registered with valid tokens could receive notifications and crash. The old app does strict JSON parsing in notification handlers (e.g. [`ScoreNotification.java` line 75](https://github.com/the-blue-alliance/the-blue-alliance-android/blob/3addf71de2d931a605de2fb46fac9e46541ec641/android/src/main/java/com/thebluealliance/androidclient/gcm/notifications/ScoreNotification.java#L75)) that would throw on any payload changes.

## Proposed Solution: New Client Type

Add a new `OS_ANDROID_FCM` client type so the server can distinguish new Android clients from old ones.

### Server changes

1. **`client_type.py`** — Add `OS_ANDROID_FCM = 5` to the `ClientType` enum
2. **`tbans_helper.py`** — Add `ClientType.OS_ANDROID_FCM` to the `FCM_CLIENTS` set
3. **`client_api.py`** — Map `"android-fcm"` operating system string to `OS_ANDROID_FCM` in the registration handler

### Android app change

4. **`RegisterDeviceRequestDto`** — Change default `os` from `"android"` to `"android-fcm"`

### Why this works

- Old clients register as `OS_ANDROID` (type 0) → not in `FCM_CLIENTS` → never receive subscription notifications → no crashes
- New clients register as `OS_ANDROID_FCM` (type 5) → in `FCM_CLIENTS` → receive all subscription notifications
- Old invalid GCM tokens fail with `UnregisteredError` and get auto-cleaned
- Admin UI ping still works for both since it sends directly to a token
- Fully backward compatible, zero risk to existing users

gregmarra:
## Summary
- Adds `OS_ANDROID_FCM = 5` to the `ClientType` enum for new Compose-based Android clients
- Adds it to `FCM_CLIENTS` so the server sends push notifications to these clients
- Old Android clients (type 0 / `OS_ANDROID`) are unaffected — they won't receive FCM payloads they can't handle

## Why a new type?
The new Android app supports FCM push notifications, but the old Android app would crash on the new notification payloads. We can't just add `OS_ANDROID` to `FCM_CLIENTS`. Instead, the new app registers as `"android-fcm"` → server maps to `OS_ANDROID_FCM` → included in `FCM_CLIENTS`.

## Test plan
- [x] Added `test_register_android_fcm_client` — registers with `operating_system="android-fcm"`, asserts `client_type == ClientType.OS_ANDROID_FCM`
- [x] All existing registration tests still pass (9/9)
- [x] All tbans_helper tests still pass (73/73)

Fixes #8974
Companion Android PR: the-blue-alliance/the-blue-alliance-android#1026

### DIFF
--- a/src/backend/api/handlers/tests/client_api_registration_test.py
+++ b/src/backend/api/handlers/tests/client_api_registration_test.py
@@ -46,6 +46,31 @@ def test_register_new_client(api_client: Client, mock_clientapi_auth: User) -> N
     assert client.display_name == "Test Device"
 
 
+def test_register_android_fcm_client(
+    api_client: Client, mock_clientapi_auth: User
+) -> None:
+    req = RegistrationRequest(
+        operating_system="android-fcm",
+        mobile_id="abc123",
+        device_uuid="asdf",
+        name="Test Device",
+    )
+    resp = make_clientapi_request(api_client, "/register", req)
+    assert resp["code"] == 200
+
+    user_clients = MobileClient.query(
+        MobileClient.user_id == str(none_throws(mock_clientapi_auth.account_key).id())
+    ).fetch()
+    assert len(user_clients) == 1
+
+    client = user_clients[0]
+    assert client.user_id == "1"
+    assert client.messaging_id == "abc123"
+    assert client.client_type == ClientType.OS_ANDROID_FCM
+    assert client.device_uuid == "asdf"
+    assert client.display_name == "Test Device"
+
+
 def test_register_existing_client(
     api_client: Client, mock_clientapi_auth: User
 ) -> None:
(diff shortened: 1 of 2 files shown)


phil-lopreiato (approved):
💯

PR #9033, opened 2026-02-13
Pin click==8.1.8 to resolve pyre-check/Flask dependency conflict

### LINKED ISSUE #9030: pyre-check dependency conflicts with Flask's click requirement
## Bug

`pyre-check==0.9.25` in `requirements.txt` requires `click<8.2.0`, but `Flask==3.1.2` (in `src/requirements.txt`) requires `click>=8.1.3`. pip resolves this to `click==8.3.1` which satisfies Flask but violates pyre-check's constraint:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed.
pyre-check 0.9.25 requires click<8.2.0,>=8.0, but you have click 8.3.1 which is incompatible.
```

This shows up when building the test Docker image, which installs both `requirements.txt` and `src/requirements.txt`.

## Context

The typecheck service and `make typecheck` target are already commented out in both `docker-compose.yml` and the `Makefile`. The maintainer has noted that pyre is "busted" and they'd like to move away from it.

## Fix

Remove `pyre-check==0.9.25` from `requirements.txt` since type checking is disabled. `pyre-extensions==0.0.32` in `src/requirements.txt` does not depend on `pyre-check` and is used independently for type annotations.

gregmarra:
## Summary
- Pins `click==8.1.8` in `requirements.txt` to resolve the dependency conflict between `pyre-check==0.9.25` (requires `click<8.2.0`) and `Flask==3.1.2` (requires `click>=8.1.3`)
- `click==8.1.8` is the latest 8.1.x release and satisfies both constraints
- Without this pin, pip defaults to `click==8.3.1` which violates pyre-check's constraint

## Test plan
- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] `[py3] Type Check` CI passes (pyre still works)

Fixes #9030

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ flake8-tidy-imports
 blinker==1.9.0
 freezegun==1.5.5
 packaging==26.0
+click==8.1.8
 pre-commit
 pyre-check==0.9.25
 pytest


phil-lopreiato:
we probably should migrate away from `pyre` though

PR #9059, opened 2026-02-15
PWA: Add favorite button to match page

gregmarra:
## Summary
- Adds FavoriteButton component to the match detail page header
- Allows users to favorite/unfavorite individual matches from the match page
- Consistent with existing favorite buttons on event and team pages

## Test plan
- [ ] Navigate to a match page (e.g., `/match/2024mil_f1m2`)
- [ ] Verify the star button appears next to the match title
- [ ] Click the star to toggle favorite (requires sign-in)
- [ ] Verify sign-in prompt appears for unauthenticated users
- [ ] Run `npm run typecheck` and `npm run lint` pass

### DIFF
--- a/pwa/app/routes/match.$matchKey.tsx
+++ b/pwa/app/routes/match.$matchKey.tsx
@@ -1,11 +1,12 @@
 import { createFileRoute, notFound } from '@tanstack/react-router';
 
 import { getEvent, getMatch } from '~/api/tba/read';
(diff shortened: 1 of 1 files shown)


gregmarra:
I think this doesn't actually make sense and we shouldn't do it. Events and Teams should be favorite-able, but the rest is too noisy.

PR #9080, opened 2026-02-15
PWA: Add District Points tab to event page

gregmarra:
## Summary
- Adds "District Points" tab to the event detail page
- Sortable table showing qual, elim, alliance, award, and total district points per team
- Tab only appears when district points data is available from the API
- Uses existing `DataTable` component for sorting

## Test plan
- [ ] Navigate to a district event (e.g., `/event/2024miket`) and verify District Points tab appears
- [ ] Verify table is sortable by clicking column headers
- [ ] Navigate to a regional event and verify District Points tab does NOT appear (or shows calculated points)
- [ ] Run `npm run typecheck` and `npm run lint` pass

## Screenshot Pages
- /event/2024oncmp1 District Event with Points

### DIFF
--- a/pwa/app/routes/event.$eventKey.tsx
+++ b/pwa/app/routes/event.$eventKey.tsx
@@ -103,7 +103,7 @@ import {
   TableRow,
 } from '~/components/ui/table';
 import { Tabs, TabsContent, TabsList, TabsTrigger } from '~/components/ui/tabs';
-import { SEASON_EVENT_TYPES } from '~/lib/api/EventType';
+import { DISTRICT_EVENT_TYPES, SEASON_EVENT_TYPES } from '~/lib/api/EventType';
 import { PlayoffType } from '~/lib/api/PlayoffType';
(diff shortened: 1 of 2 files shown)


pwa/app/routes/event.$eventKey.tsx
```diff
@@ -350,6 +356,14 @@ function EventPage() {
                 </InlineIcon>
               </TabsTrigger>
             )}
+          {districtPointsQuery.data && (
```
tervay: This will show district points on all events, since we calculate them for all events, even if they're not a district event. This should only show for district or district CMP event types

tervay (changes requested):

pwa/app/routes/event.$eventKey.tsx
gregmarra: Fixed — added a `DISTRICT_EVENT_TYPES` set (DISTRICT, DISTRICT_CMP, DISTRICT_CMP_DIVISION) and gated both the tab trigger and content on it. Also rebased onto main.

PR #9150, opened 2026-02-21
Add 2026 ranking sort orders

phil-lopreiato:
As title, so rankings will render.

towards github.com/the-blue-alliance/the-blue-alliance/issues/9135


bdaroz:
image
Looks good

PR #9204, opened 2026-02-28
Reorder district page tabs: Events first

gregmarra:
## Summary
- Reorders district page tabs to: Events, Teams, Rankings, Overview
- Sets Events as the default active tab (was Overview)

## Screenshot Pages
- /district/fim/2024
- /district/ne/2025

### DIFF
--- a/pwa/app/routes/district.$districtAbbreviation.{-$year}.tsx
+++ b/pwa/app/routes/district.$districtAbbreviation.{-$year}.tsx
@@ -187,15 +187,15 @@ function DistrictPage() {
         {districtHistory[districtHistory.length - 1].display_name} {year}
       </h1>
 
-      <Tabs defaultValue={'overview'} className="mt-4">
+      <Tabs defaultValue={'events'} className="mt-4">
         <TabsList
           className="flex h-auto flex-wrap items-center justify-evenly
             [&>*]:basis-1/2 lg:[&>*]:basis-1"
         >
-          <TabsTrigger value="overview">Overview</TabsTrigger>
-          {hasRankings && <TabsTrigger value="rankings">Rankings</TabsTrigger>}
           <TabsTrigger value="events">Events</TabsTrigger>
           <TabsTrigger value="teams">Teams</TabsTrigger>
+          {hasRankings && <TabsTrigger value="rankings">Rankings</TabsTrigger>}
+          <TabsTrigger value="overview">Overview</TabsTrigger>
         </TabsList>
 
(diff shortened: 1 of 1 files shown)


pwa/app/routes/district.$districtAbbreviation.{-$year}.tsx
```diff
@@ -187,15 +187,15 @@ function DistrictPage() {
...
         <TabsList
           className="flex h-auto flex-wrap items-center justify-evenly
             [&>*]:basis-1/2 lg:[&>*]:basis-1"
         >
-          <TabsTrigger value="overview">Overview</TabsTrigger>
-          {hasRankings && <TabsTrigger value="rankings">Rankings</TabsTrigger>}
           <TabsTrigger value="events">Events</TabsTrigger>
           <TabsTrigger value="teams">Teams</TabsTrigger>
+          {hasRankings && <TabsTrigger value="rankings">Rankings</TabsTrigger>}
+          <TabsTrigger value="overview">Overview</TabsTrigger>
```
tervay, line 198: Overview should still be the first listed tab, even if it isn't default

pwa/app/routes/district.$districtAbbreviation.{-$year}.tsx
```diff
@@ -187,15 +187,15 @@ function DistrictPage() {
         {districtHistory[districtHistory.length - 1].display_name} {year}
       </h1>
 
-      <Tabs defaultValue={'overview'} className="mt-4">
+      <Tabs defaultValue={'events'} className="mt-4">
```
tervay, line 190: I'm not sure that events is actually the most important view (default) here. Maybe we should simply rearchitect the layout of this route.

gregmarra:
I was mirroring the semantics of the Jinja site - filtering to a district shows you all the events from that district as the default. I think either we want an /events/ path that does that, or we want to use /districts/ as the HQ for a particular district (and currently the other apps default to events on districts).

How are you thinking about what to focus these pages on?

tervay:
I think this is one of the few areas that breaking changes from prod to beta is the right move. I think that going events -> district is weird because a district is more than a collection of events. It's an organization that consists of teams, events, ranked performances, community, and history.

I think that we should have a `/districts/` route which shows all districts (which you have in another PR) and `/districts/ne/<year>`, as well as `/districts/ne/history`, similar to team pages, I think.

In essence, I would like to move districts from being represented as a collection events, more into being a "first class" data type that we properly capture.

PR #9236, opened 2026-03-03
Nullapalooza: Add null safety for corrupted Datastore entities

gregmarra:
## Summary

Two production 500 errors are caused by corrupted Datastore entities with null values in `required=True` NDB fields. GAE's NDB `required=True` constraint is only enforced on `put()`, not on read, so corrupted data can crash at runtime.

- **`event.py`**: `event_teams_statuses` crashes on `event_team.team.id()` when an EventTeam entity has `team=None`. Now skips the corrupted entity and logs the key.
- **`team_renderer.py`**: `_fetch_events_async` crashes on `datetime.datetime(e.year, 12, 31)` when an Event entity has `year=None`, or when `ndb.get_multi_async` returns `None` for a deleted entity. Now filters both cases and logs.

All guards are tagged with `# Nullapalooza` comments so they can be easily found and removed once the root cause is identified and fixed.

## Test plan

- [x] Added `test_event_teams_statuses_corrupted_null_team` — verifies corrupted EventTeam with null team is skipped, valid teams still returned
- [x] Added `test_fetch_events_skips_null_year_event` — verifies corrupted Event with null year is filtered out  
- [x] Added `test_fetch_events_skips_none_events` — verifies None events from get_multi_async are filtered out
- [x] All existing event handler tests still pass (11/11)

### DIFF
--- a/src/backend/api/handlers/event.py
+++ b/src/backend/api/handlers/event.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Optional
 
 from flask import Response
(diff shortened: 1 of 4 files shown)


phil-lopreiato:
I don't think we need this anymore :)

PR #9239, opened 2026-03-03
Pickling Modernization

gregmarra:
## Summary
- Upgrade all `pickle.dumps` calls from protocol 2 (Py2 compat) to protocol 5 (modern Python 3)
- Remove `ImportFixingUnpickler` and `ImportFixingPickleProperty` — these handled Py2→Py3 module path renaming (`models.` → `backend.common.models.`) and are no longer needed since the Py2→Py3 migration is long complete
- Replace with a simplified `PickleProperty` using standard `pickle.dumps`/`pickle.loads`
- Remove `six` usage from deferred tests
- Bump `DATABASE_QUERY_VERSION` from 4 to 5 to invalidate any caches with old pickle format
- Simplify compatibility tests to verify Py3 round-trips only (removed Py2 base64-encoded test fixtures)

## Status
**DO NOT MERGE** — This is a work-in-progress modernization. Needs:
- [ ] Full test suite verification
- [ ] Review of any other pickle usage across the codebase
- [ ] Deployment plan for cache invalidation (version bump should handle this)
- [ ] Verify no production data still uses Py2 pickle format

## Test plan
- [ ] Run existing pickle compatibility tests
- [ ] Run full test suite to catch any regressions
- [ ] Verify CachedQueryResult round-trip with real model data
- [ ] Test deferred task serialization/deserialization

### DIFF
--- a/src/backend/common/helpers/deferred.py
+++ b/src/backend/common/helpers/deferred.py
@@ -19,7 +19,7 @@ _DEFAULT_URL = "/_ah/queue/deferred_encoded"
 
 def _serialize(obj: Any, *args, **kwargs) -> bytes:
     curried = _curry_callable(obj, *args, **kwargs)
-    return pickle.dumps(curried)
+    return pickle.dumps(curried, protocol=2)
 
 
 @ndb.tasklet
--- a/src/backend/common/helpers/tests/deferred_test.py
+++ b/src/backend/common/helpers/tests/deferred_test.py
@@ -1,3 +1,4 @@
+import six
 from google.appengine.api import taskqueue
 from google.appengine.ext import testbed
 
@@ -11,6 +12,7 @@ def _drain_deferred(
     assert len(tasks) == 1
 
     task = tasks[0]
+    six.ensure_text(task.payload)
     run_from_task(task)
 
 


src/backend/common/cache/flask_response_cache.py
```diff
@@ -19,7 +19,7 @@ def dump_object(self, value: Any) -> bytes:
         t = type(value)
         if t == int:
             return str(value).encode("ascii")
-        return b"!" + zlib.compress(pickle.dumps(value))
+        return b"!" + zlib.compress(pickle.dumps(value, protocol=5))
```
phil-lopreiato: we can remove this one, 5 is currently the default, so this is a noop

src/backend/common/helpers/deferred.py
```diff
@@ -19,7 +19,7 @@
 
 def _serialize(obj: Any, *args, **kwargs) -> bytes:
     curried = _curry_callable(obj, *args, **kwargs)
-    return pickle.dumps(curried, protocol=2)
+    return pickle.dumps(curried, protocol=5)
```
phil-lopreiato: I would set `pickle.DEFAULT_PROTOCOL` instead of hard coding numbers

src/backend/common/models/cached_query_result.py
```diff
@@ -1,55 +1,17 @@
...
-    def find_class(self, module, name):
-        renamed_module = module
-        if module.startswith("models."):
-            renamed_module = "backend.common." + module
-
-        return super().find_class(renamed_module, name)
-
-
-class ImportFixingPickleProperty(ndb.BlobProperty):
+class PickleProperty(ndb.BlobProperty):
```
phil-lopreiato: I don't think we need this class at all, we can use `ndb.PickleProperty`

https://github.com/GoogleCloudPlatform/appengine-python-standard/blob/413688fc87d7e8a3c2b2f2630ffc0d1c23eb62c9/src/google/appengine/ext/ndb/model.py#L1901

src/backend/common/queries/database_query.py
```diff
@@ -70,7 +70,7 @@ def fetch_dict_async(
 class CachedDatabaseQuery(
     DatabaseQuery, Generic[QueryReturn, DictQueryReturn], metaclass=abc.ABCMeta
 ):
-    DATABASE_QUERY_VERSION = 4
+    DATABASE_QUERY_VERSION = 5
```
phil-lopreiato, line 73: a followup... we should probably have an annual cronjob that cleans up data with old versions, huh...

phil-lopreiato (changes requested):
Some changes, but I think with them implemented, this becomes mergable (and a good idea!)

src/backend/common/cache/flask_response_cache.py
gregmarra: Done — removed the explicit `protocol=5` argument entirely, since it's the default.

src/backend/common/helpers/deferred.py
gregmarra: Done — went with omitting the protocol argument entirely (rather than `DEFAULT_PROTOCOL`) for consistency across all call sites. Since 5 is the default, the result is the same.

src/backend/common/models/cached_query_result.py
gregmarra: Done — removed the custom `PickleProperty` class entirely and switched to `ndb.PickleProperty(compressed=True)`. It uses `pickle.HIGHEST_PROTOCOL` under the hood, which is equivalent.

src/backend/common/queries/database_query.py
gregmarra: Agreed, that would be good. Filed as a follow-up idea — not in scope for this PR.

phil-lopreiato (approved):
hold onto your butts, let's give it a try

PR #9246, opened 2026-03-04
[pwa] Add ts-pattern for some nice pattern matching code

tervay:
Found this library and thought it was cool. Helps clean up some code and make it a little nicer on the developer.


gregmarra:
In some cases the code output is shorter, but I am not sure it is easier to read and comprehend. The code will be read much more than written, so I think the more-verbose-but-explicitly-all-together switch statements feel better to me.

PR #9261, opened 2026-03-04
Add pre_write/post_fetch validation that all model required propertie…

phil-lopreiato:
This might help us hunt down where our data corruption is coming from?

### DIFF
--- a/src/backend/common/models/cached_model.py
+++ b/src/backend/common/models/cached_model.py
@@ -84,7 +84,9 @@ class CachedModel(ndb.Model):
         Hook called before the entity is written to the datastore.
         Validates all required properties are set.
         """
-        # Only validate if this is a CachedModel instance with the method
+        # Check if the method exists before calling. While _validate_required_properties
+        # is defined on CachedModel, NDB's hook system can invoke this on different
+        # model types (especially when tests manipulate NDB's kind map).
         if hasattr(self, "_validate_required_properties"):
             self._validate_required_properties()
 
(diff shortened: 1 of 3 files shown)


phil-lopreiato:
Changes Made
1. Clarified [hasattr](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) checks with better comments
Files: [cached_model.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [cached_query_result.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Reason: While the checks initially appeared redundant (methods are defined on the same class), they're actually necessary for defensive programming. NDB's hook system can invoke these hooks on different model types, especially when tests manipulate NDB's kind map.
Solution: Added clear explanatory comments explaining why these guards are needed:
Edited
cached_query_result_test.py+0-1
2. Refactored [test_pre_put_hook_validates](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) test
[cached_query_result_test.py](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Reason: The test was relying on NDB's [BadValueError](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) which might be raised for different reasons. The focus should be on what our validation hook does (logging), not on NDB's incidental behavior.
Solution: Changed the test to:
Use try/except instead of [pytest.raises](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to handle NDB's validation independently
Assert that our hook logs the error correctly (using [caplog](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html))
Accept that NDB might raise exceptions for its own reasons, while our hook still does its job

PR #9265, opened 2026-03-05
[pwa] Add Google/Apple calendar event links

tervay:
image

image

When clicking Google Calendar icon:

image

When clicking Apple calendar icon:

image

image

May help the overwhelmed parents attempting to keep track of their children


gregmarra:
I think that this is kind of visually cluttered for how often these pages will be viewed compared to how often these buttons will be used. My guess is that most teams probably send out calendar invites to their mailing lists, or something.

If we wanted to do these, I'd add them behind a [...] menu next to the dates, and include at least Outlook as well.

tervay:
image

image

image

image

PR #9274, opened 2026-03-05
manipulator: fix corrupted models by treating them as missing

phil-lopreiato:
Now that we have a way to identify corrupt models in the datastore (although the "how" is still a baffling mystery to me), this will teach the manipulators to try and self-heal.  If we load a corrupt model (defined as missing a required property) in `findOrSpawn`, then treat it as missing. This way, if we write the "old" model, we should be able to try and recover.

### DIFF
--- a/src/backend/common/models/cached_model.py
+++ b/src/backend/common/models/cached_model.py
@@ -48,14 +48,17 @@ class CachedModel(ndb.Model):
         # constructors, so make sure we have a common set of properties defined
         self._fix_up_properties()
 
-    def _validate_required_properties(self) -> None:
+    def _validate_required_properties(self) -> bool:
         """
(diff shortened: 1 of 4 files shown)


src/backend/common/models/cached_model.py
```diff
@@ -48,14 +48,17 @@ def __init__(self, *args, **kwargs):
         # constructors, so make sure we have a common set of properties defined
         self._fix_up_properties()
 
-    def _validate_required_properties(self) -> None:
+    def _validate_required_properties(self) -> bool:
         """
         Validates that all required properties on the model are set.
         Logs an error with stack trace and model key if validation fails.
+
+        Returns True when one or more required properties are missing,
```
gregmarra, line 56: Confusing but explained.

PR #9278, opened 2026-03-06
Fix notification times displaying UTC instead of local timezone

gregmarra:
## Summary
- Notification times (schedule updated, event level starting, match upcoming) were displaying the raw UTC time but labeling it with the event's local timezone abbreviation
- For example, a match at 14:55 UTC at an event in CST would show "14:55 CST" instead of "08:55 CST"
- Now properly converts UTC to local time using `pytz.utc.localize().astimezone()` before formatting
- Fixed in all three notification types: `EventScheduleNotification`, `EventLevelNotification`, `MatchUpcomingNotification`

## Test plan
- [x] Updated existing timezone tests to assert correct UTC→local conversion
- [ ] CI passes

### DIFF
--- a/src/backend/common/models/notifications/event_level.py
+++ b/src/backend/common/models/notifications/event_level.py
@@ -21,18 +21,21 @@ class EventLevelNotification(Notification):
     @property
     def fcm_notification(self) -> Optional[Any]:
         if self.match.time:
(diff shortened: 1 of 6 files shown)


phil-lopreiato (approved):

gregmarra:
bypassed flaky ci

PR #9049, opened 2026-02-14
Remove the Vagrantfile, use docker-compose instead

ZachOrr:

PR #9283, opened 2026-03-06
Replace Vagrant with direct Docker in CI

gregmarra:
## Summary
- Remove "Install Vagrant" step from both CI workflows (saves ~30s of apt install)
- Replace `vagrant up` with `docker run -d` using bind mount for source + named volume for `node_modules`
- Update `ops/test_vagrant_startup.py` to use `docker exec` instead of `vagrant ssh`
- Update `ops/dev/pull-gae-logs.sh` to use `docker cp` instead of vagrant SSH + scp

## Motivation
Vagrant adds unnecessary overhead in CI: installing it (~30s), SSH setup, rsync sync. Since CI already builds the Docker image directly, we can skip the Vagrant layer and run the container directly.

The `Vagrantfile` is kept in the repo for local dev users who prefer it.

## Test plan
- [ ] Ops Fullstack Test CI job passes on both `pull_request` and `push` workflows
- [ ] Container starts, all modules boot, webpack compiles, homepage returns 200
- [ ] `test_ops.sh` passes (uses `docker exec` for log retrieval)

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -444,6 +444,7 @@ jobs:
             -p 9000:9000 \
             -p 9005:9005 \
             -p 9099:9099 \
+            -e CI=true \
             --workdir /tba \
             localhost:5000/tba-py3-dev:latest \
             bash -c "./ops/dev/vagrant/bootstrap-dev-container.sh && ./ops/dev/vagrant/start-devserver.sh && sleep infinity"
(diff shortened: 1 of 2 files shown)


ZachOrr:
I think this one might be out of date now Greg with https://github.com/the-blue-alliance/the-blue-alliance/pull/9049

gregmarra:
Superseded by #9049

PR #9284, opened 2026-03-06
Reduce startup poll interval from 5s to 2s in CI

gregmarra:
## Summary
- Reduce `time.sleep(5)` to `time.sleep(2)` in `ops/test_vagrant_startup.py`

## Motivation
The startup polling loop checks whether all GAE modules and webpack have started, sleeping 5s between checks. With ~15 polling iterations, we waste an average of ~2.5s per overshoot. Reducing to 2s saves ~45s total while still being gentle on the container.

## Test plan
- [ ] Ops Fullstack Test CI job passes
- [ ] No false negatives from polling too fast (2s is still plenty for log checks)


gregmarra:
This would only save time on the last loop, so like 3 seconds.

Issue #9249, opened 2026-03-04
Nullapalooza: GAE NDB Datastore entities with null required fields: comprehensive corruption report


gregmarra:
## Summary

Google App Engine NDB allows entities with `required=True` properties to be **read from Datastore with `None` values** — validation is only enforced on writes. This has caused persistent production 500 errors across multiple model types in The Blue Alliance. This issue documents the full scope, the NDB internals that allow it, and what we know about how entities end up in this state.

This report is intended to be useful to **other GAE/NDB projects** that may be hitting the same class of bug.

### What `required=True` actually does

In `google.appengine.ext.ndb.model` (package `appengine-python-standard`), the `required` flag is stored as `self._required` on the property. It is checked by exactly one method:

```python
# model.py — Property._is_initialized
def _is_initialized(self, entity):
    return (not self._required or
            ((self._has_value(entity) or self._default is not None) and
             self._get_value(entity) is not None))
```

This is called by `Model._check_initialized()`, which raises `BadValueError` if any required properties are uninitialized:

```python
# model.py — Model._check_initialized
def _check_initialized(self):
    baddies = self._find_uninitialized()
    if baddies:
        raise datastore_errors.BadValueError(
            'Entity has uninitialized properties: %s' % ', '.join(baddies))
```

### Write path: validated

`_check_initialized` is called from **exactly one place** — `Model._to_pb()`:

```python
# model.py — Model._to_pb
def _to_pb(self, pb=None, allow_partial=False, set_key=True):
    if not allow_partial:
        self._check_initialized()   # <-- ONLY call site in all of NDB
    ...
```

`_to_pb` is invoked during `put()`, so writes are validated.

### Read path: NOT validated

`Model._from_pb()` — the deserialization path used when reading entities from Datastore — **never calls `_check_initialized`**. It creates an empty entity, deserializes each property from the protobuf, and returns it. If a required property is missing from the stored protobuf, the entity is returned with `None` for that field — no error, no warning.

```python
# model.py — Model._from_pb (simplified)
@classmethod
def _from_pb(cls, pb, set_key=True, ent=None, key=None):
    if ent is None:
        ent = cls()              # No validation
    # ... deserialize properties from pb ...
    return ent                   # Returned without validation
```

### The self-healing paradox

[...]

Issue #9237, opened 2026-03-03
Investigate and fix NDB model corruption causing 500s (Nullapalooza)


gregmarra:
## Problem

Production 500 errors are caused by corrupted Datastore entities with null values in `required=True` NDB fields. NDB's `required=True` is only enforced on `put()`, not on read — so corrupted data passes through without error until application code accesses the null field.

Two specific crashes were found and patched in #9236 (tagged `# Nullapalooza` for easy grep):
1. `EventTeam.team` is `None` → crash in `event_teams_statuses` on `.id()`
2. `Event.year` is `None` → crash in `_fetch_events_async` on `datetime.datetime(e.year, ...)`

## Root cause (unknown)

How do entities end up with null required fields? Possible causes:
- Bulk migration or data repair script that didn't set all fields
- A manipulator bug that writes partial entities
- Schema change that added `required=True` to a field that previously allowed null
- Direct Datastore writes bypassing NDB validation (e.g., from Cloud Console or migration tooling)

The `EventTeam` model even has a comment on `year`: *"technically immutable, but corruptable and needs repair. See github issue #409"* — confirming this is a known historical issue.

### 1. `None` from `ndb.get_multi_async` (deleted/missing entities)

When a join table (e.g., `EventTeam`) references an entity key that no longer exists in Datastore, `ndb.get_multi_async()` returns `None` for that slot. Some queries already handle this, others don't.

**Already filtering None:**
- `TeamEventMatchesQuery` — `list(filter(None, matches))`
- `TeamYearMatchesQuery` — `list(filter(None, matches))`
- `TeamDistrictsQuery` — `list(filter(lambda x: x is not None, districts))`
- `DistrictQuery` — conditional `if district:` check

**NOT filtering None (10 queries):**
- `TeamEventsQuery`, `TeamYearEventsQuery`, `EventDivisionsQuery`
- `DistrictChampsInYearQuery`, `CmpDivisionsInYearQuery`
- `DistrictTeamsQuery`, `EventTeamsQuery`
- `DistrictsInYearQuery`, `DistrictHistoryQuery`, `DistrictAbbreviationQuery`

### 2. Corrupted required fields (bad data in Datastore)

The entity exists but has null values where `required=True` is declared. NDB doesn't validate on read. Models with `required=True` fields that could be corrupted include:

- `EventTeam`: `event`, `team`, `year`
- `Event`: `event_type_enum`, `event_short`, `year`
[...]

PR #9287, opened 2026-03-06
cachedmodel: remove call to fix_up_properties

phil-lopreiato:
This is actually wrong!

And could have been the source of our corruption (although it's old, so idk why it started recently)

The call [here](https://github.com/the-blue-alliance/the-blue-alliance/blob/25a763a9a3e124b97b219481a8bed69fabc75c6c/src/backend/common/models/cached_model.py#L49) --> [to here](https://github.com/GoogleCloudPlatform/appengine-python-standard/blob/413688fc87d7e8a3c2b2f2630ffc0d1c23eb62c9/src/google/appengine/ext/ndb/model.py#L3412) meant this could have been a concurrency bug, where we broke the state tracking what properties belong to each model kind. This map is used for all sorts of things in the serialization layer, so if we ended up reading at a state where it was "empty", then this could have been the cause.

I think this was originally here to paper over an original bug in the tests (where multiple things were called `DummyModel`, which caused oddness with ndb's kind map tracking, behaving differently depending on import order). So, we rename the test models for independence 

This is the stack that tipped me off

```
root: dictionary changed size during iteration
Traceback (most recent call last):
  File "/workspace/backend/common/queries/database_query.py", line 148, in _do_query
    ).put_async()
      ~~~~~~~~~^^
  File "/layers/google.python.pip/pip/lib/python3.13/site-packages/google/appengine/ext/ndb/model.py", line 3562, in _put_async
    self._pre_put_hook()
    ~~~~~~~~~~~~~~~~~~^^
  File "/workspace/backend/common/models/cached_query_result.py", line 94, in _pre_put_hook
    for prop_name, prop in model._properties.items():
                           ~~~~~~~~~~~~~~~~~~~~~~~^^
RuntimeError: dictionary changed size during iteration
```

Which got me looking at things that affected the `_properties` map at runtime

### DIFF
--- a/src/backend/common/manipulators/tests/base_manipulator_test.py
+++ b/src/backend/common/manipulators/tests/base_manipulator_test.py
@@ -77,7 +77,9 @@ class DummyCachedQuery(CachedDatabaseQuery[ManipulatorDummyModel, None]):
     DICT_CONVERTER = DummyConverter
 
     @ndb.tasklet
-    def _query_async(self, model_key: str) -> Generator[Any, Any, ManipulatorDummyModel]:
+    def _query_async(
+        self, model_key: str
+    ) -> Generator[Any, Any, ManipulatorDummyModel]:
         model = yield ManipulatorDummyModel.get_by_id_async(model_key)
         return model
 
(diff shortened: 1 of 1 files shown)


gregmarra:
Part of the Nullapalooza investigation (#9237, #9249) — this concurrency bug in `_fix_up_properties()` is a likely root cause of the model corruption.

gregmarra:
Fix of the week!

phil-lopreiato:
@fangeugene thinks this could have been originally exposed by https://github.com/the-blue-alliance/the-blue-alliance/pull/8581

PR #9292, opened 2026-03-06
time predictions: preserve scheduled time for first match of the day

phil-lopreiato:
For the first match, if nexus hits "on field" before opening ceremonies, the predictive time can be super early.

This diff fixes the bug where we keep the scheduled time for the first match of the day


gregmarra:
A couple things:

1. Line 230 (the "different day / different comp level" branch) also uses `nexus_predicted_time or ...` but isn't patched here. If Nexus sends bogus predictions for next-day matches, that path still trusts them.

2. The test passes `is_live=False`, but the real bug happens when `is_live=True`. With `is_live=False`, `now` is set to epoch, so `first_unplayed_timedelta` gets clamped to 0 regardless — the test passes without actually exercising the buggy code path. Should use `is_live=True` to be a meaningful regression test.

PR #9324, opened 2026-03-08
Add score breakdown to seeded test matches

gregmarra:
## Summary

- Adds `score_breakdown_json` with random bonus RPs and total points to seeded test event matches
- Makes test data more realistic for local dev, allowing features that depend on score breakdowns to be exercised

## Test plan

- [ ] Run "Seed Test Event" from `/local/bootstrap` and verify matches have score breakdowns
- [ ] Verify match detail pages show score breakdown data

### DIFF
--- a/src/backend/web/local/dev_tools.py
+++ b/src/backend/web/local/dev_tools.py
@@ -74,6 +74,18 @@ def seed_test_event() -> Response:
         match_time = total_completed_time + datetime.timedelta(
             minutes=(i - 1) * MATCH_SPACING_MINUTES
         )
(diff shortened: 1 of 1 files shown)


phil-lopreiato:
This might be problematic in some cases - I think there's code that assumes, if it exists, the breakdown matches the expected format. The frc api types do have the schemas for this, so we could at least have a year -> type mapping and fill them in?

Or if you discover the main site works fine with this after opening a match page, then maybe it'll be fine

gregmarra:
This is less important now that we have 2026 matches played. Have been primarily improving these to make testing easier.

PR #9342, opened 2026-03-09
move all youtube fetches to datafeed pattern

phil-lopreiato:
This will let us remove the random urlfetch calls, and instead use the consistent datafeed pattern (for API key management, error handling, etc), and improve type hints for the API response schema.


src/backend/common/helpers/youtube_video_helper.py
```diff
@@ -37,6 +36,11 @@ class YouTubeVideoHelper(object):
         r".*/shorts/([a-zA-Z0-9_-]*)",  # Shorts: youtube.com/shorts/ID
     ]
 
+    @classmethod
+    def _youtube_datafeed_class(cls, class_name: str) -> Any:
```
gregmarra: The `importlib` dynamic import breaks type safety, IDE navigation, and makes it easy to pass a wrong class name string. This exists to avoid a circular import (`common/helpers` → `tasks_io/datafeeds`), but I think the right fix is to complete the refactor: move the API-calling methods (`get_scheduled_start_time`, `resolve_channel_name`, `videos_in_playlist`, `get_upcoming_streams`) out of this helper and into `tasks_io/` — either as methods on the datafeed classes or a new facade. These methods are I/O, not pure utilities, so they belong in the tasks layer.

That leaves `YouTubeVideoHelper` with only the pure utility methods (URL parsing, match partial guessing) and no need for `importlib` at all.

src/backend/common/helpers/youtube_video_helper.py
```diff
@@ -105,45 +109,41 @@ def get_scheduled_start_time(
...
-            scheduled_start_time = live_details.get("scheduledStartTime")
+            parsed_data = datafeed.parser().parse(raw_data)
+            scheduled_start_time = (
+                parsed_data.get("scheduled_start_time") if parsed_data else None
+            )
+
+            if not scheduled_start_time:
+                live_details = raw_data["items"][0].get("liveStreamingDetails", {})
+                scheduled_start_time = live_details.get("scheduledStartTime")
+
```
gregmarra: This fallback re-parses `raw_data` manually when `parsed_data` returns None for `scheduled_start_time`. But the parser (`YoutubeVideoDetailsParser`) does the exact same extraction — `live_details.get("scheduledStartTime")[:10]`. If the parser returns None, this fallback will too. And note this fallback doesn't truncate to `[:10]`, so it could return a full ISO timestamp where the parser returns YYYY-MM-DD, causing inconsistent behavior.

I'd drop the fallback and just trust the parser. If a parser bug surfaces later, fix it in the parser.

src/backend/common/helpers/youtube_video_helper.py
```diff
@@ -161,41 +161,33 @@ def resolve_channel_name(
...
+            if not raw_data or not raw_data.get("items"):
                 raise ndb.Return(None)
 
-            first_item = data["items"][0]
-            resolved_channel_id = first_item.get("id", {}).get("channelId")
-            resolved_channel_name = first_item.get("snippet", {}).get("title")
+            parsed_data = datafeed.parser().parse(raw_data)
+            if parsed_data:
+                first_item = parsed_data[0]
+                resolved_channel_id = first_item.get("channel_id")
```
gregmarra, line 179: Same fallback duplication concern here — two parallel parsing paths for the same data. If `parsed_data` is empty, the raw fallback does the same extraction the parser would have done. Drop the `else` branch and trust the parser.

src/backend/common/helpers/youtube_video_helper.py
```diff
@@ -312,75 +286,75 @@ def get_upcoming_streams(
...
-                        {
-                            "stream_id": video_id,
-                            "title": snippet.get("title", ""),
-                        }
-                    )
+            parsed_search_results = search_datafeed.parser().parse(search_result)
+            if parsed_search_results:
+                for item in parsed_search_results:
+                    video_id = item.get("video_id")
+                    if video_id:
```
gregmarra: Same pattern — fallback raw parsing duplicates what the parser does. If `parsed_search_results` is empty (no items), the raw path iterating `search_result.get("items", [])` would also produce nothing. Drop the `else` branch.

src/backend/common/datafeeds/parsers/youtube/youtube_search_parser.py
```diff
@@ -0,0 +1,123 @@
...
+                results.append(parsed)
+
+        return results
+
+    def _parse_item(self, item: _SearchResult) -> "ParsedSearchResult | None":
+        """Parse a single search result item."""
+        item_id = item.get("id")
+        snippet = item.get("snippet")
+
+        if not item_id or not snippet:
```
gregmarra, line 92: Minor: this overwrites `channel_id` (previously set from `item_id.get("channelId")` on line 84) with `snippet.channelId`. And if `snippet.channelId` is missing, it falls back to the channel *title* as the channel *id* via the default arg, which seems like a bug — a channel title is not a channel ID.

src/backend/common/helpers/youtube_video_helper.py
gregmarra: The `importlib` removal is good — thanks for fixing that. But the I/O methods (`get_scheduled_start_time`, `resolve_channel_name`, `videos_in_playlist`, `get_upcoming_streams`) still live in `common/helpers/`, and the datafeed classes moved to `common/datafeeds/` (with re-export stubs in `tasks_io/datafeeds/`). I think this moves in the wrong direction architecturally. Here's why:

**`common/` should stay I/O-free.** Today, `common/` contains models, queries, manipulators, consts, and pure helpers — no HTTP calls. That's what makes it reliably testable and easy to reason about. Putting YouTube API datafeed classes and I/O orchestration methods in `common/` breaks that invariant.

**The existing datafeed pattern lives in `tasks_io/datafeeds/`.** FMS API, Twitch, HTML, and Nexus datafeeds all live there. YouTube should follow the same convention rather than creating a parallel `common/datafeeds/` hierarchy.

**The cross-service caller problem is solvable.** I checked — `web/` and `common/` never import from `tasks_io/` (they're separate GAE services). The callers that create the apparent need for `common/` placement:

- `tasks_io/handlers/live_events.py` → already in `tasks_io/`, no problem
- `web/handlers/admin/districts.py` (`resolve_channel_name`) → this handler can call the YouTube search datafeed directly instead of going through the helper
- `web/handlers/suggestions/suggestion_submission.py` (`videos_in_playlist`) → same, call the playlist datafeed directly
- `common/suggestions/suggestion_creator.py` (`get_scheduled_start_time`) → this is the only tricky one. The I/O should be lifted to the handler level: the web/tasks handler that invokes `suggestion_creator` should fetch the scheduled start time and pass it in as a parameter, rather than `suggestion_creator` reaching out to YouTube mid-execution

That last refactor is small but important — it keeps `common/` purely about data manipulation, and pushes I/O to the edges (handlers).

**End state I'd target:**
- `tasks_io/datafeeds/datafeed_youtube.py` — YouTube API datafeed classes (alongside the existing FMS/Twitch/etc datafeeds)
- `common/helpers/youtube_video_helper.py` — pure utility methods only (URL parsing, `time_to_seconds`, `guessMatchPartial`)
- `web/` and `tasks_io/` handlers call datafeed classes directly for I/O
- No `common/datafeeds/` package, no re-export stubs

PR #9365, opened 2026-03-10
Fix header offset in Scalar apidocs

GrahamSH-LLK:
## Description
Previously the header overlapped when you scrolled. Now it doesn't.
## Motivation and Context

I think this was removed at some point during the apidocs PR.
## How Has This Been Tested?

Tested locally, works.
## Screenshot Pages

- /apidocs/v3
## Types of changes


tervay:
fixed in #9618

PR #9391, opened 2026-03-11
[pwa] Migrate to oxlint and oxfmt

tervay:
Migrates from eslint and prettier to oxlint and oxfmt. These are by voidzero which is the same group that makes vite and vitest.

These tools are significantly faster (50-100x) than the alternatives. Both complete in under 1 second.

Removes 18 different devdependencies in favor of 2.

Most linting and formatting rules are identical, but there are some minor changes. Most notably, we lose tailwind class wrapping - I'd like to look into getting this back in the future.

Overall I think its worth it, given the speed, especially on the formatting end (~20sec on my machine to 0.4 sec).

These were created with LLMs following these instructions:

https://oxc.rs/docs/guide/usage/linter/migrate-from-eslint.html
https://oxc.rs/docs/guide/usage/formatter/migrate-from-prettier.html

There are a lot of manually configured rules from the auto-migration script that the agents ran (see above links). I attempted to remove these but there are a lot of warnings that get created. These are fixable warnings, but are too much for this PR. I will work on fixing them over time and moving towards a more default oxlint config.


tervay:
Closing to split into multiple prs

PR #9411, opened 2026-03-13
Add pwa preview deployments (opt-in)

tervay:
Adds a two-workflow system for deploying preview PWA builds to GAE. A maintainer (or anyone) can apply the `preview-pwa` label to a PR to trigger a preview deployment, including PRs from forks.

  **How it works:**
  - `pwa_preview_build.yml`: triggered by `pull_request`, builds the PWA and uploads the artifact + PR metadata. No secrets available here, so fork code is safe to run.
  - `pwa_preview.yml`: triggered by `workflow_run` (deploy) and `pull_request_target` (cleanup). Downloads the pre-built artifact and deploys to GAE — no fork code runs here, so secrets are safe to use.

  Each PR gets its own GAE version (`pr-{number}`) with a dedicated preview URL. Previews are cleaned up when the PR is closed or the label is removed. A `pwa-preview.yaml` with `max_idle_instances: 0` is used instead of the production config to avoid keeping idle instances warm.

### DIFF
--- a/.github/workflows/pwa_preview.yml
+++ b/.github/workflows/pwa_preview.yml
@@ -1,0 +1,164 @@
+name: PWA Preview Deploy
+
+on:
+  # Triggered after a successful PWA Preview Build run — downloads the pre-built artifact
+  # and deploys it. Secrets are available here; no fork code is executed.
+  workflow_run:
+    workflows: ["PWA Preview Build"]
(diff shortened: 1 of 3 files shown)


phil-lopreiato (approved):
Oh, cleaning up the version when the PR closes is such a good idea

PR #9426, opened 2026-03-13
add event team list update notifications

mrhappyma:
## Description
I added a webhook/push notification for teams being added to or removed from an event, and documentation for the webhook event. i hope I did it right :)

## Motivation and Context

My team is building a scouting app that needs to download a list of teams at our events, and I would like to receive webhook events for when that list of teams is updated before a competition. It would also just be nice to know when the list changes

Currently it will fire 2 separate notifications if teams are added and removed at the same time, this doesn't feel ideal but I couldn't find a clean way to do it otherwise. I'd be happy to change it if you can point me in the right direction.

## How Has This Been Tested?
I tested this locally using a test offseason event and the events wizard page, seeing that my webhook received events as I updated the data. I tested other webhook events and subscriptions as well.

## Types of changes


phil-lopreiato:
Thanks for submitting this diff! I'll let @ZachOrr take the review for this one

src/backend/common/helpers/tbans_helper.py
```diff
@@ -2,6 +2,7 @@
 import enum
 import logging
 import time
+from typing import List
```
ZachOrr, line 5: nit: We don't use the `from typing import ...` anymore (if we don't have to). We should be able to use the builtin `list` now.

ZachOrr (changes requested):
Thanks for this PR! I think this idea makes a lot of sense.

I think we should keep the notifications less state-ful. Ex: For webhook subscribers - I think we ought to just send the event + the full list of teams. Clients can implement the diffing if they'd like. I think from a TBA-side feature, just sending all the teams for the event is fine. For dispatching on our end, we'd likely want to use the diffs (so only send to teams that were added).

The mobile push notification messages might need to be reworked as such. Currently I think the messages are going to be way too long trying to list the added/removed.

The "removed" is a really interesting case to think about for a Team subscriber. Should we dispatch a notification if a team gets removed from an Event? I don't think we have to. Worth thinking through.

There's also some small asks for this PR that I have. Things like - make sure we're not using the `typing` module if we don't need it. I was working on this a few months ago with PRs like https://github.com/the-blue-alliance/the-blue-alliance/pull/8536 - would be nice not to add to that tech debt moving forward.

If we could add an example of the payload in the `src/backend/common/models/notifications/examples` folder that would be helpful as well.

Let's make some of the higher-level changes and I can do a more granular code review after 🙏

ZachOrr:
@mrhappyma Hey! Just checking in on this one. Are you still interested in trying to get this landed?

PR #9711, opened 2026-04-18
District ranking tiebreaks

AGPapa:
## Description
The manual says that the final tiebreaks should be highest match score, including the playoffs, but TBA's implementation only looked at qual matches.

This rule has been like this since at least 2017 (the manual doesn't include ranking rules in years before that).

## Motivation and Context
#9678 shows that we are matching points, but the order is incorrect.

## How Has This Been Tested?
I cross referenced the output for the 2025 Michigan District rankings from this and the official rankings, confirming all teams matched.

image

## Types of changes


src/backend/common/helpers/district_helper.py
```diff
@@ -800,15 +800,15 @@ def _calc_rank_based_match_points(
 
         _count, organized_matches = MatchHelper.organized_matches(matches)
 
-        # qual match calculations. only used for tiebreaking
-        for match in organized_matches[CompLevel.QM]:
```
AGPapa, line 804: This is the only critical change to fix the rankings. If we want to keep the name "highest_qual_scores" we can undo the rest of the PR and keep just this change (and needed test changes)

ZachOrr (changes requested):
This is great! I have no qualm with the rename here - I don't mind changing that variable name everywhere.

Help me though - where do we actually change the callsite into `DistrictRankingTeamTotal`? That part I think is the only thing missing for me logically here. Like - I can't spot the line of code where we went from just giving the helper qual scores -> giving it all matches.

The other thing we need to look at here is - do we need to roll the district ranking cache version to blow the old cached models for this?

Last big thing here - the rename in the API is a breaking change for consumers. At minimum we need to call it out in the API changelog. We should check if any of our first-party apps (Android and PWA - I know iOS doesn't use this directly) use this field and if we need to do fast follows there.

ZachOrr:
Alright gonna merge this one myself! Addressing my feedback -

I see where we use `matches` - it's like, one line above the diff. I don't know why I didn't see that before.

There isn't an API cache for the district rankings - just our backend cache that should get cleared when we recalculate.

The PWA is obviously good because this PR is good. Mobile apps should be good as well. Sorry API consumers 🙏

ZachOrr (approved):

bdaroz:
One note, a field rename is a very breaking change for many of the auto-generated code paths from this model and probably best not done mid-season, or at least done as an _additional_ field (dupe the values and deprecate). Either way, would be at least minor version update, not a point release per the versioning spec.

ZachOrr:
@bdaroz Absolutely acknowledge that! I kind of yolo'd this one knowing the implications. If this were a more prominent field like match score or something, or if any of the first-party clients used this field explicitly (a breaking change) - we would've held it and moved deliberately. But it's the 6th tiebreaker field for district rankings that folks already don't use/trust as the source of truth because we don't calculate them properly. The pros of moving towards being more trusted as a source of truth outweighed the theoretical cons for me here.

ZachOrr:
@bdaroz Absolutely acknowledge that! I kind of yolo'd this one knowing the implications. If this were a more prominent field like match score or something, or if any of the first-party clients used this field explicitly (a breaking change) - we would've held it and moved deliberately. But it's the 6th tiebreaker field for district rankings that folks already don't use/trust as the source of truth because we don't calculate them properly. It pros of moving towards being more trusted as a source of truth outweighed the theoretical cons for me here.

PR #9734, opened 2026-04-21
Add enum types for bare integers in API spec

ZachOrr:
Working against some of the playoff stuff in the iOS repo using the `PlayoffType` I was like - why don't we surface these as enums in the API spec? This PR adds enum values (both the ints and the strings that match them) in our API spec.


tervay (changes requested):
There is a more approved way to do this outlined here: https://spec.openapis.org/oas/v3.1.1.html#model-with-annotated-enumeration

so it would look like:

```json
  "AwardType": {
    "title": "AwardType",
    "oneOf": [
      { "const": 0, "title": "CHAIRMANS" },
      { "const": 1, "title": "WINNER" },
      ...
    ]
  }
```

then, within pwa `openapi-ts/read.config.ts`, we add typescript enums to the config:

```ts
  {
    name: '@hey-api/types',
    enums: 'typescript',
  }
```

which SHOULD generate:

```ts
  export enum AwardType {
    CHAIRMANS = 0,
    WINNER = 1,
    ...
  }
```

ZachOrr:
That's way better! Lemme look into this. Thanks Justin.

PR #9749, opened 2026-04-22
First prototype of v2 insights

tervay:
Provides a simple implementation of blue banner insights according to the plan at #9511. Intentionally not exposed on the api spec since it is not stable and the url will likely change. I would like to merge this in to verify functionality before proceeding further.

### DIFF
--- a/src/backend/common/helpers/insights_v2/blue_banners.py
+++ b/src/backend/common/helpers/insights_v2/blue_banners.py
@@ -1,0 +1,48 @@
+from collections import defaultdict
+from typing import Dict, Optional
+
+from backend.common.consts.award_type import BLUE_BANNER_AWARDS
+from backend.common.helpers.insights_v2.compute import (
+    build_leaderboard_rankings,
+    InsightV2Calculator,
+)
+from backend.common.models.event import Event
+from backend.common.models.insight_v2 import (
+    InsightCategory,
+    InsightV2,
+    InsightV2Name,
+    LeaderboardDataV2,
+)
+from backend.common.models.keys import Year
+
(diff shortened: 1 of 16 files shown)


src/backend/common/models/insight_v2.py
```diff
@@ -0,0 +1,76 @@
...
+    V2 insight model. Fully typed, category-discriminated.
+    Coexists with Insight (v1) until full cutover.
+    key_name format: {year}_v2_{category}_{name}[_{district}]
+    """
+
+    name = ndb.StringProperty(required=True)
+    display_name = ndb.StringProperty(required=True)
+    year = ndb.IntegerProperty(required=True)  # 0 = all-time
+    category = ndb.StringProperty(required=True, indexed=True)
+    data_json = ndb.TextProperty(required=True, indexed=False)
```
phil-lopreiato: We can probably just make this `JSONProperty`? (and maybe also set `compressed=True`?)

src/backend/common/helpers/insights_v2/blue_banners.py
```diff
@@ -0,0 +1,49 @@
...
+)
+from backend.common.models.event import Event
+from backend.common.models.insight_v2 import (
+    InsightCategory,
+    InsightV2,
+    LeaderboardDataV2,
+)
+from backend.common.models.keys import Year
+
+BLUE_BANNERS_NAME = "blue_banners"
```
phil-lopreiato: I think we should make make this an overall enum, so we have a listing of all valid types?

phil-lopreiato (approved):
LGTM!

PR #9790, opened 2026-04-27
Add code in trusted API to remove invalid matches like f1m3 for offseason events

ZachOrr:
Was spot-checking some matches today to try to find one with an unplayed match and I found a few for 2025. All of these were offseason matches and they were all f1m3.

| Match Key | Event Key |
|---|---|
| 2025cabl_f1m3 | 2025cabl |
| 2025cacc_f1m3 | 2025cacc |
| 2025cacg_f1m3 | 2025cacg |
| 2025cafb_f1m3 | 2025cafb |
| 2025catt_f1m3 | 2025catt |
| 2025iltrr_f1m3 | 2025iltrr |
| 2025ncfay_f1m3 | 2025ncfay |
| 2025nmrc_f1m3 | 2025nmrc |

The trusted API doesn't do any sort of match cleanup like we do for official FRC events. This PR adds a call to `MatchHelper.delete_invalid_matches` in the trusted API to try to help clean up these invalid matches before they enter our system.

### DIFF
--- a/src/backend/api/handlers/trusted.py
+++ b/src/backend/api/handlers/trusted.py
@@ -325,6 +325,19 @@ def update_event_matches(event_key: EventKey) -> Response:
 
     if event.remap_teams:
         EventRemapTeamsHelper.remapteams_matches(matches, event.remap_teams)
+
+    # Merge with existing matches so delete_invalid_matches can see the full
+    # set and identify unplayed elim matches whose set has already been
+    # decided (e.g. an f1m3 placeholder when finals ended 2-0).
+    new_match_keys = {m.key.id() for m in matches}
+    existing_matches = Match.query(Match.event == event.key).fetch()
+    for existing_match in existing_matches:
+        if existing_match.key.id() not in new_match_keys:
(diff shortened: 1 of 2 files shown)


src/backend/api/handlers/trusted.py
```diff
@@ -325,6 +325,19 @@ def update_event_matches(event_key: EventKey) -> Response:
 
     if event.remap_teams:
         EventRemapTeamsHelper.remapteams_matches(matches, event.remap_teams)
+
+    # Merge with existing matches so delete_invalid_matches can see the full
+    # set and identify unplayed elim matches whose set has already been
+    # decided (e.g. an f1m3 placeholder when finals ended 2-0).
+    new_match_keys = {m.key.id() for m in matches}
+    existing_matches = Match.query(Match.event == event.key).fetch()
```
phil-lopreiato: Do we want this to be `EventMatchesQuery` instead?

phil-lopreiato (approved):
otherwise, LGTM

PR #9802, opened 2026-04-28
Render nexus pitmap SVGs

phil-lopreiato:
render nexus pitmaps! This is mainly for an easy embedding in the mobile apps

There's some examples in the repo for the test cases

image

And also support highlighting teams (which eg we could use for team@event or favorites)

image


src/backend/web/handlers/event.py
```diff
@@ -490,6 +493,45 @@ def format_match_time(match_time: Optional[datetime]) -> Tuple[str, str]:
     )
 
 
+@cached_public
+def event_pitmap(event_key: EventKey) -> Response:
```
ZachOrr, line 498: nit: I'm surprised the robot isn't complaining but - we could validate the event key is in the proper format before passing it to the event query

src/backend/web/handlers/event.py
```diff
@@ -490,6 +493,45 @@ def format_match_time(match_time: Optional[datetime]) -> Tuple[str, str]:
...
+    event: Optional[Event] = event_query.EventQuery(event_key).fetch()
+    if not event:
+        abort(404)
+
+    nexus_pit_map = NexusPitMap.get_by_id(event_key)
+    if not nexus_pit_map:
+        abort(404)
+
+    highlight_team_keys = set()
+    for highlight_value in request.args.getlist("highlight"):
```
ZachOrr: nit: Maybe we make this `teams`? So the request is like `/event/{event_code}/pitmap?teams=frc7332,frc254`

src/backend/web/handlers/event.py
```diff
@@ -490,6 +493,45 @@ def format_match_time(match_time: Optional[datetime]) -> Tuple[str, str]:
...
+        for maybe_team_key in highlight_value.split(","):
+            team_key = maybe_team_key.strip().lower()
+            if re.fullmatch(r"frc\d+", team_key):
```
ZachOrr: Wild to me we don't have a helper like `Team.validate_key_name` that does this

PR #9850, opened 2026-05-06
PWA: Add a hover tooltip with team name and avatar to team number links

johnb8:
## Description

This adds a small tooltip on all team number links that contains the team name and avatar.

## Motivation and Context

I often forget team names and it can be annoying to have to go to the team page just to look at their name. It's mostly inspired by the super useful Wikipedia hover tooltips.

image

## How Has This Been Tested?

I think I found most of the places where there's just a plain team number link, they all seem to work. Let me know if I've missed any.

## Screenshot Pages

/event/2026bcvi Event Page
image

image

## Types of changes

Thanks for maintaining The Blue Alliance!

### DIFF
--- a/pwa/app/components/tba/teamTooltip.tsx
+++ b/pwa/app/components/tba/teamTooltip.tsx
@@ -1,11 +1,14 @@
 import { useSuspenseQuery } from '@tanstack/react-query';
 import { ComponentProps, Suspense, useMemo } from 'react';
 
+import BiTrophy from '~icons/bi/trophy';
+
 import { MediaAvatar } from '~/api/tba/read';
 import {
   getTeamMediaByYearOptions,
   getTeamOptions,
 } from '~/api/tba/read/@tanstack/react-query.gen';
+import InlineIcon from '~/components/tba/inlineIcon';
 import { TeamLink } from '~/components/tba/links';
 import TeamAvatar from '~/components/tba/teamAvatar';
 import {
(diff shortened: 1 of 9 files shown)


pwa/app/components/tba/teamTooltip.tsx
```diff
@@ -0,0 +1,115 @@
...
+  year,
+  disqualified,
+  surrogate,
+}: TeamTooltipProps) {
+  const { data: media } = useSuspenseQuery(
+    getTeamMediaByYearOptions({ path: { team_key: teamKey, year } }),
+  );
+
+  const { data: team } = useSuspenseQuery(
+    getTeamOptions({ path: { team_key: teamKey } }),
```
tervay: this will fetch for every team hovered on, which is what i would have done at first too. however, we actually keep a search index of every team available to the browser, which has team name:

Image

I bet you could instead get this from the query client and it would resolve instantly rather than sending a fetch for every single team.

should be like `getSearchIndexOptions({})` i think

tervay (changes requested):
thanks for the pr!

pwa/app/components/tba/teamTooltip.tsx
tervay: well, actually, this doesn't provide the avatar. personally i dont think the avatar is super helpful, but open to hearing what you think.

johnb8: Ah cool, just changed it to use that. I think the avatar just makes it look a bit nicer, but I agree it might not be worth fetching all the team media for the year just to get the avatar on every hover. I didn't see a way to just fetch the avatar

tervay: lets see how people respond to it without the avatar and we can come back to it in the future. not against it, but lets try it without the avatar for now.

tervay (approved):

PR #9935, opened 2026-05-24
[pwa] Adjust color scheme

tervay:
I started by finding the most popular light and dark themes in VSCode that I could steal a color palette from. I wound up using Github Light and Github Dark -- this is why the base ui colors very slightly changed (such as the background in dark mode).

I then dove rather deep into the color world and read about oklch and accessibility stuff. The alliance colors are based on the audience display colors seen [here](https://www.youtube.com/watch?v=qIXcaiI9fZk).


tervay:
replaced by #9938

PR #10016, opened 2026-06-06
[pwa] Fix team list row border clipped on hover

gregmarra:
## Summary

On the team list pages, the bottom border of a hovered row visibly disappears underneath the Name cell while remaining intact under Number and Location. Cause: the Name cell applied `truncate` directly to the `<td>`, and Tailwind's `truncate` includes `overflow-hidden`, which clips the row's `border-b` at that cell. The hover background tint is what makes the missing slice noticeable.

Moves the `truncate` onto an inner `<div>` so the `<td>` itself stops clipping. Width constraint and ellipsizing both still work.

## Screenshot Pages

- /teams/1
- /teams/2

## Test plan

- [ ] Hover any row on the team list — bottom border is continuous across all three cells.
- [ ] Long team nicknames still truncate with an ellipsis.

### DIFF
--- a/pwa/app/components/tba/teamListTable.tsx
+++ b/pwa/app/components/tba/teamListTable.tsx
@@ -34,8 +34,8 @@ export default function TeamListTable({ teams }: { teams: TeamSimple[] }) {
                 {team.team_number}
               </Link>
             </TableCell>
-            <TableCell className="w-6/12">
-              <div className="truncate">{team.nickname}</div>
+            <TableCell className="w-6/12 truncate max-w-px">
+              {team.nickname}
             </TableCell>
             <TableCell className="w-4/12">
               {team.city != null && (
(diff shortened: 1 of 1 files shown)


pwa/app/components/tba/teamListTable.tsx
```diff
@@ -34,7 +34,9 @@ export default function TeamListTable({ teams }: { teams: TeamSimple[] }) {
                 {team.team_number}
               </Link>
             </TableCell>
-            <TableCell className="w-6/12 truncate">{team.nickname}</TableCell>
+            <TableCell className="w-6/12">
+              <div className="truncate">{team.nickname}</div>
+            </TableCell>
```
ofekashery: This change still doesn't fix the issue.

Image

Please apply the following to resolve it:
```suggestion
            <TableCell className="w-6/12 truncate max-w-px">{team.nickname}</TableCell>
```

Image

ofekashery (changes requested):

Issue #10102, opened 2026-06-28
2026 Einstein matches shown as qualification matches; champs teams on Einstein before divisions


gregmarra:
Reported on Chief Delphi: [#89](https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/89) (@dan) and [#70](https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/70) / [#72](https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/72).

## Problem

Two related Championship/Einstein modeling issues:

1. **Einstein matches are shown as qualification matches.** On the 2026 Einstein Field event, the Einstein round-robin matches appear under qualification matches. (@dan, [#89], with screenshot.)
2. **Before divisions are released, every team registered for the Championship is shown as competing on Einstein** (the `2026cmptx` Einstein Field event) rather than at the Championship generally. (@TAF_International, [#70]; @bdaroz noted all teams sit on the Einstein Field event until split into divisions, [#71].)

@plnyyanks acknowledged the second one ([#72]): _"I do think there's a bug somewhere, since we shouldn't flip the name to Einstein until after divisions are released."_

## Notes

2026 champs has passed, so the live symptom may no longer be visible, but the underlying modeling bug will recur every championship — filing to track a proper fix before 2027 champs.

PR #10109, opened 2026-06-29
[data] Don't label Einstein matches as qualification matches (#10102)

gregmarra:
Fixes the Einstein-matches-shown-as-qualification half of #10102 (see Notes for the rest).

## Problem
On the 2026 Einstein Field event (`2026cmptx`), the Einstein matches were shown under **Qualification** instead of the playoff bracket ([CD #89](https://www.chiefdelphi.com/t/the-blue-alliance-2026-site-updates-features/515744/89)). Einstein has no qualification phase.

## Root cause (ground-truth verified)
2026 Einstein is an 8-alliance **double-elimination** bracket (not a round robin). `PlayoffTypeHelper.get_comp_level` returned `CompLevel.QM` for any match whose `tournamentLevel` is `"Qualification"` *before* considering the event. The FMS feed sometimes tags Einstein (a `CMP_FINALS` event, which has no quals) playoff matches as `"Qualification"`, so they were ingested as `qm` and rendered under "Qualification Results". Production has since re-synced to the correct `sf`/`f`, but the modeling bug recurs every championship.

## Fix
Add an optional `event_type` to `get_comp_level` and skip the `Qualification → qm` short-circuit when `event_type == CMP_FINALS`, so Einstein matches fall through to the existing playoff logic (`DOUBLE_ELIM_MAPPING → sf/f`). `event_type` defaults to `None`, preserving all existing behavior; championship **divisions** (which run real quals) are explicitly unaffected. The three callers (`fms_api_match_parser` ×2, nexus `queue_status_parser`) now pass `event.event_type_enum`.

## Tests
- Unit: `Qualification`-tagged `CMP_FINALS` matches map to the double-elim `sf`/`f` (asserted against `DOUBLE_ELIM_MAPPING`), while `REGIONAL` / `DISTRICT` / `CMP_DIVISION` quals stay `qm`.
- Parser: a 2026 `CMP_FINALS` double-elim hybrid schedule re-tagged `"Qualification"` parses to **0 qm / 13 sf / 3 f**. The legacy 2017 round-robin Einstein parse is unchanged. (22 tests pass; black + flake8 clean. Local pyre couldn't run — the dev container's `pyre.bin` is the wrong CPU arch; CI runs pyre.)

## Visual evidence (local running site)
The same 16 Einstein matches (real `2026cmptx` data), rendered with the buggy `qm` comp_level vs the fixed `sf`/`f`:

## Screenshots

| | Before | After |
|---|---|---|
| **Einstein matches on the event page (Results tab)** |  |  |

**Before** — Einstein matches under "Qualification Results" (Quals 1-16), the reported bug (seeded with the buggy-parser `qm` output). **After** — the same matches under "Playoff Results" + the double-elim "Playoff Bracket" on the real `/event/2026cmptx`.

## Notes / scope
- The **second half** of #10102 (all champ teams shown on Einstein before divisions are released / premature "Einstein" name flip) was **already fixed** in April 2026 (#9681 / #9719 / #9741 / #9742).
- Split district-championship finals fields share the same FMS mistag but are a `DISTRICT_CMP` event (which runs real quals), so they can't be blanket-skipped — called out in a code comment as a known follow-up.
- Past `CMP_FINALS` events already stored as `qm` will correct on the next re-sync of those events.

## Review
Gated on a 5-seat adversarial review panel (tech lead, PM, designer, web nerd, FRC nerd) — all five approve.

### DIFF
--- a/src/backend/common/helpers/playoff_type_helper.py
+++ b/src/backend/common/helpers/playoff_type_helper.py
@@ -1,6 +1,7 @@
-from typing import Tuple
+from typing import Optional, Tuple
 
 from backend.common.consts.comp_level import CompLevel
+from backend.common.consts.event_type import EventType
 from backend.common.consts.playoff_type import (
     BRACKET_ELIM_MAPPING,
     BRACKET_OCTO_ELIM_MAPPING,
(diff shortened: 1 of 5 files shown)


gregmarra:
Feels like we should find a cleaner way to fix this bug / maybe fix it in FMS?

PR #10150, opened 2026-07-06
Add moderation API (phase 1)

gregmarra:
Phase 1 of the moderation API: a Firebase-Bearer-authenticated `/api/moderation/v1` for reviewing the suggestion queue. **Backend only** — split per review feedback; the PWA review UI that consumes this API is #10187.

**PR stack:** #10160 (YouTube duration parsing, land first) → **this PR** (API) → #10187 (PWA UI) → #10161 (web review home "try the beta" banner, draft until the UI deploys to beta). #10178 (require offseason dates) is stacked on this branch.

## Reviewer's guide

The fastest way to review is one type end-to-end — the other seven follow the same pattern:

1. `require_moderation_permission` in `src/backend/api/handlers/decorators.py` (auth: Firebase Bearer, verified email, permission gate)
2. `SuggestionReviewer.accept_suggestion` → `_accept_match_video` in `src/backend/common/suggestions/suggestion_reviewer.py` (per-type permission re-check, XG transaction, audit log)
3. `moderation_suggestion_list` + `_add_match_metadata` in `src/backend/api/handlers/moderation.py` (serialization + review metadata)

## What's here

**API** (`src/backend/api/handlers/moderation.py`, blueprint in `moderation_api_main.py`):
- `GET /queue` — pending counts per type, filtered to the caller's permissions
- `GET /suggestions/<type>` — pending suggestions with rich review metadata (candidate media, target team/event, existing preferred images/webcasts/API keys, similar offseason events, author reputation, YouTube title/duration) plus a `total` for capped pages
- `POST /suggestions/<key>/accept` — transactional accept with type-specific overrides
- `POST /suggestions/reject` — batch reject (capped at 500)

All accept/reject logic lives in a shared `SuggestionReviewer` service ported from the 8 web review controllers (which can migrate onto it in a follow-up), with `AuditLogEntry` writes and the Trusted-API-key admin alert delivered to Slack (`suggestion-nag`) since the old admin-email path was never wired up.

**Admin seed page** (`/admin/suggestions/seed`, gated out of prod): one pending suggestion of every type with real, embeddable media so review UIs preview properly; can grant review permissions to a local test account.

## Security posture

All enforcement is server-side; clients are untrusted.

- **Transport/routing**: `/api/moderation/v1/*` is served only by the api service (dispatch.yaml); all 4 routes are wrapped in `@require_moderation_permission`.
- **Authentication**: Firebase ID token via `Authorization: Bearer` only, verified with `verify_id_token(check_revoked=True)` (signature, audience, expiry, revocation). No cookie/session fallback → no ambient credentials → CSRF-immune; `CORS *` is safe because credentials are never implicit. Tokens with unverified emails are rejected (accounts are linked to tokens by email claim).
- **Authorization (read)**: decorator requires a review AccountPermission (or the server-set `admin` custom claim); queue counts and lists are filtered/403'd per-type.
- **Authorization (write)**: `SuggestionReviewer` independently re-checks the per-type permission derived from the stored suggestion's `target_model` (never client input), per suggestion, inside the service layer — the write path does not trust the route gate.
- **Integrity**: accept/reject run in XG transactions with a REVIEW_PENDING re-check (no double-review); every decision writes an AuditLogEntry.
- **Input**: type-specific override allowlists (no mass assignment), enum validation, list limit clamped to 1–100, reject batches capped at 500.
- **Seeding**: the admin seed page 404s in production (`Environment.is_prod()`), even for GAE admins.
- **Tests**: 401 on all endpoints, 403 for missing/wrong/unverified auth on read and write paths, replay conflict, and cap behavior.

## Deployment notes

- Match-video title/duration enrichment uses the existing `GoogleApiSecret` sitevar (`videos.list`, 1 quota unit per 50 videos); when unset (dev), the fields are simply absent. Duration requires #10160.
- Author reputation counts are equality-only queries on existing indexed properties (`author` + `review_state`, same shape as the account page) — no new composite index — and are memcached for an hour.

## Test plan

- 54 moderation API tests, 7 seed tests; `pyre`, `black`, `flake8` clean
- Full e2e against the local dev stack (docker datastore + Firebase auth emulator) driving the PWA UI (#10187) through seed → sign-in → grant → review for all 8 types

### DIFF
--- a/src/backend/api/handlers/decorators.py
+++ b/src/backend/api/handlers/decorators.py
@@ -3,11 +3,13 @@ import logging
 from functools import wraps
 from typing import Callable, Type, TypeVar
 
-from flask import g, jsonify, request, Response
+from flask import g, jsonify, make_response, request, Response
 
+from backend.api.client_api_auth_helper import ClientApiAuthHelper
 from backend.api.client_api_types import VoidRequest
 from backend.api.trusted_api_auth_helper import TrustedApiAuthHelper
 from backend.common.auth import current_user
+from backend.common.consts.account_permission import AccountPermission
 from backend.common.consts.auth_type import AuthType
 from backend.common.consts.event_code_exceptions import EventCodeExceptions
 from backend.common.consts.fms_report_type import FMSReportType
@@ -91,6 +93,62 @@ def require_write_auth(auth_types: set[AuthType] | None, file_param: str | None 
                 TrustedApiAuthHelper.do_trusted_api_auth(
                     event_key, fms_report_type, auth_types, file_param
                 )
+            return func(*args, **kwargs)
+
+        return decorated_function
+
+    return decorator
+
+
+def require_moderation_permission(permissions: set[AccountPermission]):
+    """
+    Authenticate the request via a Firebase ID token in the Authorization
+    header (the same mechanism as the client API) and require the resolved
+    account to hold any of the given AccountPermissions. Admins always pass.
+    The resolved User is stored on flask.g.moderation_user.
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def decorated_function(*args, **kwargs):
+            with Span("require_moderation_permission"):
+                user = ClientApiAuthHelper.get_current_user()
+                if user is None:
+                    return make_response(
+                        jsonify(
+                            {
+                                "Error": "Authorization required. Pass a Firebase ID token "
+                                "as 'Authorization: Bearer <token>'."
+                            }
+                        ),
+                        401,
+                    )
+                # Accounts are linked to tokens by email claim, so an
+                # unverified email must never confer the linked account's
+                # permissions
+                if not user.email_verified:
+                    return make_response(
+                        jsonify(
+                            {
+                                "Error": "Moderation requires a verified email on your "
+                                "sign-in provider."
+                            }
+                        ),
+                        403,
+                    )
+                if not user.is_admin and not permissions.intersection(
+                    user.permissions or []
+                ):
+                    return make_response(
+                        jsonify(
+                            {
+                                "Error": "You do not have permission to moderate suggestions. "
+                                "If this is incorrect, please contact TBA admins."
+                            }
+                        ),
+                        403,
+                    )
+                g.moderation_user = user
             return func(*args, **kwargs)
 
         return decorated_function
--- a/src/backend/api/handlers/moderation.py
+++ b/src/backend/api/handlers/moderation.py
@@ -1,0 +1,662 @@
+import datetime
+import json
+import logging
+import re
+from difflib import SequenceMatcher
+from typing import Any, Dict, List, Optional
+
+from flask import g, jsonify, make_response, request, Response
+from google.appengine.ext import ndb
+from pyre_extensions import none_throws
+
+from backend.api.handlers.decorators import require_moderation_permission
+from backend.common.consts.account_permission import SUGGESTION_PERMISSIONS
+from backend.common.consts.auth_type import WRITE_TYPE_NAMES
+from backend.common.consts.event_type import EventType
+from backend.common.consts.media_type import IMAGE_TYPES
+from backend.common.consts.suggestion_state import SuggestionState
+from backend.common.consts.suggestion_type import SuggestionType, TYPE_NAMES
+from backend.common.datafeeds.datafeed_youtube import YoutubeVideoDetailsDatafeed
+from backend.common.helpers.outgoing_notification_helper import (
+    OutgoingNotificationHelper,
+)
+from backend.common.helpers.suggestion_fetcher import SuggestionFetcher
+from backend.common.memcache import MemcacheClient
+from backend.common.models.api_auth_access import ApiAuthAccess
+from backend.common.models.district import District
+from backend.common.models.event import Event
+from backend.common.models.media import Media
+from backend.common.models.suggestion import Suggestion
+from backend.common.models.user import User
+from backend.common.queries.event_query import EventListQuery
+from backend.common.sitevars.google_api_secret import GoogleApiSecret
+from backend.common.sitevars.slack_hook_urls import SlackHookUrls
+from backend.common.suggestions.suggestion_reviewer import (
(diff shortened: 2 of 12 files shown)


pwa/generate-api.sh
```diff
@@ -2,6 +2,21 @@
 npx @hey-api/openapi-ts -f openapi-ts/read.config.ts
 npx @hey-api/openapi-ts -f openapi-ts/mobile.config.ts
 npx @hey-api/openapi-ts -f openapi-ts/nexus.config.ts
+npx @hey-api/openapi-ts -f openapi-ts/moderation.config.ts
```
tervay: would prefer this be a separate PR for reviewability's sake

pwa/generate-api.sh
```diff
@@ -2,6 +2,21 @@
...
+
+# Define a function to run sed in Docker
+docker_sed() {
+  docker run --rm -v "$PWD":/work -w /work alpine sed -i "$@"
```
tervay: we used to do this to get award/event types as enums but that shouldn't be necessary anymore

tervay (changes requested):

pwa/generate-api.sh
gregmarra: The API separately from the PWA code that uses it?

pwa/generate-api.sh
gregmarra: Yep — these were accidentally resurrected from an old version of the script during a rebase conflict, and they actually break regeneration now that the spec exports those enums directly (the prepended imports collide with the enums the generated file now exports itself). Already removed on the current head — the `[pwa] Check API regeneration` job caught it, and the script is now main's version plus the moderation config line.

pwa/generate-api.sh
tervay: Yeah

gregmarra: Done — split into this PR (backend API only, now rebased onto current main) and #10187 (the PWA review UI + generated client). The stacked #10178 was rebased onto the new head too.

gregmarra:
## Rebased onto `main` + adversarial review

**Rebase:** onto `d5503e833` (282 commits behind). Both commits replayed cleanly with no conflicts. Backend-only diff — the pre-rebase `[pwa]` CI failures were drift, not this PR. **27/27 checks green** at `9069c841f`. Backup of the pre-rebase head kept locally as `backup/moderation-api-20260905`.

### Checked and cleared

- **Per-type authorization on every write path.** `accept_suggestion` and `reject_suggestions` both call `user_can_review(user, SuggestionType(suggestion.target_model))` and return `FORBIDDEN` *before* entering the transaction — the decorator's "any suggestion permission" check is only the outer gate. Covered by `test_accept_requires_type_permission`, `test_reject_requires_type_permission`, `test_list_requires_type_permission`.
- **Email linking.** `User` resolves the Account by email (`AccountQuery(email=…)`), and the decorator refuses tokens without `email_verified`. Right mitigation. It still leans on the Firebase project's *one account per email* setting to stop a second provider asserting a verified matching email — worth a one-line confirmation that setting is on, since it's config, not code.
- Race on double-review handled in-transaction (`ALREADY_REVIEWED` → 409, tested). Audit log written on both accept and reject. Batch reject capped at 500. Seed endpoint aborts 404 in prod on both GET and POST and sits under admin routes. CORS `*` with bearer auth is fine (no ambient credentials).

### Findings

**1. The default expiration for a past event is "never" here, but "7 days" in the web UI.** ~~*(policy divergence — please decide)*~~

> ✅ **Resolved in `7d26d5c0b`.** Greg chose the 7-day `max(after_event, after_grant)` default everywhere — never-expiring keys should be a rarity, not a default. The `-1` fallback is removed, `test_accept_apiwrite_default_expiration_past_event` now asserts 7 days from the grant, and the OpenAPI description describes the whichever-is-later rule. Analysis kept below for context.

`_accept_api_write` (`suggestion_reviewer.py:443-454`): when `expiration_days` is omitted, default to 7 only if `event.end_date + 8d` is still in the future, **otherwise `-1` = never expires**. Pinned by `test_accept_apiwrite_default_expiration_past_event` (`assert auth.expiration is None`). The PWA mirrors this exactly in `defaultExpirationDays` and only sends `expiration_days` when the moderator changes the dropdown, so an untouched accept of a stale request via the PWA/API issues a **permanent key**.

The web review page (`suggest_apiwrite_review_list.html`) defaults to `7` for every event. Same suggestion, two surfaces, two outcomes — and the newer surface picks the more permissive one.

The `-1` fallback looks like it exists to avoid handing out a dead-on-arrival key for a past event. But it isn't needed for that: the same function's `max(event_end + N + 1, now + N)` at `:456-463` already guarantees a late-granted 7-day key is valid for 7 days from grant. So the permanent-key default is solving a problem the code already solves, at the cost of least-privilege. **Recommend defaulting to 7 unconditionally** (matching the web UI), or requiring `expiration_days` explicitly for `api_auth_access`. Either way, update that test and the client helper together.

**2. Expiration semantics now live in three places.** `suggestion_reviewer._accept_api_write`, the web controller (being consolidated into `compute_expiration()` in #10520), and the client's `defaultExpirationDays`. They already disagree (finding 1). `backend.common` can't import `backend.web` (ETBA0 lint rule), so the shared helper has to live in `common/` — I'll move `compute_expiration` there on the #10520 side if you want this reviewer to call it; that would make #10150 depend on #10520 landing first, so your call on sequencing.

**3. Batch reject re-fetches every suggestion for alerting.** `moderation_suggestions_reject` calls `_send_apiwrite_review_alert` per rejected outcome, and that helper does `Suggestion.get_by_id` just to read `target_model`. `reject_suggestions` already loaded each entity. Up to 500 extra sequential gets per request; returning `target_model` on `ReviewOutcome` removes them.

**4. Minor.**
- `moderation_suggestion_list` queries `Suggestion` directly rather than via a query class (repo convention), and the images-first sort runs *after* `fetch(limit=…)`, so it only reorders the fetched page.
- `decorators.py` is also touched by #10489 (removes the auth `logging.info`). Different hunks, auto-merges either order — just noting the overlap.

tervay (approved):

PR #10164, opened 2026-07-07
Add profile avatar menu to the PWA navbar

gregmarra:
Adds a profile avatar to the top-right of the navbar (Google account photo via Firebase `photoURL` with `referrerPolicy="no-referrer"`, initials fallback, generic icon when signed out) opening a dropdown with:

- **Account header**: TBA display name + email (Firebase identity, same as the account page)
- **Account** link (myTBA is already in the navbar, so it isn't duplicated here)
- **Theme selector** — light/dark/system radio, moved here from the footer (the footer toggle is removed; this also exposes the previously unreachable "system" option)
- **Data source (dev only)**: an `import.meta.env.DEV`-gated radio to point the read API client at `http://localhost:8080/api/v3` instead of production, persisted in localStorage and applied at client startup (selecting reloads the page). Local mode also swaps `X-TBA-Auth-Key` to the dev stack's well-known `tba-dev-key` (seeded by `/local/bootstrap`), since the prod read key isn't valid against the local datastore. Production builds never render any of this and always read prod.
- **Logout** (matching the account page's wording) behind a shared confirmation dialog, also used by the account page's Logout button now

Built on the Base UI menu primitives from #10149.

Menu items use the popover foreground color (not hyperlink blue) and the shared dropdown-menu component now uses `cursor-pointer` for all interactive items (previously `cursor-default` per the shadcn desktop-menu convention — this also fixes the year selector's cursor). The account page and the menu now share `getDisplayName` so both fall back identically for accounts without a display name.

Adding the menu to the right-hand navbar group squeezed that row enough to wrap the "Leave beta" link onto two lines at some widths, so it now carries `whitespace-nowrap` like the wordmark beside it.

Rebased on main. Since this branch was opened, main moved the read client's cached fetch into an SSR-only guard ("do not install a second TTL in the browser"), so the dev-only `baseUrl` override is now applied as its own `client.setConfig` call rather than being folded into that block — `setConfig` merges, and `isLocalDataSource()` is false without a browser, so SSR still always reads prod.

## Screenshots

| Signed in (light) | Signed in (dark) | Signed out |
|---|---|---|
| light | dark | signed out |

| Logout confirmation |
|---|
| confirm |

## Test plan

- vitest: data-source persistence (default prod, round-trip, garbage values, SSR-safe) and initials fallback chain (name → email → ?)
- Headless Chrome against the dev stack: signed-out menu (sign-in link + theme radios), signed-in menu (email, Account, myTBA, sign out), dark mode applied via the menu, data-source toggle verified end-to-end (read API requests observed hitting `localhost:8080/api/v3/...` after toggling), footer toggle gone
- `tsc`, `eslint`, `prettier` clean; 333 vitest tests green
- Playwright `index.spec.ts` + `theme.spec.ts` green (12 specs, chromium + Mobile Chrome), covering the navbar profile-menu trigger and the menu's light/dark/system radio group

## Screenshot Pages

- / Homepage with avatar menu

### DIFF
--- a/pwa/app/components/tba/auth/logoutConfirmDialog.tsx
+++ b/pwa/app/components/tba/auth/logoutConfirmDialog.tsx
@@ -1,0 +1,49 @@
+import { type JSX } from 'react';
+
+import { Button } from '~/components/ui/button';
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from '~/components/ui/dialog';
+
+/** Shared logout confirmation used by the navbar menu and the account page. */
+export function LogoutConfirmDialog({
+  open,
+  onOpenChange,
+  onConfirm,
+}: {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+  onConfirm: () => void;
+}): JSX.Element {
+  return (
+    <Dialog open={open} onOpenChange={onOpenChange}>
+      <DialogContent className="max-w-sm">
+        <DialogHeader>
+          <DialogTitle>Logout of The Blue Alliance?</DialogTitle>
+          <DialogDescription>
+            Your myTBA favorites and settings will be here when you sign back
+            in.
+          </DialogDescription>
+        </DialogHeader>
+        <DialogFooter>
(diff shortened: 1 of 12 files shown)


gregmarra:
Needs some UI work

tervay:
Overall in favor of this direction

PR #10187, opened 2026-07-07
Add PWA suggestion review UI (phase 2)

gregmarra:
Phase 2 of the moderation work: the moderator-facing UI for the phase-1 API, split out of #10150 per review feedback (this PR: `pwa/` only; #10150: the backend API). Compiles and tests independently — the generated client is checked in — but the review pages need #10150's API deployed to actually function.

**PWA** (`/suggest/review`): per-type queues in the web dashboard's order, bulk select + single submit, review guidelines carried over from the Jinja templates, webcast suggestions grouped per event with existing-webcast context, and:
- Social profile validation (facebook profile.php/group links, YouTube video-as-channel, malformed handles) with auto-selected Reject; dead-imgur detection likewise
- Facebook page-plugin, GitHub avatar-card, and Instagram post embeds
- Author reputation on every card ("137 accepted · 93 rejected" / "first-time suggester")
- YouTube title/duration on match videos with amber warnings for title-mismatch and long-stream-without-timestamp uploads (needs #10160)
- Keyboard shortcuts: j/k navigate (glowing focus ring), a/r accept/reject, p toggle preferred, ? shows the shortcut dialog
- Per-outcome submit toasts ("5 accepted · 3 rejected · 1 already reviewed by someone else", "Queue cleared! 🎉")

Also updated for the Base UI migration (#10149), which landed after the original branch: `asChild` usages became `render` props, and the new breadcrumb component uses `useRender` instead of Radix `Slot`.

## Old vs. new, per review flow

Same seeded dev datastore, both sides signed in as a moderator, captured in headless Chrome at 1280px.

| Flow | Old (Jinja) | New (PWA) |
|---|---|---|
| Match Videos | old | new |
| Webcasts | old | new |
| Team Media | old | new |
| Social Media | old | new |
| Event Videos | old | new |
| CAD Models | old | new |
| Offseason Events | old | new |
| API Key Requests | old | new |

More detail shots (embeds, warnings, keyboard dialog, mobile) in [shots/moderation-api](https://github.com/the-blue-alliance/the-blue-alliance/tree/screenshot-assets/shots/moderation-api).

## Test plan

- 350 vitest tests (19 for moderation utils: validation, warnings, ordering, defaults); `tsc`, `eslint`, `prettier` clean against current main (post-Base-UI)
- Full e2e against the local dev stack driving headless Chrome through seed → sign-in → grant → review for all 8 types (`pwa/scripts/moderation-e2e.mjs`); mobile pass at 390px


gregmarra:
## Rebased onto `main` + adversarial review

**Rebase:** onto `d5503e833` (282 commits behind), single commit replayed. Four conflicts, resolved as follows — please sanity-check the first two since they're judgment calls:

| file | resolution |
|---|---|
| `pwa/.env.example` | Kept only `VITE_TBA_MODERATION_API_BASE_URL`. **Dropped the `NEXUS_*` lines** the branch carried from its old base — `main` no longer references those env vars anywhere in `pwa/`. |
| `pwa/generate-api.sh` | Kept `main`'s `colors.config.ts` line, added `moderation.config.ts`, **dropped `nexus.config.ts`** (that file doesn't exist on `main`; the line would fail). Added `app/api/tba/moderation` to the prettier step. |
| `pwa/app/routes/__root.tsx` | Kept both import blocks (`main`'s `getSearchIndexOptions`/`getStatusOptions` and the branch's `moderationClient`), sorted. |
| `pwa/app/routeTree.gen.ts` | Took `main`'s, then **regenerated via `pnpm run build:app`** rather than hand-merging. +43 lines, both `suggest/review` routes present. |

**Then fixed 4 lint errors** that would have failed `[pwa] Format & Lint` — all new Tailwind v4 rules on `main` hitting the branch's older markup, not logic changes:

- `break-words` → `wrap-break-word` (`reviewGuidelines.tsx:119`, `ui/breadcrumb.tsx:17`)
- `hover:text-destructive-foreground` → `hover:text-white` (`suggest.review.$suggestionType.tsx:87`, `:360`) — the `destructive-foreground` token was removed on `main`; this matches `main`'s own `Button` destructive variant.

Verification: `pnpm lint` 0 errors, `prettier --check` clean, `tsc` clean, vitest 40/40 (`moderationUtils`, `reviewGuidelines`). Pushed as `192ee4d9e`; pre-rebase head kept locally as `backup/moderation-pwa-20260905`.

### Checked and cleared

- **Keyboard shortcuts** (`suggest.review.$suggestionType.tsx:148-213`): ignore events with modifiers or originating in `input, textarea, select, [contenteditable]`, and `a`/`r` only *toggle a pending decision* — nothing submits from a keystroke. Good.
- **Links to user-submitted URLs** (`ExternalLink`, GitHub profile card): both `target="_blank"` sites carry `rel="noreferrer"`. Social profile URLs are rebuilt from the parsed foreign key, not echoed raw.
- **Auth**: fresh `getIdToken()` per request, `throwOnError` surfaces failures to toasts. `scripts/moderation-e2e.mjs` uses only emulator/fake credentials.
- `breadcrumb.tsx` is net-new (nothing on `main`), single importer.

### One finding — shared with #10150

> ✅ **Resolved in `ac63d1131`.** `defaultExpirationDays` is replaced by `DEFAULT_EXPIRATION_DAYS = 7`; the PWA now always sends the displayed `expiration_days` (same pattern as `set_preferred`); option labels are by duration with a one-line note on the whichever-is-later rule, and "Never expires" moves last; the moderation client is regenerated from the updated spec (one-line doc change). Original note kept below for context.

`defaultExpirationDays` (`moderationUtils.ts:305-312`) mirrors the server's rule and is well-tested, so there's **no display/behaviour mismatch**. But the rule itself defaults a **past event to "Never expires"**, whereas the web review page defaults to 7 days — see the write-up on #10150. Whatever policy you pick there, consider having the PWA **always send the displayed `expiration_days`** (`suggestionReviewCard.tsx:993-1004` currently omits it until the moderator touches the dropdown). That makes the client default presentational only and removes a second implementation of the policy from the payload path.

pwa/app/lib/moderationUtils.ts
```diff
@@ -0,0 +1,311 @@
+import { Temporal } from 'temporal-polyfill';
+
+// Queue display order, matching the web review home
+// (suggestions/pending_suggestion_rows_partial.html)
+export const SUGGESTION_TYPE_ORDER = [
```
tervay: could we make this an enum that's derived from the api? i think you can use like x-varnames or something in the openapi spec to generate them in TS. it's done somewhere in this repo

tervay (changes requested):

pwa/app/lib/moderationUtils.ts
gregmarra: Done — good call. Derived from the spec now, with one deliberate split:

**Membership comes from the API.** `SuggestionType` in `api_moderation_v1.json` now carries `x-enum-varnames` mirroring `suggestion_type.py` (that half is on #10150, `0db9fd535`), and `moderation.config.ts` turns on hey-api's `enums: 'typescript'` — the same setting the read client uses. The regenerated client exposes `SuggestionType` and `ReviewResult` as real enums, and every wire-string literal in the moderation UI now uses a member (`TYPE_NAMES` keys, `TEAM_HEADER_TYPES`, the type and result comparisons, the queue page's media-tools check). `a601d14a9`.

**Order stays a client constant, on purpose.** `SUGGESTION_TYPE_ORDER` was doing two jobs — *which* types exist and *what order* the queue displays them — and only the first belongs to the API. Deriving display order from the spec's array order would let a tidy-up of the JSON silently reorder the review queue. So it's now `readonly SuggestionType[]` built from members, plus a test that it's a permutation of `Object.values(SuggestionType)` — a type added to the API can't be forgotten here, and a typo can't compile.

One thing to know for merge order: #10187's regenerated client comes from the spec on #10150, so #10150 lands first (it has to anyway — the UI needs the API).

tervay (approved):

Issue #10248, opened 2026-07-16
Add `@tanstack/eslint-plugin-query` and `eslint-plugin-router`


tervay:
## Problem

Neither [`@tanstack/eslint-plugin-query`](https://tanstack.com/query/latest/docs/eslint/eslint-plugin-query) nor [`@tanstack/eslint-plugin-router`](https://tanstack.com/router/latest/docs/eslint/eslint-plugin-router) is installed, despite the PWA having ~94 query call sites.

The query plugin's `exhaustive-deps` rule catches exactly the class of bug present at `app/routes/match_suggestion.tsx:649-707`:

```ts
queryKey: ['eventMatches', events],   // `events` is the full loader array
queryFn: () => Promise.all(events.map(...)),
```

The entire loader array is embedded in the key — it re-serializes on every render and busts on any identity or content change. The `queryFn` also hides an N-request fan-out inside a single query, so there's no per-event caching, dedup, or error isolation.

That same block contains a raw `fetch('https://api.statbotics.io/v3/team_years?year=2026')` with a **hard-coded year**.

## Also: `await-thenable` is disabled

`eslint.config.js` explicitly turns off `@typescript-eslint/await-thenable` (alongside `restrict-template-expressions` and `no-throw-literal`). That's what lets `await` on a non-promise slip through silently — which matters a lot in a codebase this dependent on getting `await` placement right (see the loader-styles and streaming issues, both of which are fundamentally about where `await` goes).

## Why this is high-leverage DX

Given the `staleTime` work and the Style-B loader conversions coming, **having the linter enforce query correctness before those refactors land** means it guards the change rather than cleaning up after it. Suggest doing this early in the sequence.

## Note

`app/api/tba/` and `app/api/nexus/` are already correctly ignored by the linter, so generated code won't trip the new rules.

PR #10257, opened 2026-07-17
Add tanstack eslint plugins

tervay:
Adds tanstack specific eslint rules for Router and Query. Fixes lint problems as well


tervay:
Closes #10248

Issue #2618, opened 2019-11-23
Instagram media thumbnail broken


MC42:
As time goes on, certain images are returning empty thumbnails for TBA Media for robots submitted via Instagram.

## Expected Behavior
Correct functionality would mean the image loads correctly every time.  
## Current Behavior
The `thumbnail_url` attribute in TBA stores a URL which expires, returning both a 403 status code and a response of `URL signature expired`.

## Possible Solution
I believe we need a GAE job to check our media submissions done more than some arbitrary window (usually 6m) and generate a new direct URL based on the submission key stored (as `direct_url`?)

## Steps to Reproduce (for bugs)
Visit any page on TBA using instagram (ie frc195's 2019 media).

## Your Environment
Prod, Firefox Latest, Ubuntu 16.04LTS

fangeugene:
They look fine to me?

MC42:
![image](https://user-images.githubusercontent.com/6531081/69473167-ef4d0a80-0d80-11ea-9ca3-8db14d28a836.png)

![image](https://user-images.githubusercontent.com/6531081/69473173-f6741880-0d80-11ea-8916-b3902d252ddc.png)

ZachOrr:
I thought we fixed this via https://github.com/the-blue-alliance/the-blue-alliance/pull/2319 (I did a deep dive on what was wrong with this a while back, comments on https://github.com/the-blue-alliance/the-blue-alliance/issues/2297)

bdaroz:
This is still semi broken on an irregular basis, removing stale tag.

bdaroz:
Current broken image from an event team list:

<img width="650" height="133" alt="Image" src="https://github.com/user-attachments/assets/50b22582-eb59-46ff-88d4-a7000568219f" />

When it does load:

<img width="606" height="128" alt="Image" src="https://github.com/user-attachments/assets/a177980c-aba2-4ac4-882b-5d0734ea28be" />

<img width="650" height="130" alt="Image" src="https://github.com/user-attachments/assets/08abacf3-2e50-4ada-9024-fcf3bf2ed61d" />

<img width="1044" height="760" alt="Image" src="https://github.com/user-attachments/assets/088682ab-1507-4833-8db3-ac50d7f94601" />

At a glance a page refresh _often_ but not _always_ seems to temporarily correct the loading issue and we get an embed of some kind. It appears off-sized in both the thumbnail layout and on the py3 team page, extending well below the area border.

PR #10376, opened 2026-08-08
Fall back to non-preferred team images

### LINKED ISSUE #5159: Preferred Image/Media Fallback
If a team does not have any preferred images for that year, but they do have images that aren't preferred, use those until an image is marked preferred. This way a team will at least have images added to the team tab on the event page until their preferred image is selected

Guflly:
## Description
Uses preferred team images in event team lists when available, then falls back to that year's other images. The same selection is used for the team CSV image URL.

## Motivation and Context
Teams with submitted images currently show the add-media button until a moderator marks an image preferred.

Fixes #5159.

## How Has This Been Tested?
- `uv run --group dev pytest src/backend/common/helpers/tests/media_helper_test.py src/backend/common/queries/tests/event_teams_medias_query_test.py src/backend/common/queries/tests/event_teams_preferred_medias_query_test.py src/backend/web/handlers/tests/event_detail_test.py -q` (36 passed)
- `uv run --group lint flake8 src/backend/common/helpers/media_helper.py src/backend/common/helpers/tests/media_helper_test.py src/backend/web/handlers/event.py src/backend/web/handlers/tests/event_detail_test.py`
- `uv run --group dev black --check --fast src/backend/common/helpers/media_helper.py src/backend/common/helpers/tests/media_helper_test.py src/backend/web/handlers/event.py src/backend/web/handlers/tests/event_detail_test.py`

N/A

## Screenshot Pages
N/A

## Types of changes

### DIFF
--- a/src/backend/common/helpers/media_helper.py
+++ b/src/backend/common/helpers/media_helper.py
@@ -1,5 +1,7 @@
 from typing import Dict, List, Optional
 
+from google.appengine.ext import ndb
+
 from backend.common.consts.media_type import (
     IMAGE_TYPES,
     MediaType,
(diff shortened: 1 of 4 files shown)


bdaroz:
Something I noticed in the months since un-stale-ing the linked issue -- If we mark Instagram images as preferred, they often get shown as a shaded white box with a "View on Instagram" text in rotation  -- the user has to click into instagram to see the image. Sometimes they load, sometimes they don't.

Eg: 
image

So now to avoid that, if the images don't load off IG, I may click into moderate them, but I don't promote them on purpose. I don't know if this is an IG thing, an us thing, or where the problem lies. (Nevermind the py3 page layout gets a bit broken for an IG image when it does load)

We may want to look at promoting non-IG images here as a stop-gap, or looking deeper into the IG display issue. (#2618)

PR #10406, opened 2026-08-16
Add PWA docker files for regenerating PWA models

ZachOrr:
I don't have npm/javascript package things locally, so I'd like to be able to regenerate the PWA API clients without needing all the tooling installed on my machine.

Adds a `pwa-tools` container (mise + the node/pnpm versions pinned in `pwa/mise.toml`) and one Make target:

- `make pwa-generate-api` — runs `pwa/generate-api.sh` in Docker

`node_modules`, the pnpm store, and the mise toolchain are named volumes so the container's Linux installs stay off the host bind mount.


ops/dev/pwa/Dockerfile.pwa
```diff
@@ -0,0 +1,17 @@
+# Versions mirror pwa/mise.toml, enforced by ops/pwa/check_node_versions.sh
+FROM node:24.18.1-bookworm-slim
```
tervay: I'd rather it just installs mise into a linux image and installs it via mise, rather than having Yet Another Node String to update

ops/dev/pwa/Dockerfile.pwa
```diff
@@ -0,0 +1,17 @@
...
+
+ENV PNPM_VERSION=11.18.0
+ENV PNPM_HOME=/pnpm
+ENV PNPM_STORE_DIR=/pnpm/store
+ENV PATH=$PNPM_HOME:$PATH
+
+ENV PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
+ENV PUPPETEER_SKIP_DOWNLOAD=true
```
tervay: we don't use puppeteer on pwa

tervay (changes requested):

tervay (approved):

PR #10454, opened 2026-08-28
Add match suggestion firebase feed

tervay:
Adds a firebase data feed to the backend for match suggestions. Uses handwaved constants for now. Can adjust over time.

Not connected to any frontend yet.

Adds a local route for debugging match suggestion scores per event. Nice for testing. Can remove later.


src/backend/common/helpers/match_suggestion_helper.py
```diff
@@ -0,0 +1,444 @@
...
+# PAST tau keeps the former hot while the latter fades, and the PAST horizon
+# drops anything that overdue from scoring at all.
+TIME_DECAY_TAU_FUTURE_S: int = 15 * 60
+TIME_DECAY_TAU_PAST_S: int = 5 * 60
+TIME_HORIZON_FUTURE_S: int = 3 * 60 * 60
+TIME_HORIZON_PAST_S: int = 30 * 60
+
+FAVORITE_CACHE_TTL_S: int = 24 * 60 * 60
+# Shared with TeamHelper.getPopularTeamsEvents
+FAVORITE_CACHE_NAMESPACE: str = "team-favorite-counts"
```
phil-lopreiato: we should refactor this into something inheriting from `MemcacheModel` so we can share it

src/backend/common/helpers/match_suggestion_helper.py
```diff
@@ -0,0 +1,444 @@
...
+
+DEGENERATE_NORMALIZED_VALUE: float = 0.5
+EPSILON: float = 1e-9
+
+
+def compute_match_suggestions(
```
phil-lopreiato: Put functions in a helper class, we want to avoid model-level functions

src/backend/common/helpers/firebase_pusher.py
```diff
@@ -299,6 +300,20 @@ def update_live_events(
...
 
+    @classmethod
+    def update_match_suggestions(cls, suggestions: MatchSuggestions) -> None:
+        """
+        Replaces the GameDay match suggestion feed
+        """
+        defer_safe(
+            cls._put_data,
+            "match_suggestions",
+            suggestions.model_dump(mode="json"),
```
phil-lopreiato: instead of putting the raw object here, we should have a separate model that has shorter keys for use in firebase, because we get billed per byte (which is why the other stuff in firebase tends to be annoyingly terse)

src/backend/tasks_io/handlers/match_suggestions.py
```diff
@@ -0,0 +1,18 @@
...
+
+from backend.common.helpers.firebase_pusher import FirebasePusher
+from backend.common.helpers.match_suggestion_helper import compute_match_suggestions
+
+blueprint = Blueprint("match_suggestions", __name__)
+
+
+@blueprint.route("/tasks/do/update_match_suggestions")
+def update_match_suggestions() -> Response:
+    suggestions = compute_match_suggestions()
```
phil-lopreiato: do we `put` these to the ndb anywhere?

phil-lopreiato (changes requested):

src/backend/tasks_io/handlers/match_suggestions.py
tervay: nope, firebase is the only consumer. open to changing it, but not sure where i'd put them.

phil-lopreiato (approved):
LGTM

PR #10504, opened 2026-09-04
port hof parser py3

tervay:
The current HOF page is missing videos, essays, and more, because the original HOF parser never got ported to py3.


phil-lopreiato (approved):
Seems legit

PR #10510, opened 2026-09-04
Add smugmug support

tervay:
Supports both smugmug images and smugmug albums. An image can belong to a team. An album can belong to a team or an event (not both). 

firefox_k6mb8GhEaH
firefox_0Rq8iTxB9O
firefox_9FrMEQKAZe
firefox_c7T6vdcPsy
firefox_II69AuyBYO
firefox_yes69KpOLo

### DIFF
--- a/pwa/app/api/tba/read/index.ts
+++ b/pwa/app/api/tba/read/index.ts
@@ -625,6 +625,8 @@ export {
   type MediaGrabCad,
   type MediaNoDetails,
   type MediaOnshape,
+  type MediaSmugmugAlbum,
+  type MediaSmugmugPhoto,
   type MediaTag,
   MobilityRobot2023,
   type NexusEventInfo,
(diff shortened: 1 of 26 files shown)


gregmarra:
Is smugmug being used by teams to host public media?

bdaroz:
There are a *very few* linked to on CD. Apparently they were in the vKoP this year?

*Edited:*

Actually there were two: This one (https://studio1444.smugmug.com/Events/FRC/2023-LA-Regionals) and another from a team in their build thread that's now behind a password prompt.

tervay:
It is how all event photos in NE are hosted, so I wanted to add it to event media. I've also gotten specific asks for it from Open Alliance teams, so I figured I'd also add it to team profiles. Some of them specifically want to post to Smugmug but only will if it has TBA support.

There are a lot of event photos here, many of which are on smugmug: https://frc.tools/photos

phil-lopreiato (approved):

PR #10524, opened 2026-09-05
Point moderators at the PWA suggestion review queue

gregmarra:
## What

Suggestion review now lives in the PWA — `/suggest/review`, shipped in #10187 on the moderation API from #10150. This adds a banner to the top of **every Jinja review page** (the review home plus the eight per-type queues) linking to the equivalent PWA queue, so moderators try the new tooling instead of the pages it supersedes.

One partial, `suggestions/partials/pwa_review_banner.html`, included per page with the matching PWA route:

| Jinja page | PWA queue |
|---|---|
| `/suggest/review` | `/suggest/review` |
| apiwrite | `/suggest/review/api_auth_access` |
| designs | `/suggest/review/robot` |
| event media | `/suggest/review/event_media` |
| event webcast | `/suggest/review/event` |
| match video | `/suggest/review/match` |
| offseason event | `/suggest/review/offseason-event` |
| team media | `/suggest/review/media` |
| team social | `/suggest/review/social-media` |

The route segments are the `SuggestionType` wire values, the same ones the PWA's `$suggestionType` route validates against.

## Copy

> **Suggestion review has moved to the new web app.** Please review these in the PWA instead — same queue, with bulk accept/reject, keyboard shortcuts, and better previews: **beta.thebluealliance.com/suggest/review »**
> *These pages still work, but the PWA is where fixes and new review tooling land now. If something is missing there, please let us know.*

Bootstrap `alert alert-info`, matching the existing alerts on these pages. The Jinja pages are left fully functional — this is a nudge, not a redirect — and the copy invites reports of anything the PWA is missing, which is the feedback we want while it's the beta.

## Notes

- The `beta.thebluealliance.com` host is hard-coded; that's where the PWA is served per `dispatch.yaml`. If it ever moves to `www`, this is one string in one partial.
- Only the reviewer pages get the banner; the public submission forms don't (tested).

## Testing

New `suggest_review_pwa_banner_test.py`: banner present with the correct link on the review home and on the apiwrite queue; absent from a public submit form. Ran every suggestion/review handler test (36) so all nine includes are exercised. `make lint` and `make typecheck` clean.

### DIFF
--- a/src/backend/web/handlers/tests/suggest_review_pwa_banner_test.py
+++ b/src/backend/web/handlers/tests/suggest_review_pwa_banner_test.py
@@ -1,0 +1,47 @@
+"""The Jinja suggestion-review pages point moderators at the PWA queue.
+
+Suggestion review shipped in the PWA (/suggest/review). These pages remain
+functional, but each one now carries a banner linking to the equivalent PWA
+queue so moderators try the new tooling instead.
+"""
+
+from werkzeug.test import Client
+
+from backend.common.consts.account_permission import AccountPermission
+
+BANNER = b'data-testid="pwa-review-banner"'
+PWA = b"https://beta.thebluealliance.com"
+
+
+def _as_reviewer(login_user, *permissions: AccountPermission) -> None:
+    login_user.has_permission.return_value = True
+    login_user.permissions = list(permissions)
+
+
+def test_review_home_links_to_pwa_queue_index(login_user, web_client: Client) -> None:
+    _as_reviewer(login_user, AccountPermission.REVIEW_MEDIA)
+
+    resp = web_client.get("/suggest/review")
+
+    assert resp.status_code == 200
+    assert BANNER in resp.data
+    assert PWA + b"/suggest/review" in resp.data
+
+
+def test_apiwrite_review_links_to_its_pwa_queue(login_user, web_client: Client) -> None:
+    _as_reviewer(login_user, AccountPermission.REVIEW_APIWRITE)
+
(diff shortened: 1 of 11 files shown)


bdaroz:
FYI - the `/suggest/review` path on `beta....` currently 404s.

gregmarra:
Thanks — confirmed, and it's a real blocker rather than a propagation delay.

**Cause:** PWA deploys to prod have been skipped since #10187 merged. Its merge run failed `[pwa] Check API regeneration`, which gates `Deploy PWA Service`, so the `pwa` service is still serving the pre-#10187 version (deployed 19:11 UTC). The check fails because the `x-enum-varnames` spec change that #10187's regenerated client was built from landed on `moderation-api` *after* #10150 had already been squash-merged — `main` has the enum client but not the spec that produces it. My sequencing mistake.

**Fix:** #10525 — the missing eight-line spec commit, cherry-picked. Verified locally that regeneration produces a diff on `main` as-is and **no diff** with it applied. Once it lands, the next push deploys the PWA and this link resolves. I'll rebase this PR on top so its regeneration check goes green too — it shouldn't merge before beta actually serves the page.

gregmarra:
## Screenshots

Rendered through the Flask test client against the dev server's CSS, per the recipe in AGENTS.md. Before is `main`; after is this branch. Viewport 1280×900 at 2x. Rendered as a reviewer with `REVIEW_APIWRITE` and `REVIEW_MEDIA`.

| page | before | after |
|---|---|---|
| Review home `/suggest/review` | before | after |
| Trusted API queue `/suggest/apiwrite/review` | before | after |

PR #10444, opened 2026-08-27
Add sort order arrows for user account review Write API Keys list

cpapplefamily:
## Description
Add a sorting function to the Write API Keys listed on a users account overview page

## Motivation and Context
- Currently, the Write API Keys seem to have no logical order. Every year I add several events that I need to search the list to find the entry. This sort enables the user to sort in ascending or descending order. 
- No open issues found

## How Has This Been Tested?

I bootstrapped a few events I have hosted the past few years into a locally running instance of TBA, I then added these events to the local account using the /admin/api_auth/add interface

Windows 11
VSCODE with CODEX Pro add-on installed
Docker Desktop

compose exec tba make test  
    ==== short test summary info ====
    FAILED src/backend/common/environment/tests/environment_test.py::test_auth_emulator_host_none - AssertionError: assert 'firebase:9099' is None
    ==== 1 failed, 4490 passed, 38 skipped in 764.83s (0:12:44) ====
compose exec tba make lint  
   All done!
compose exec tba make typecheck
   no errors

## Screenshots (if appropriate):
image

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would change API specifications or require data migrations)

PR #10527, opened 2026-09-06
Make the Write API Keys "Event" header the sort control, newest first

gregmarra:
Builds on #10444 by @cpapplefamily, which gave the Write API Keys table on `/account` a sort for the first time — thank you again. Two adjustments now that the sort exists.

## Newest first by default

The row you're looking for is almost always the key you just requested for *this* season's event. #10444 defaulted to ascending, which put that row at the bottom. The default is now descending (newest first). The order of keys *with* events is fully reversible; keys with no event still trail the list either way.

## The header is the control

Instead of a two-button `btn-group` of arrows beside "Event", the **header itself is a link** that flips the direction — the conventional sortable-table idiom, and one control instead of two.

| state | header | link goes to |
|---|---|---|
| default (`/account`) | `Event` — plain, no caret | `?api_write_keys_sort=asc` |
| after clicking once | `Event ▲` (oldest first) | `?api_write_keys_sort=desc` |
| clicking again | `Event ▼` (newest first) | `?api_write_keys_sort=asc` |

The caret appears **only after the user has actively sorted**, so the default view stays a clean header; when present it points in the direction of the current order (`glyphicon-triangle-top` / `-bottom`). An invalid `?api_write_keys_sort` value falls back to the default with no caret. The link carries a `title` and `aria-label` describing what a click will do ("Sort by event (oldest first)").

## Testing

`test_api_write_keys_sorted_by_event_key` is rewritten as four cases — default, invalid value, explicit asc, explicit desc — asserting the row order, the header link's next-direction `href`, and that the caret is absent by default and points the right way when present. Account overview suite: 37 passed. `make lint` and `make typecheck` clean.

Rendered the default and clicked states locally through the Flask test client against the dev server's CSS: the default is a plain `Event` header (in link colour, so it reads as clickable) with the newest event on top; after one click it reads `Event ▲` with the oldest on top. Screenshots to follow as a comment.

### DIFF
--- a/src/backend/web/handlers/account.py
+++ b/src/backend/web/handlers/account.py
@@ -75,9 +75,17 @@ def _sorted_api_write_keys(
 @require_login
 def overview() -> str:
     user = none_throws(current_user())
-    api_write_keys_sort_direction = request.args.get("api_write_keys_sort", "asc")
-    if api_write_keys_sort_direction not in {"asc", "desc"}:
-        api_write_keys_sort_direction = "asc"
+    # Newest events first by default: the key someone just requested for this
+    # season is the one they're looking for. Clicking the "Event" header sets
+    # ?api_write_keys_sort explicitly, which also reveals the direction caret.
+    requested_sort = request.args.get("api_write_keys_sort")
+    api_write_keys_sort_explicit = requested_sort in {"asc", "desc"}
+    api_write_keys_sort_direction = (
+        requested_sort if api_write_keys_sort_explicit else "desc"
+    )
+    api_write_keys_next_sort = (
+        "asc" if api_write_keys_sort_direction == "desc" else "desc"
+    )
 
     template_values = {
         "status": session.pop("account_status", None),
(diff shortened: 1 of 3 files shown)


gregmarra:
## Screenshots

Rendered through the Flask test client against the dev server's CSS, per the recipe in AGENTS.md. Before is `main`; after is this branch. Viewport 1280×900 at 2x. Two seeded write keys (2026casj, 2025nyro) so the sort is visible. Before is the two-arrow-button control from #10444; after is the clickable header with a caret only once a sort has been chosen.

| state | before | after |
|---|---|---|
| default (before: oldest first; after: newest first, plain header) | before default | after default |
| `?api_write_keys_sort=asc` (oldest first) | before asc | after asc |
| `?api_write_keys_sort=desc` (newest first) | before desc | after desc |

cpapplefamily:
I like the UI clean up. I been wishing for this sorting system for a few years. Its been talked about in the Slack channel but no action taken. It's truly one of those "It doesn't effect me so its no a priority." for the everyday developers. 

So I finally took it on myself. Thanks for adding the final 10% here to polish it.

gregmarra:
Thank you for the kick to get it going!

PR #10542, opened 2026-09-06
Remove the last Vagrant remnants

gregmarra:
The Vagrantfile was removed in #9049 when docker compose became the dev environment. Two leftovers remained:

- `.gitignore` still ignored `.vagrant/`
- `ops/check_dev_container_version.sh` is referenced by nothing and its only output tells the user to run `vagrant halt && vagrant destroy && vagrant up`

Both are gone. `grep -ri vagrant` outside `old_py2/` now returns nothing.


ZachOrr (approved):
🙏

PR #10599, opened 2026-09-07
Add Subresource Integrity to every pinned CDN script and stylesheet

gregmarra:
## Problem

Every Jinja page loads a dozen-plus libraries from cdnjs, ajax.googleapis, and gstatic with no `integrity` attribute. If any of those CDNs is compromised or serves a tampered file, it runs on every TBA page with the user's session. (This is what CodeQL's 30 `js/functionality-from-untrusted-source` warnings are about.)

## Change

Every pinned `<script>` and `<link rel=stylesheet>` now carries `integrity="sha384-…"` and `crossorigin="anonymous"`: 48 tags across 11 templates (public `base.html`, `admin/base.html`, the tasks-service datafeeds base, login, gameday2, eventwizard, apiwrite, match, event insights, match suggestion).

Hashes were computed from the bytes each CDN serves today. **Library versions are unchanged in this PR**; the jQuery/Bootstrap/Firebase upgrades come as separate stacked PRs so each hash change is reviewable on its own.

The two Firebase tags move from `//gstatic.com/firebasejs/…` to `//www.gstatic.com/firebasejs/…`, which is where the old host 301s to anyway. Same bytes, no redirect under CORS.

Not hashed, because they can't be: Instagram's `embed.js` and the github-cards widget (both unpinned, dynamic third-party), and the Google Fonts stylesheet (content varies by user agent).

## Verification

- All four CDN hosts send `Access-Control-Allow-Origin: *`, which `crossorigin=anonymous` requires.
- Rendered `/`, `/account/login`, `/gameday`, and `/apidocs/v3` through the Flask test client and loaded them in headless Chromium against the real CDNs: every tag loaded, zero integrity errors, jQuery/Bootstrap/Firebase report their expected versions. That exercises 41 of the 48 tags in a browser; the remaining 7 (Chart.js, crypto-js, jquery-csv, xlsx, countdown, Highcharts) are on data-dependent pages and are covered by the hash computation itself.
- `make test` on the index, account, gameday, apidocs, and admin handler tests: 308 passed.

## Ongoing cost

A future version bump of any of these libraries must update the hash too, or the browser will refuse to load it. That's the point, but worth knowing.

### DIFF
--- a/src/backend/tasks_io/templates/datafeeds/base.html
+++ b/src/backend/tasks_io/templates/datafeeds/base.html
@@ -14,11 +14,11 @@
 
     <!-- Le javascript
     ================================================== -->
-    <script src="//ajax.googleapis.com/ajax/libs/jquery/1.7.1/jquery.min.js"></script>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.0.0/js/bootstrap.min.js"></script>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/fancybox/2.0.6/jquery.fancybox.pack.js"></script>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/fitvids/1.0.1/jquery.fitvids.min.js"></script>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/jquery.tablesorter/2.5.2/jquery.tablesorter.min.js"></script>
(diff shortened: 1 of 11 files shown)


gregmarra:
## Screenshots

Rendered through the Flask test client against the dev server's CSS, per the recipe in AGENTS.md. Before is `main`; after is this branch. Viewport 1280×900 at 2x. No visual change is expected from adding integrity hashes; this confirms every hashed script and stylesheet still loads.

| page | before | after |
|---|---|---|
| Home | before | after |

PR #10600, opened 2026-09-07
Upgrade the public site to jQuery 3.7.1 and Bootstrap 3.4.1; drop unused fancybox

gregmarra:
Stacked on #10599 (SRI hashes). Merge that first; this PR's diff will shrink to `base.html`, `tba.js`, `webpack.config.js`, and three deleted fancybox assets.

## Problem

- **jQuery 2.1.3**: CVE-2015-9251, CVE-2020-11022, CVE-2020-11023 (XSS when untrusted markup reaches `.html()`, `.append()`, etc.).
- **Bootstrap 3.3.2**: CVE-2018-14041, CVE-2018-20676, CVE-2018-20677, CVE-2019-8331 (XSS via `data-target`, tooltip/popover, affix).

Bootstrap 3.4.1 is the last Bootstrap 3 release and is a drop-in for 3.x sites. jQuery 3.7.1 is current.

## Change

- `base.html`: jQuery 2.1.3 → 3.7.1 (same Google CDN), Bootstrap JS 3.3.2 → 3.4.1, hashes updated.
- fancybox 2.1.5 removed. It was loaded and initialized on every page, but no template renders a `.fancybox` element (featherlight replaced it years ago). Also removes the initializer in `tba.js`, the bundled fancybox 2.0.6 CSS from the webpack entry, and the two sprite images only that CSS referenced.

Bootstrap 3.4.1 adds an HTML sanitizer to tooltips/popovers. No template or script on the site uses `data-html="true"` or `.popover()`, so nothing is affected.

## Verification

- **Live-site compatibility run.** Loaded eleven production pages (home, event, rankings tab, team, team history, match, insights, events list, apidocs, login, media suggestion form) in headless Chromium with the CDN jQuery and Bootstrap responses replaced by 3.7.1 + jquery-migrate 3.5.2 and 3.4.1, then clicked through nav tabs, tablesorter headers, the typeahead search box, lightbox links, and dropdowns. **Zero errors and zero JQMIGRATE deprecation warnings** on every page. A grep of `tba_js/` for jQuery-3-removed APIs (`.size()`, `.andSelf()`, jqXHR `.success()`, `$.browser`, `.live()`) finds none.
- **This branch in a browser.** Rendered `/`, `/account/login`, and `/apidocs/v3` through the Flask test client and loaded them in Chromium against the real CDNs: all integrity hashes pass, pages report jQuery 3.7.1 and Bootstrap 3.4.1, no console errors.
[...]

PR #10601, opened 2026-09-07
Upgrade admin and tasks-service pages to jQuery 3.7.1 and Bootstrap 3.4.1

gregmarra:
Stacked on #10600 → #10599. Merge those first; this PR's diff will shrink to `admin/base.html` and `tasks_io/templates/datafeeds/base.html`.

## Problem

The admin pages and the tasks-service datafeeds pages load **jQuery 1.7.1** and **Bootstrap 3.0.0**, older than the public site was. jQuery 1.7.1 predates every jQuery XSS fix (CVE-2012-6708 through CVE-2020-11023); Bootstrap 3.0.0 has the tooltip/popover/`data-target` XSS issues fixed in 3.4.1. Both templates also load fancybox 2.0.6, which no admin template uses.

Exposure is admin-only (these pages need a Google admin login), so this is hygiene rather than an outsider-reachable hole, but it also removes the last reason the admin needs a different jQuery than the rest of the site.

## Change

- `admin/base.html` and `tasks_io/templates/datafeeds/base.html`: jQuery 1.7.1 → 3.7.1, Bootstrap 3.0.0 → 3.4.1, hashes updated, fancybox 2.0.6 removed.
- Highcharts 3.0.2 on the admin home page is left alone; see verification.

## Verification

- Rendered `/admin/` and `/admin/tasks` through the Flask test client as an admin and loaded them in headless Chromium against the real CDNs: all integrity hashes pass, pages report jQuery 3.7.1 and Bootstrap 3.4.1, `tablesorter`/`typeahead`/`fitVids`/`featherlight` plugins are all attached, and a Highcharts 3.0.2 chart is created successfully under jQuery 3 (the admin home page draws memcache stats with it). No new console errors versus the same pages rendered from `main`.
- A grep of the admin and tasks templates for jQuery-3-removed APIs (`.live()`, `.die()`, `.size()`, `.andSelf()`, `$.browser`, jqXHR `.success()`) finds none.
- `make test` on all admin handler tests (212 passed) and the FRC API tasks handler tests that render the datafeeds templates (46 passed).

### DIFF
--- a/src/backend/tasks_io/templates/datafeeds/base.html
+++ b/src/backend/tasks_io/templates/datafeeds/base.html
@@ -14,9 +14,8 @@
 
     <!-- Le javascript
     ================================================== -->
-    <script src="//ajax.googleapis.com/ajax/libs/jquery/1.7.1/jquery.min.js" integrity="sha384-npxfGiG5C/F5X72RqcKFYSfzr1AXsDiu1YC/ydsOrS+jL554Jh4zFAx9GpQi4lXQ" crossorigin="anonymous"></script>
-    <script src="//cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.0.0/js/bootstrap.min.js" integrity="sha384-Zzs5x1/YUvlxpCu06c197tRCubLCMA7pCoHbZeoZuz/oEgYD6NVmvLzDSKYBoc3J" crossorigin="anonymous"></script>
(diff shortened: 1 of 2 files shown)


gregmarra:
## Screenshots

Rendered through the Flask test client against the dev server's CSS, per the recipe in AGENTS.md. Before is `main`; after is this branch. Viewport 1280×900 at 2x. Rendered as a GAE admin. The admin home draws the memcache chart with Highcharts 3.0.2, which is the piece most likely to care about the jQuery version; it renders (empty in the test fixture, as on `main`).

| page | before (jQuery 1.7.1, Bootstrap 3.0.0) | after (jQuery 3.7.1, Bootstrap 3.4.1) |
|---|---|---|
| `/admin/` | before | after |
| `/admin/tasks` | before | after |

PR #10613, opened 2026-09-08
Put the Realtime Database security rules in the repo and deploy them from CI

gregmarra:
## Problem

The production Realtime Database (`tbatv-prod-hrd`) rules lived only in the Firebase console. They contained:

```json
"champ_numbers": {
  ".read": true,
  ".write": "newData.hasChild('update_counter')",
  "update_counter": { ".validate": "newData.val() === data.val()+1" }
}
```

which validates the counter and nothing else, so any anonymous client can write arbitrary children under `/champ_numbers` as long as it bumps the counter. The node holds three totals from the 2017 Steamworks championship; the code that wrote them was deleted in #4150 (January 2022), nothing in the org reads the path, and neither mobile app uses the Realtime Database at all.

## Change

- `database.rules.json` (repo root): the live rules with `champ_numbers` removed. Every remaining path (`e`, `le`, `live_events`, `special_webcasts`, `api_updated`) is public-read, no client writes. The backend writes through the Admin SDK, which bypasses rules, so live event data keeps flowing.
- `firebase.json` (repo root): points the Firebase CLI at the rules file, so `firebase deploy --only database` works from the root.
- `.github/workflows/push.yml`: new `deploy-firebase-rules` job, modeled on `deploy-index`. Runs on push to `main` when either file changes, authenticates with `GCLOUD_AUTH`, and runs `firebase-tools deploy --only database`. RTDB rules are a Firebase resource, not an App Engine one, so this is the CLI rather than `gcloud app deploy`.
- Short section in `docs/Developing/Development-Runbook.md`.

The local emulator config in `ops/dev/firebase/` is unaffected; it runs without rules.

## One-time setup needed before the CI job can deploy

The `github-actions@tbatv-prod-hrd` service account has App Engine, Cloud Tasks, Scheduler, Storage, and Datastore index roles but **no Firebase Database role**, so the job will fail until an owner grants it:

```
gcloud projects add-iam-policy-binding tbatv-prod-hrd \
  --member serviceAccount:github-actions@tbatv-prod-hrd.iam.gserviceaccount.com \
  --role roles/firebasedatabase.admin
```

Until then, deploying by hand works for anyone with console access: `firebase deploy --only database --project tbatv-prod-hrd` from the repo root.

## Verification

- Both JSON files parse; `push.yml` parses.
- The rules file is byte-for-byte the live rules minus the `champ_numbers` block (the live copy also has a trailing comma the console tolerates; this file is strict JSON).


tervay:
We've since added match_suggestions to the security rules, so that needs to be added before this gets merged

PR #10663, opened 2026-09-11
Preview SmugMug album contents on the PWA suggestion review page

gregmarra:
## Problem

SmugMug album suggestions on `beta.thebluealliance.com/suggest/review/event_media` render as a bare link. A reviewer has to open every album in a new tab to see whether it's a real gallery for that event.

## What's already there

`MediaParser._parse_smugmug` looks the album up in SmugMug's API when the suggestion is created, so the suggestion's `details_json` already holds the title, `web_uri`, `image_count`, and cover URLs. The moderation API only surfaced image fields for `is_image` media, which albums are not, so none of it reached the PWA.

## Why not an embed

SmugMug's oEmbed endpoint returns `<iframe src="https://api.smugmug.com/services/embed/album/<key>">`, and that page serves an empty `<body>` (verified both top-level and framed, headless Chromium, no console errors, no images). The gallery page itself sends `X-Frame-Options: DENY`. So there is no iframe that shows anything.

## Change

**Backend**
- New `common/helpers/smugmug_helper.py` with `album_preview_images(album_key)`: one `album!images` request (`count=48`, `_expand=ImageSizeDetails`) using the existing `SmugmugApiSecret` key, then 8 photos sampled evenly across the window. Albums are uploaded in shooting order, so the first 8 are usually eight frames of the same moment; the first screenshot below is what that looked like. Memoized 6 hours. Any failure (no key, non-200, bad JSON, odd shapes) returns `[]`.
- `_serialize_candidate_media` for `SMUGMUG_ALBUM` now emits `view_image_url` (album page), `image_direct_url` (cover), `title`, `image_count`, and `preview_images: [{thumbnail_url, image_url, web_uri}]`.
- OpenAPI spec updated (`CandidateMedia` gains the three fields; new `SmugmugPreviewImage` schema); PWA client regenerated.

**PWA**
- `SmugmugAlbumPreview` in the review card: a 4×2 grid of the sampled photos, each linking to its SmugMug page, with `title · N photos` beneath. Falls back to the cover image when there is no sample. `is_image` stays false, so nothing else about album handling changes.

## Tests

- `smugmug_helper_test.py`: response mapping incl. the size expansion and thumbnail fallback, even sampling across the window, and every failure path (no key, HTTP error, request exception, bad JSON, unexpected shape).
- `moderation_test.py`: an event-media SmugMug album suggestion returns the new fields with the helper patched.
- `make lint`, `make typecheck`, 63 tests in the two files pass; `pnpm run format/lint/typecheck` clean.

## Screenshots

In a comment below.

### DIFF
--- a/pwa/app/api/tba/moderation/index.ts
+++ b/pwa/app/api/tba/moderation/index.ts
@@ -42,6 +42,7 @@ export {
   type RejectResponse,
   ReviewResult,
   type SimilarEvent,
+  type SmugmugPreviewImage,
   type SuggestionAuthor,
   type SuggestionListResponse,
   type SuggestionReference,
--- a/pwa/app/api/tba/moderation/types.gen.ts
+++ b/pwa/app/api/tba/moderation/types.gen.ts
@@ -84,6 +84,18 @@ export type CandidateMedia = {
   view_image_url?: string;
   image_direct_url?: string;
   social_profile_url?: string;
+  /**
+   * Album or photo title, when the provider supplied one (SmugMug)
+   */
+  title?: string;
+  /**
+   * Number of photos in the album (SmugMug albums)
+   */
+  image_count?: number;
+  /**
+   * The first few photos of the album, for previewing its contents (SmugMug albums)
+   */
+  preview_images?: Array<SmugmugPreviewImage>;
 };
 
 /**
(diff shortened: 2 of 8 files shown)


gregmarra:
## Screenshots

Rendered locally against the auth emulator with the moderation API mocked using the real SmugMug API response for the album in the report (NE FIRST, 2025 CT States, 1,972 photos), sampled the same way the helper does: 8 photos spread across a 48-photo window. Light theme, 2x.

| before | after |
|---|---|
| before | after |

PR #10666, opened 2026-09-11
Sync every PWA tab bar with the URL hash

gregmarra:
## What

Every tab bar on a details page now lives in the URL hash, as on the Jinja site:

- `/event/2024mil#rankings` opens Rankings; clicking Media makes the URL `…#media`.
- District overview (`#overview #rankings #events #teams`), district champs (`#all-rankings #all-matches`), district stats, district insights, and myTBA (`#teams #events`) all do the same.
- The event page also accepts the old site's names, `#event-insights` and `#cmp-points`, so links people have saved keep working.
- Hash writes use `replace`, so a click never adds a history entry and the back button still leaves the page. Back/forward across hashes switches tabs.

## How

New `useHashTab({ values, defaultValue, legacyHashes })` in `app/lib/useHashTab.ts`, used by each page:

```tsx
const tabs = useHashTab({ values: EVENT_TAB_VALUES, defaultValue: 'results', legacyHashes });
<AnimatedTabs key={tabs.key} defaultValue={tabs.defaultValue} onValueChange={tabs.onValueChange}>
```

The tabs stay **uncontrolled**. The server never sees the hash, so the page renders and hydrates on the default tab with no mismatch; once the client reads the hash, `key` changes and the tab bar remounts on the hashed tab. The hash is read through `useSyncExternalStore` subscribed to `hashchange` only. Our own clicks go through `navigate({ hash, replace: true })`, which is `history.replaceState` and fires no `hashchange`, so they never remount; only external changes (initial load, back/forward) do. `hashScrollIntoView: false` and `resetScroll: false` keep the page from jumping.

I tried a controlled `value` first; Base UI's controlled Tabs changed hydration timing enough to trip the indicator test below, and controlled/uncontrolled switching isn't supported by Base UI's `useControlled`.

## Animated tab indicator

#10557 lazy-loads the framer-motion indicator "after hydration", but only by timing: it was requested ~100 ms after the hydration marker on warm loads and coincided with it on cold ones, so `defers the animated tab indicator until a tab changes` raced (it failed on this branch's first cold run and passed on every warm one). `AnimatedTabs` now renders the static CSS indicator until the first idle callback after mount, then swaps in the lazy motion component. That is deterministically after the root's hydration effect. The second test is rewritten to assert the chunk arrives after load rather than on click, since it now loads at idle.

## Tests

- `event.spec.ts`: click → `#rankings`/`#awards` with no history entry; `#media` opens Media; `#event-insights` opens Insights; unknown hash falls back to the default; indicator deferral (deterministic) and idle load.
- `district.spec.ts`: click → `#teams`; `#events` opens Events.
- Ran the two specs five times: 32/32 on four runs; one run had a single `toBeVisible` failure in an existing test unrelated to tabs (data-dependent), which passed on rerun.
- `pnpm run format`, `lint`, `typecheck` clean.

No screenshots: the visible change is the address bar.


gregmarra:
Overall this is a really solid PR — using `useSyncExternalStore` with uncontrolled remounting via `key` is a clean way to handle SSR/hydration with hash URLs without hydration mismatches, and using `requestIdleCallback` for the indicator chunk is much more deterministic than arbitrary timers.

A couple of issues/suggestions to consider:

### 1. Division tabs missing from `useHashTab` values in District Champs page (Bug)
In `pwa/app/routes/district.$districtAbbreviation.champs.$year.tsx`:
```tsx
const tabs = useHashTab({
  values: ['all-rankings', 'all-matches'] as const,
  defaultValue: 'all-rankings',
});
```
However, that page also dynamically renders tabs for each championship division (`divisionTabIds = cmpDivisions.map((div) => div.short_name ?? div.name)`). Because those division names are not in `values`:
- When clicking a division tab, the URL updates (e.g. `#DTE`).
- But reloading or deep linking to `#DTE` rejects it as an unknown hash and falls back to `all-rankings`.

### 2. Stale closures in `useHashTab` if `values` is dynamic
In `pwa/app/lib/useHashTab.ts`:
```ts
const readHash = useCallback((): T | undefined => {
  const hash = window.location.hash.slice(1);
  const value = legacyHashes[hash] ?? hash;
  return (values as readonly string[]).includes(value)
    ? (value as T)
    : undefined;
  // values/legacyHashes are static per call site
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, []);
```
If `values` is passed dynamically (such as when fixing point 1 above where `cmpDivisions` loads asynchronously), `readHash` will capture the initial/empty array because of the empty deps array. Storing `values` and `legacyHashes` in a ref (e.g., `const valuesRef = useRef(values); valuesRef.current = values;`) or including them in dependencies would make it safe for dynamic tab lists.

### 3. Formatting
In `account.mytba.tsx`, `district.$districtAbbreviation.champs.$year.tsx`, and `district.$districtAbbreviation.stats.tsx`, the closing bracket `>` for `<Tabs` is on the same line as `className="mt-4">`:
```tsx
onValueChange={(value) => tabs.onValueChange(String(value))} className="mt-4">
```
Formatting should break that onto its own line.

gregmarra:
Thanks, all three addressed in the latest commit.

**1. Division tabs.** Real bug. The hook call now sits below `divisionTabIds` and passes `['all-rankings', 'all-matches', ...divisionTabIds]`, so `#<division>` deep links open the division. Added a Playwright test on `/district/fim/champs/2019` that clicks the first division tab, checks the hash, reloads with that hash, and checks the tab is selected.

**2. Stale closure.** Fixed without refs: the `useSyncExternalStore` snapshot is now just the raw hash string, and which tab it names is decided during render against the current `values`. So when `cmpDivisions` loads and the tab list grows, the same hash resolves on that render and the tab bar remounts onto it. (I tried the ref approach first; the React lint rule rejects reading `ref.current` during render, and this is cleaner anyway.)

**3. Formatting.** Already fixed by the earlier prettier commit; the one-line `<Tabs …>` tags are gone. Root cause was that I ran `pnpm run format` (the check) instead of `pnpm run format:fix` (the write), which AGENTS.md already says to use before pushing.

One more change while I was in here: the idle-callback deferral of the motion indicator still raced Playwright's detection of `data-hydrated` (an idle browser fires the callback before the poll notices the attribute). The chunk now loads on the first pointer-enter or focus on the tab bar, or a tab change. That's deterministic for both indicator tests, and in practice the chunk has arrived by the time the user actually clicks, so the first click still animates. The original "loads when a tab changes" test is back unchanged. Event + district specs: 33/33, three consecutive runs.

gregmarra:
Rebased onto `main` after #10675 reorganized the district routes; squashed to one commit while at it.

What changed in the resolution:

- **District overview** (`/district/fim/2024`) now has `rankings`, `events`, `teams`, and the new `champs` tab in its hash values, default `rankings` (or `events` when there are no rankings), matching main's new default.
- **District champs** moved from its own route into `DistrictChampsTab`, rendered *inside* the district page's tab bar. Two hash-synced tab bars on one page would fight over the single hash (`#all-rankings` would deselect the outer `champs` tab on reload), so the inner division tabs are deliberately left uncontrolled. The earlier division-tab hash test went away with the route; `/district/fim/2024#champs` still opens the champs tab.
- **District insights** re-applied on main's new `championships`/`events`/`awards` tabs.
- **District stats** route was deleted on main; nothing to sync.

Event + district specs: 45/45 locally (after evicting an unrelated `next-server` that was squatting on port 3000 and timing out every district test).

PR #10668, opened 2026-09-11
Add the pwa format:fix script that AGENTS.md already documents

gregmarra:
`pwa/AGENTS.md` says:

```
pnpm run format:fix       # Auto-format (Prettier; also sorts Tailwind classes) — run before pushing, CI checks `format`
```

but `pwa/package.json` only defines `format` (`prettier --check .`). Anyone following the docs gets "Missing script", and anyone who reaches for `pnpm run format` instead gets a check that writes nothing while looking like it did. That's how #10666 shipped six unformatted files to CI.

This adds `"format:fix": "prettier --write ."` and leaves `format` as the check CI runs, so the docs and CI both stay correct.

### DIFF
--- a/pwa/package.json
+++ b/pwa/package.json
@@ -16,6 +16,7 @@
     "lint:fix": "oxlint --fix",
     "format": "prettier --check .",
     "format:fix": "prettier --write .",
+    "format:fix": "prettier --write .",
     "start": "cross-env NODE_ENV=production node ./server.js",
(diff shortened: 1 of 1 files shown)


gregmarra:
Withdrawn: `format:fix` already exists in `pwa/package.json` (line 18). My grep for `"format"` missed it and this PR would have added a duplicate key. The real mistake in #10666 was running `pnpm run format` (the check) instead of `pnpm run format:fix`, exactly as AGENTS.md says.

PR #10677, opened 2026-09-12
Give PWA blue banners the notch and fixed layout of the Jinja banners

gregmarra:
## Problem

On `beta.thebluealliance.com/team/6201`, the district championship banner had "WINNER" printed over "2026", and the event line ran six lines of the full event name. Two causes, plus a missing detail:

- The title and event text were positioned with absolute offsets from the top and the bottom respectively, so a long event name grew upward into the award name.
- It used `event.name` ("New England FIRST District Championship - Newsom Division presented by GE Aerospace") where the Jinja site uses `Event.normalized_name` ("NEDC - Newsom District Championship Division").
- No V notch at the bottom, which every TBA banner has had since the 2014 CSS.

## Change

`Banner` is now a flex column: logo, a fixed-height title zone, and an event zone that fills the rest, vertically centres its text, and clamps it to four lines. The event line is `<year> <normalized name>` as one run of text, like the Jinja macro. The notch is a `clip-path` cut from the bottom 24 px, so it scales with whatever size a caller sets; the hall of fame passes `h-48 w-28` and was overlapping too, and is fixed by the same change.

New `getEventNormalizedName(event)` in `eventUtils.ts` mirrors the Python `normalized_name`: short name plus the type suffix unless already present, `<city> Championship` for Einstein from 2017, bare short name for offseasons, full name when there is no short name or for FoC. Unit tests cover each branch.

## Screenshots

In a comment below: the Jinja banner for the same award as reference, beta today, and this branch.

## Test plan

- [x] `pnpm run format:fix`, `lint`, `typecheck`
- [x] `vitest` on `eventUtils.test.ts` (95 tests incl. 5 new)
- [x] Rendered `/team/6201` and `/hall-of-fame` from this branch against the live API and compared with beta

### DIFF
--- a/pwa/app/components/tba/banner.tsx
+++ b/pwa/app/components/tba/banner.tsx
@@ -3,6 +3,7 @@ import { Fragment } from 'react';
 
 import { Award, Event } from '~/api/tba/read';
 import { getNormalizedName } from '~/lib/api/AwardType';
+import { getEventNormalizedName } from '~/lib/eventUtils';
 
 export function AwardBanner({
   award,
@@ -16,7 +17,7 @@ export function AwardBanner({
   return (
     <Banner
       title={getNormalizedName(award.award_type, event.event_type, event.year)}
-      description={event.name}
+      description={getEventNormalizedName(event)}
       className={className}
       year={event.year}
     />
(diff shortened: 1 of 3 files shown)


gregmarra:
## Screenshots

Reference is FIRST's own store render. Before is beta today; after is this branch rendered locally against the live API. 2x. Updated after the readability pass: default width 160px, smallest text 11.2px (long event names), everything else 12.8px or larger; contrast 5.9:1.

| store render (reference) | before (beta) | after: 2026 district champs |
|---|---|---|
|  | before | after |

Two regional winner banners (team 254, 2023), and the hall of fame (now the same size as everywhere else):

| after: regionals | before hof | after hof |
|---|---|---|
|  | before hof | after hof |

tervay:
I had intentionally left the notch out in order to align them closer with what actual banners look like. Personally I prefer the rectangular ones but open to discussion.

tervay (approved):

Issue #10723, opened 2026-09-17
Suggestion emails have been dead since the Python 3 port (Oct 2020)


gregmarra:
## What's broken

No suggestion-related email has been sent since the suggestion handlers were ported to Python 3 in October 2020 (#3118, #3119). The whole email helper is inside a docstring, and every call site is commented out with a `TODO port outgoing email` note.

**Dead helper:** `src/backend/common/helpers/outgoing_notification_helper.py` — `send_admin_alert_email` and `send_suggestion_result_email` are both inside a triple-quoted string. Only `send_slack_alert` is live code.

**Dead call sites** (all inside docstrings or `#` comments):

| Where | Email | Recipient |
|---|---|---|
| apiwrite review (accept) | "Your auth tokens for {event} are ready" + moderator message | requester |
| apiwrite review (reject) | "We have regretfully declined ... with the following message: {message}" | requester |
| apiwrite review (either) | admin alert with the moderator's verdict + message | contact@ |
| offseason event review (accept) | "Your suggestion has been approved, find it at /event/{key}" + EventWizard pointer | suggester |
| apiwrite request submission | admin alert that a new key request is pending | contact@ |
| offseason event submission | admin alert that a new event was suggested | contact@ |

## Why it matters

The moderator-facing side partially survived as Slack: the moderation API posts the apiwrite verdict + message to the `suggestion-nag` channel, and the daily nag covers pending queues. But the **suggester-facing** emails are gone. When a moderator types a rejection reason into the apiwrite review form, that message only reaches Slack. The person who requested keys never sees it, and never hears that their keys were issued either.

## Feasibility

`google.appengine.api.mail` imports fine on the python313 runtime and both `web.yaml` and `api.yaml` already set `app_engine_apis: true`, so porting is a matter of un-commenting and re-plumbing, not infrastructure.

## Options

1. **Fix:** port `send_suggestion_result_email` / `send_admin_alert_email` for real (skip in dev like the original did), and call them from `SuggestionReviewer` (apiwrite accept/reject, offseason accept) and the two submission handlers. Needs tests that assert the mail call with a stubbed mail API.
2. **Delete:** remove the docstring helper and the four commented blocks so the code stops implying it emails anyone.

The Jinja review controllers that hold the review-side comments are being deleted in favor of the beta review UI; whichever option is chosen should land against `backend/common/suggestions/suggestion_reviewer.py` and `suggestion_submission.py`.

PR #10724, opened 2026-09-18
Remove the Jinja suggestion review pages in favor of the beta PWA

gregmarra:
The moderator suggestion review UI now lives on the PWA at `beta.thebluealliance.com/suggest/review`, backed by the moderation API and the shared `SuggestionReviewer`. This PR deletes the Jinja review surface and fixes everything that still depended on it. Part of retiring Jinja ahead of next season.

## Deleted
- The 8 `suggest_*_review_controller.py` handlers, `suggestion_review_base.py`, `suggestion_review.py`, and their blueprint registration
- The 8 review templates, `suggest_review_home.html`, `pending_suggestion_rows_partial.html`, and two partials that were only used by them (`event_partials/suggest_event_info.html`, `media_partials/image_thumbnail_partial.html`)
- The 9 review controller test files
- `require_any_permission` (only the review home used it) and its tests
- A `tba.js` special case for the review nav pills

## Team admin review kept working
`/mod` let team admins accept/reject their own team's media, social media, and CAD suggestions by POSTing to three of the deleted controllers. Their authorization came from `TeamAdminAccess`, which the shared reviewer didn't know about.

- `SuggestionReviewer` gains `user_can_review_suggestion(user, suggestion, delegated_team_keys)` and threads `delegated_team_keys` through accept/reject. Delegation only applies to `TEAM_ADMIN_REVIEWABLE_TYPES` (media, social-media, robot) and only for the delegated team's target key.
- New `POST /mod/review` parses the existing dashboard form and calls the reviewer. It authorizes the whole batch up front (one unauthorized key = 403, nothing applied) and always redirects to `/mod`. The old controllers redirected to a caller-supplied `return_url`, which was an open redirect.
- This also makes team-admin review available to the API/PWA later.

## Links and redirects
- Slack nag now links only to Beta
- `/suggest/review` and the 8 old `/suggest/*/review` URLs 302 to their PWA equivalents so bookmarks and previously-posted Slack nags keep working (GET only; POST is 405)
- Account overview and admin index link to the PWA review UI via a new `pwa_url_context_processor`
- New `PwaUrlHelper` is env-aware: local dev gets `http://localhost:5173`, never production
- The local dev seed tool redirects to the local PWA dev server

## PWA parity (added after review)
- The Jinja review home was the only page linking to the **Offseason Dashboard** (`/mod/offseasons`). The PWA review home now shows a Review Offseason Tools section with that link for reviewers who can review offseason events, alongside the existing media tools.
- The offseason event review card now has a **Preseason/Offseason** selector. The API already accepted `event_type_enum` but the PWA never sent it, so every accepted suggestion became an offseason event.
- The review tools moved into `reviewTools.tsx` so they can be unit tested.
- The PWA account page now has the same **Suggestion Reviews** block the Jinja account page has, with the pending total and a Review Pending Suggestions button. Shown when the moderation queue endpoint returns a queue (it 403s for non-moderators). Before this, a moderator on beta had no in-app path to `/suggest/review`.

## What moderators will notice
- The Slack nag and the account/admin buttons now land on `beta.thebluealliance.com`. Beta has its own Firebase session, so the first visit asks you to sign in once.
- Old www bookmarks keep working via redirect.
- `/mod` (team admin dashboard), `/mod/webcasts`, and `/mod/offseasons` are unchanged and still on www; the PWA review home links to them.
- Rollback is a plain revert: nothing here migrates data.

## Review follow-ups
A high-effort code review turned up a few things, all fixed here:
- The account page card asked the moderation queue for everyone; a 403 is now "not a moderator" (`null`) rather than a query error, so nothing is reported to Sentry.
- `SuggestionReviewer` looks up `TeamAdminAccess` itself, so any caller gets delegation without remembering a kwarg. `/mod` derives its allowed types from the reviewer's constant.
- `/mod` only lists rows the user can act on; `/mod/review` validates keys, loads the batch in one `get_multi`, and shows unapplied outcomes (already reviewed, invalid) on the dashboard instead of redirecting silently.
- Audit entries record which suggestion was acted on in `url_args`, on the API path too.
- Offseason FIRST codes are stripped again before uppercasing, as the deleted controller did.

Second round:
- A cleared FIRST code makes the event unofficial; only an untouched field falls back to the suggested code.
- The offseason event type defaults from the type the suggestion already carries (Jan/Feb suggestions are preseason), in the reviewer and in the PWA select.
- `/mod/review` returns a forced-team viewer to `?team=&year=` instead of bouncing them to `/mod/redeem`, reports every unapplied outcome, and measures key length in bytes.
- The moderation API stays permission-only (explicitly), so its read and write sides agree; delegation lives on `/mod`.
- The PWA queue hook only treats the permission 403 as "not a moderator"; an unverified-email 403 still surfaces. The moderation client now gets the same `ApiError` mapping as the other clients.
- The navbar's "Try beta" link uses the env-aware PWA base URL.

## Tests
- `pwa_url_helper_test.py` (new): prod/dev/unit-test base URLs, every suggestion type path
- `suggestion_reviewer_test.py` (new): delegated permission matrix (own team, other team, non-team types, missing target key), accept/reject with delegation, forbidden leaves suggestions pending, audit log written
- `suggestion_review_redirects_test.py` (new): every legacy path, dev vs prod target, no login required, POST rejected
- `suggest_team_admin_review_test.py`: 13 existing tests ported to `/mod/review` (401 -> 403), plus 11 new: login redirect, GET 405, redirect target, `return_url` ignored, other team forbidden, batch with one forbidden applies nothing, unknown key 400, no access 403, global permission works without team access, year override, audit log
- `slack_nag_test.py`: asserts message content and the Beta link
- `account_overview_test.py`, `admin_blueprint_test.py`, `test_seed_media_suggestions.py`: assert the new link targets
- PWA: `reviewTools.test.tsx` (dashboard links, manage-team-media URL), `moderation/suggestionReviewCard.test.tsx` (event type select defaults to Offseason, sends a numeric `event_type_enum`), `account/suggestionReviewCard.test.tsx` (pending total, link to the review home)

`make lint`, `make typecheck`, and the web/tasks_io/suggestions/moderation suites (1647 passed) are green.

## Related
- #10723: suggestion emails have been dead since the Python 3 port. The commented-out email blocks lived in the controllers deleted here; the fix belongs in `SuggestionReviewer`.


bdaroz:
@gregmarra Keep in mind the `prod` Jinga template is the only one with a link/button to the Offseason Dashboard. That is not on the PWA `/suggest/review` page.

gregmarra:
Good catch, thanks. Added a Review Offseason Tools section to the PWA review home with the Offseason Dashboard link, shown to anyone who can review offseason events (same gating as the media tools). While checking parity I also found the PWA offseason card never sent `event_type_enum`, so every accepted suggestion became an offseason event; there's now a Preseason/Offseason selector. Both have unit tests.

